PR #26, opened 2025-09-03
feat: allow specifying custom sampler class for use with `random_walk` operator

michael-johnston:
This PR changes the parameters of `random_walk` in two ways. 

First, it moves the fields `samplerType`, `mode` and `grouping` under a new field called `samplerConfig`.
A new model called `BaseSamplerConfiguration` has been created for this field and an upgrade path provided. 
Logic related to instantiating the sampler class has also been moved to this model.

Second, it adds an additional model `CustomSamplerConfiguration` that can also be given as the value of this field. 
A discriminator has been added to aid identifying the correct model to use. 
The `CustomSamplerConfiguration` can be used to specify a subclass of `BaseSampler` to use along with parameters to it

This PR also makes a small addition to `discoveryspace.samplers.BaseSampler`. 
It adds a new class method "parameters_model" which by default returns `None`.
This can be used by custom samplers to provide a model for their parameters, if required. 
`CustomSamplerConfiguration` then uses this to validate the provided parameters.

Docs have all be updated accordingly

### Example of changes for existing fields (mode, samplerType, grouping)

A parameters model that looked like

```yaml
parameters:
  batchSize: 1
  mode: random
  samplerType: generator
  numberEntities: 1
  singleMeasurement: true
  filter:
      filterMode: unmeasured
```

now is

```yaml
parameters:
  batchSize: 1
  samplerConfig:
    mode: random
    samplerType: generator
  numberEntities: 1
  singleMeasurement: true
  filter:
      filterMode: unmeasured
```

### Example of specifying a custom sampler

ExplicitGridSamplerSelector has been slightly modified to enable it to be used directly via this method.
The above YAML can now also be expressed using: 

```yaml
parameters:
  batchSize: 1
  samplerConfig:
    module: 
      moduleClass: ExplicitGridSampleSelector 
      moduleName: orchestrator.core.discoveryspace.samplers
    parameters:
       mode: random
  numberEntities: 1
  singleMeasurement: true
  filter:
      filterMode: unmeasured
```

### DIFF
--- a/examples/ml-multi-cloud/randomwalk_ml_multicloud_operation.yaml
+++ b/examples/ml-multi-cloud/randomwalk_ml_multicloud_operation.yaml
@@ -12,6 +12,7 @@ operation:
   parameters:
     numberEntities: 48
     batchSize: 1
-    mode: 'random'
-    samplerType: 'generator'
     singleMeasurement: True
+    samplerConfig:
+      samplerType: generator
+      mode: random
--- a/orchestrator/core/discoveryspace/samplers.py
+++ b/orchestrator/core/discoveryspace/samplers.py
@@ -7,6 +7,7 @@ import logging
 import typing
 
 import numpy as np
+import pydantic
 import ray
 
 from orchestrator.core.discoveryspace.space import (
@@ -40,6 +41,12 @@ class WalkModeEnum(enum.Enum):
 class SamplerTypeEnum(enum.Enum):
     SELECTOR = "selector"
     GENERATOR = "generator"
+
+
+class BaseParameters(pydantic.BaseModel):
+
+    mode: WalkModeEnum
+    model_config = pydantic.ConfigDict(extra="forbid")
 
 
 class BaseSampler(abc.ABC):
@@ -82,6 +89,12 @@ class BaseSampler(abc.ABC):
             batchsize: The iterator will return entities in batches of this size
 
         """
+
+    @classmethod
+    def parameters_model(cls) -> type[pydantic.BaseModel] | None:
+        """Returns a pydantic model for the init parameters of the class, if any"""
+
+        return None
 
 
 class GroupSampler(BaseSampler):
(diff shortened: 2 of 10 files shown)


orchestrator/core/discoveryspace/samplers.py
```diff
@@ -42,6 +43,12 @@ class SamplerTypeEnum(enum.Enum):
     GENERATOR = "generator"
 
 
+class BaseParameters(pydantic.BaseModel):
```
AlessandroPomponio, line 46: Maybe `BaseParameters` is too generic of a name - we could have it be `BaseSamplerParameters`, which  would be a consistent naming scheme with `BaseSamplerConfiguration`, although maybe the two might then risk being confused between one another

examples/ml-multi-cloud/randomwalk_ml_multicloud_operation.yaml
```diff
@@ -8,10 +8,12 @@ spaces:
 - 'space-630588-bfebfe'
 operation:
   module:
-    moduleClass: RandomWalk
+    operatorName: "random_walk"
```
AlessandroPomponio, line 11: Should there be an upgrade validator for this as well?

website/docs/operators/explore_operators.md
```diff
@@ -132,12 +132,15 @@ config:
     module:
       moduleClass: RandomWalk
       moduleName: orchestrator.modules.operators.randomwalk
```
AlessandroPomponio, line 134: When does one use `moduleClass` vs `operatorName`?

examples/ml-multi-cloud/randomwalk_ml_multicloud_operation.yaml
michael-johnston: It's not an upgrade - there are two possibilities, both valid.

We could consider later only allowing one possibility, but it requires more changes as internally explore operators eventually need this field to be "moduleClass/modulePath" i.e. when you use operatorName its eventually substituted to the other type.

website/docs/operators/explore_operators.md
michael-johnston: You can always use operatorName, you may if you want use moduleClass for explore operations but its not presented as a method.

Internally they switch it, which is why here, which is the dump of the YAML actually used its changed.

AlessandroPomponio (approved):
LGTM thanks

PR #28, opened 2025-09-05
chore(deps): do not pin numpy<2 anymore

AlessandroPomponio:
## Changes

- Requires `numpy` instead of `numpy<2`
- Uses generic TypeError instead of private numpy Exception

Resolves https://github.com/IBM/ado/issues/29


srikumar003:
We have performed experiments succesfully with [bohb](https://github.com/automl/HpBandSter) through Ray.tune which is an alternative to Ax

michael-johnston (approved):

PR #41, opened 2025-09-09
fix: configuring Trainer to exit a training job early

VassilisVassiliadis:
This PR has workers with local-rank 0 decide when to early-stop the training based on the measurements configuration (currently, `stop_after_seconds` and `auto_stop_method`). When all local-rank 0 workers agree, all workers stop training.

The code uses a reduce operation to implement the above.

### DIFF
--- a/plugins/actuators/sfttrainer/ado_actuators/sfttrainer/wrapper_fms_hf_tuning/scripts/wrapper_sfttrainer.py
+++ b/plugins/actuators/sfttrainer/ado_actuators/sfttrainer/wrapper_fms_hf_tuning/scripts/wrapper_sfttrainer.py
@@ -16,15 +16,13 @@ import datetime
 import os
 import sys
 import typing
-
-# Standard
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


plugins/actuators/sfttrainer/ado_actuators/sfttrainer/wrapper_fms_hf_tuning/scripts/wrapper_sfttrainer.py
```diff
@@ -323,6 +311,36 @@ def on_step_end(
...
+                )
+
+        sys.stderr.flush()
+        sys.stdout.flush()
+
+        if (
+            torch.distributed.is_initialized()
+            and torch.distributed.get_world_size() > 1
+        ):
+            # VV: Stop all workers if all rank 0 workers decided to stop training
```
VassilisVassiliadis, line 326:
```suggestion
            # VV: Stop all workers if all local-rank 0 workers agree to stop
```

PR #42, opened 2025-09-10
feat: add support for fms-hf-tuning==3.0.0 in SFTTrainer experiments

VassilisVassiliadis:
Changes:

- add support for fms-hf-tuning==3.0.0 in SFTTrainer experiments
- document that using fms-hf-tuning with GPUs requires having an appropriate version of torch already installed in the base virtual environment


AlessandroPomponio (approved):
LGTM thanks

PR #45, opened 2025-09-10
fix: measuring properties in the example_actuator

VassilisVassiliadis:
Fixes a small bug in the example actuator.


AlessandroPomponio (approved):
LGTM thanks

PR #49, opened 2025-09-12
refactor: Property and PropertyValue models

michael-johnston:
### Summary

This PR introduces additional simpler Property models and updates other models to use them where appropriate. It also simplifies and specialises PropertyValue: New subclasses are introduced for ConstitutivePropertyValue, ObservedPropertyValue and VirtualObservedPropertyValue, and all PropertyValue subclasses use the simpler Property classes. Other models and code are updated to use the specific class they actually require. 

This is done to
- remove the Union in PropertyValue.property which was necessitating model_rebuilds and also caused a recursive schema for ExperimentReference
- simplify the representations of properties and property-values in many models by removing unused/unnecessary fields (as stored in metastore and seen via e.g. `ado template`)
- simplify the schemas of these models similarly

### Breaking

This PR requires (re)installing all actuators, operators and custom_experiments distributed in ado repo.
External actuators/operators/custom_experiments must be updated to use the new PropertyValue subclasses.

### Additions

- Adds "Descriptors", variants of existing Property classes without propertyDomain: PropertyDescriptor, ConstitutivePropertyDescriptor, AbstractPropertyDescriptor, ConcretePropertyDescriptor
- Remove ObservedProperty and VirtualObservedProperty from union of PropertyValue.property
- Introduce specific PropertyValue subclasses: ConstitutivePropertyValue, ObservedPropertyValue and VirutalObservedPropertyValue. 
- For PropertyValue sub-classes with a descriptor equivalent, the type of the "property" field is the descriptor class
- Provide methods for converting between the base and descriptor variant of relevant properties
- For flexibility PropertyValues can still be initialised with serialised representations of compatible Property instances which are simplified to the relevant descriptor. For example  a ConstitutivePropertyValue can be initialised with a dict representation of a ConstitutiveProperty or a ConstitutivePropertyDescriptor. 

### Model Changes

All models with fields typed with PropertyValue now use the specific PropertyValue subclass they need
- Entity.consitutivePropertyValues (ConstitutivePropertyValue)
- MeasurementResult.measurement (ObservedPropertyValue)
- Experiment.defaultParameterization (ConstitutivePropertyValue)
- ParameterizedExperiment.parameterization (ConstitutivePropertyValue)
- ExperimentReference.parameterization (ConstitutivePropertyValue)

The return types and input/types of methods/function are also updated accordingly where relevant. In particular some methods in the above models now return ConstitutivePropertyDescriptor instead of ConstitutiveProperty

In addition the `targetProperty` field of Experiment now is a list of AbstractPropertyDescriptor. This is because no code accessing this field required domain information.

### Other code changes

The main change is to actuators which now must use ObservedPropertyValue when creating MeasurementResults for experiments. Using PropertyValue with this PR will currently raise a ValidationError. 

Similarly any code creating an Entity or Experiment must pass the correct types. 

### Tests

All fixtures and tests have been updated where relevant

### DIFF
--- a/orchestrator/schema/property.py
+++ b/orchestrator/schema/property.py
@@ -186,16 +186,11 @@ class AbstractProperty(Property):
     def __eq__(self, other):
 
         retval = super().__eq__(other)
-        if retval:
-            try:
-                retval = (
-                    self.concretePropertyIdentifiers
-                    == other.concretePropertyIdentifiers
-                )
-            except AttributeError:
-                retval = False
-
-        return retval
+        return (
+            retval
+            and hasattr(other, "concretePropertyIdentifiers")
+            and self.concretePropertyIdentifiers == other.concretePropertyIdentifiers
+        )
 
     def descriptor(self):
 


orchestrator/schema/property.py
```diff
@@ -29,6 +29,81 @@ class NonMeasuredPropertyTypeEnum(str, enum.Enum):
...
+
+        Metadata is not included"""
+
+        try:
+            retval = self.identifier == other.identifier
+        except AttributeError as error:
+            print(error)
+            retval = False
+
+        return retval
```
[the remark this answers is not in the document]
michael-johnston: When the other object does not have the field "identifier" - in which case it's definitely not a Property instance.

orchestrator/schema/measurementspace.py
```diff
@@ -560,15 +560,17 @@ def observedPropertiesForExperiment(
     @property
     def targetProperties(
         self,
-    ) -> list[AbstractProperty]:
+    ) -> list[AbstractPropertyDescriptor]:
         """Returns a list of AbstractProperties in the measurement space"""
 
         # If more that one observed property measures same target we just need to
         # keep one of them
```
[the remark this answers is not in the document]
michael-johnston: Yes

Basically it's the same as saying this should return a set (this was difficult before in many places due to issues meaning PropertyDomain could not be hashed - but could be done now.  However I wouldn't do it as part of this PR.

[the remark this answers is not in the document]
AlessandroPomponio: Ah, great, we should open an issue about it

orchestrator/schema/property.py
```diff
@@ -29,6 +29,81 @@ class NonMeasuredPropertyTypeEnum(str, enum.Enum):
...
+
+    model_config = ConfigDict(frozen=True)
+
+
+class ConcretePropertyDescriptor(PropertyDescriptor):
+
+    propertyType: MeasuredPropertyTypeEnum = pydantic.Field(
+        default=MeasuredPropertyTypeEnum.MEASURED_PROPERTY_TYPE
+    )
+    abstractProperty: AbstractPropertyDescriptor | None = None
```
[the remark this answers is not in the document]
AlessandroPomponio: It'd probably make it a bit clearer too

orchestrator/schema/entity.py
```diff
@@ -104,15 +113,16 @@ def properties(self) -> list[ObservedProperty | ConstitutiveProperty]:
 
     @classmethod
     def identifier_from_property_values(
```
AlessandroPomponio, line 115: Should this method be renamed to `identifier_from_constitutive_property_values`?

orchestrator/schema/entity.py
```diff
@@ -74,15 +83,15 @@ class Entity(pydantic.BaseModel):
...
+    def propertyValues(self) -> list[ConstitutivePropertyValue | ObservedPropertyValue]:
         v = []
         for result in self.measurement_results:
             v.extend(result.measurements)
         v.extend(self.constitutive_property_values)
         return v
 
     @property
-    def properties(self) -> list[ObservedProperty | ConstitutiveProperty]:
+    def properties(self) -> list[ObservedProperty | ConstitutivePropertyDescriptor]:
```
AlessandroPomponio, line 94: Are these observed property descriptors or is it a mix of two things? If it's the former, maybe this should be called `properties_descriptors`

orchestrator/schema/entity.py
```diff
@@ -236,19 +246,19 @@ def observedProperties(self) -> list[ObservedProperty]:
...
-    def observedPropertyValues(self) -> list[PropertyValue]:
+    def observedPropertyValues(self) -> list[ObservedPropertyValue]:
 
         return [
             p for p in self.propertyValues if isinstance(p.property, ObservedProperty)
         ]
 
     @property
-    def constitutiveProperties(self):
+    def constitutiveProperties(self) -> list[ConstitutivePropertyDescriptor]:
```
AlessandroPomponio, line 256: Should it be `constitutive_properties_descriptors`?

orchestrator/schema/entity.py
```diff
@@ -278,7 +288,9 @@ def observedPropertiesFromExperimentReference(
             if v.experimentReference == experimentReference
         ]
 
-    def propertyValuesFromExperiment(self, experiment: Experiment):
+    def propertyValuesFromExperiment(
```
AlessandroPomponio, line 291: Maybe it's worth renaming this to `observed_property_values_from_experiment`

orchestrator/schema/entity.py
```diff
@@ -54,9 +61,11 @@ class Entity(pydantic.BaseModel):
     generatorid: str = pydantic.Field(
         "unk", description="The id of the generator that created this entity"
     )
-    constitutive_property_values: tuple[PropertyValue, ...] = pydantic.Field(
-        frozen=True,
-        description=" A list of PropertyValue objects giving values for constitutive properties",
+    constitutive_property_values: tuple[ConstitutivePropertyValue, ...] = (
+        pydantic.Field(
+            frozen=True,
+            description=" A list of PropertyValue objects giving values for constitutive properties",
```
AlessandroPomponio, line 67:
```suggestion
            description="A list of ConstitutivePropertyValue objects giving values for constitutive properties",
```

orchestrator/schema/observed_property.py
```diff
@@ -25,6 +29,16 @@ class ObservedProperty(pydantic.BaseModel):
     )
     model_config = ConfigDict(frozen=True)
 
+    @pydantic.field_validator("targetProperty", mode="before")
```
AlessandroPomponio, line 32:
```suggestion
    @pydantic.field_validator("targetProperty", mode="before")
    @classmethod
```

orchestrator/schema/property.py
```diff
@@ -29,6 +29,91 @@ class NonMeasuredPropertyTypeEnum(str, enum.Enum):
     CONSTITUTIVE_PROPERTY_TYPE = "CONSTITUTIVE_PROPERTY_TYPE"  # Properties whose values are immediately known when you define the entity
 
 
+class PropertyDescriptor(pydantic.BaseModel):
+    """A named property - no domain"""
+
+    identifier: str
+    model_config = ConfigDict(frozen=True, extra="forbid")
+
+    @pydantic.model_validator(mode="before")
```
AlessandroPomponio, line 38:
```suggestion
    @pydantic.model_validator(mode="before")
    @classmethod
```

orchestrator/schema/property.py
```diff
@@ -29,6 +29,91 @@ class NonMeasuredPropertyTypeEnum(str, enum.Enum):
...
+        A PropertyDescriptor will be equal to a Property if it has the same identifier.
+
+        Metadata is not included"""
+
+        try:
+            retval = self.identifier == other.identifier
+        except AttributeError:
+            retval = False
+
+        return retval
```
AlessandroPomponio, line 61: Maybe we could simplify this to

```python
return hasattr(other, "identifier") and self.identifier == other.identifier
```

orchestrator/schema/property.py
```diff
@@ -29,6 +29,91 @@ class NonMeasuredPropertyTypeEnum(str, enum.Enum):
...
+            p.breakable()
+
+
+class AbstractPropertyDescriptor(PropertyDescriptor):
+
+    propertyType: MeasuredPropertyTypeEnum = (
+        MeasuredPropertyTypeEnum.MEASURED_PROPERTY_TYPE
+    )
+
+    @pydantic.model_validator(mode="before")
```
AlessandroPomponio, line 78:
```suggestion
    @pydantic.model_validator(mode="before")
    @classmethod
```

orchestrator/schema/property.py
```diff
@@ -84,30 +177,56 @@ class AbstractProperty(Property):
...
-            )
+            try:
+                retval = (
+                    self.concretePropertyIdentifiers
+                    == other.concretePropertyIdentifiers
+                )
+            except AttributeError:
+                retval = False
 
         return retval
```
AlessandroPomponio, line 202: This could be something like

```python
return super().__eq__(other) and hasattr(other, "concretePropertyIdentifiers") and self.concretePropertyIdentifiers == other.concretePropertyIdentifiers
```

(this likely needs to be formatted and checked)

AlessandroPomponio (commented):
A few naming suggestions

orchestrator/schema/entity.py
michael-johnston: I don't think it necessary to distinguish.

That is, the information consumed by all callers of this method hasn't changed.

orchestrator/schema/entity.py
michael-johnston: I'm not sure we need, and also we would should then change every other method name.

In general I don't think the names of the methods have to be == the exact types returned, as that information is already present.

orchestrator/schema/property.py
michael-johnston, line 62:
```suggestion
        return hasattr(other, "identifier") and self.identifier == other.identifier
```

orchestrator/schema/property.py
michael-johnston: Sounds good

michael-johnston:
@VassilisVassiliadis A reminder if you could check the SFTTrainer code and approve the PR if changes look fine.

VassilisVassiliadis (approved):
LGTM, I also tested the SFTTrainer actuator and it worked fine.

PR #69, opened 2025-09-23
docs: update metastore query docs

michael-johnston:
This PR
* adds a dedicated section on how to query the resources in the metastore to the metastore page. 
     * This section moves and builds on the previous documentation under CLI. 
* adds doc-strings to the equivalent SQLResourceStore methods

TODO:
- Clarify string quoting requirements for searching for strings in lists
- Clarify JSON path behaviour of `[N]` for array access - is it 0 offset or 1 offset. 
- Clarify if there is additional processing of the strings passed to `-q` between ado cli and the SQLResourceStore call, beyond adding `$.`


website/docs/resources/metastore.md
```diff
@@ -149,3 +152,232 @@ a new local database will be created.
...
+    ```
+
+Resources are returned if they **_contain_ the candidate value at the specified
+path**. Containment is rigorously defined in
+[MySQL's JSON_CONTAINS documentation](https://dev.mysql.com/doc/refman/8.4/en/json-search-functions.html#function_json-contains),
+but for ease of use we distill the content here.
+
+##### The path points to a scalar
+
+Resources are returned only if the candidate is a scalar of the **same type**
```
michael-johnston, line 299:
```suggestion
The value at the key-path matches the candidate, if the candidate is a scalar of the **same type**
```

website/docs/resources/metastore.md
```diff
@@ -149,3 +152,232 @@ a new local database will be created.
...
+
+- ✅ `ado get operations -q 'config.operation.parameters.batchSize=1'` (1 == 1)
+- ✅ `ado get operations -q 'config.operation.parameters.batchSize=1.0'` (1.0 ==
+  1, floats and ints count as same type)
+- ❌ `ado get operations -q 'config.operation.parameters.batchSize="1"'` ("1" !=
+  1, type mismatch)
+
+##### The path points to an array
+
+**If the candidate is an array**, resources are returned only if the array the
```
michael-johnston, line 310:
```suggestion
**If the candidate is an array**,  the value at the key-path matches the candidate
```

website/docs/resources/metastore.md
```diff
@@ -149,3 +152,232 @@ a new local database will be created.
...
+
+- ✅
+  `ado get operation -q 'status=[{"event": "finished", "exit_state": "success"}]'`
+  (all keys are contained)
+- ❌
+  `ado get operation -q 'status=[{"exit_state": "success"}, {"exit_state": "failure"}]'`
+  (no operation can have both a `success` and a `failure` exit state)
+
+**If the candidate is not an array**, resources are returned only if the array
+the path points to contains the candidate.
```
michael-johnston, line 321:
```suggestion
**If the candidate is not an array**, resources are returned only if the array
the key-path points to contains the candidate.
```

website/docs/resources/metastore.md
```diff
@@ -149,3 +152,232 @@ a new local database will be created.
...
+
+- ✅ `ado get operation -q 'status={"exit_state": "success"}'` (status has an
+  element that contains the key `exit_state` and value `success`)
+- ✅ `ado get operation -q 'config.spaces=space-3904a4-96dd91'` (the string is
+  part of the array)
+- ❌ `ado get operation -q 'status={"event": "fake"}'` (no element of the
+  `status` array has the value `fake` for the `event` key)
+
+##### The path points to an object
```
michael-johnston, line 330:
```suggestion
##### The path points to an object (dictionary)
```

I at least would not understand that this is referring to a dictionary-like data-structure. I don't know if dictionary is very python specific but I find object confusing.

website/docs/resources/metastore.md
```diff
@@ -149,3 +152,232 @@ a new local database will be created.
...
+
+Resources are returned only if the candidate is an object that has **- for each
+key and related value from the candidate - the same key and the same related
+value**.
```
michael-johnston, line 334:
```suggestion
The value at the key-path matches if **each
key:value pair in the candidate** are contained in it
```

website/docs/resources/metastore.md
```diff
@@ -149,3 +152,232 @@ a new local database will be created.
...
+
+<!-- markdownlint-disable line-length -->
+- ✅
+  `ado get operation -q 'config.operation.parameters={"batchSize": 1, "samplerConfig": {"mode": "random"}}'`
+- ❌
+  `ado get operation -q 'config.operation.parameters={"batchSize": 1, "samplerConfig": {"mode": "smart"}}'`
+  (the operation's `samplerConfig.mode` is `random`)
+<!-- markdownlint-enable line-length -->
+
```
michael-johnston, line 343: I think then there should be a Details or Advanced Use section where there is
- Reference to MySQL JSON Path Syntax
- Clarification on what MySQL JSON Path syntax are/are not supported and what is/is not required c.f. the rooting `$.` element
- Reference to JSON_CONTAINS

Some of these could be tips or notes

website/docs/resources/metastore.md
```diff
@@ -149,3 +152,232 @@ a new local database will be created.
...
+    Make sure to properly quote JSON values when appropriate, so that they are
+    parsed correctly. This is particularly important when dealing with arrays or
+    objects.
+
+Where:
+
+- **Path** is a dot-separated expression that is backed by
+  [MySQL JSON Path](https://dev.mysql.com/doc/refman/8.4/en/json.html#json-path-syntax).
+  **NOTE**: single (\*) and double-asterisk (\*\*) wildcards are not supported.
+- **Candidate** is a valid JSON value (dictionary, array, string, number,
```
michael-johnston, line 247: I just noticed here we use dictionary and below we use object so we should harmonise

michael-johnston (commented):
Looking good. 

I suggest some changes to ease the reader into it without immediately directing them to external documentation or introducing many rules and qualifications. 

I think in general we

website/docs/resources/metastore.md
AlessandroPomponio:
```suggestion
**If the candidate is not an array**, the array at the specified path will match
if it contains the candidate.
```

website/docs/resources/metastore.md
```diff
@@ -149,3 +152,239 @@ a new local database will be created.
...
+
+If the candidate is a scalar (a string, a number, or a boolean), the value at
+the specified path will match if it is of the **same type** (treating integers
+and floats as equivalent) and has the **same value**.
+
+- ✅ `ado get operations -q 'config.operation.parameters.batchSize=1'` (1 == 1)
+- ✅ `ado get operations -q 'config.operation.parameters.batchSize=1.0'` (1.0 ==
+  1, floats and ints count as same type)
+- ❌ `ado get operations -q 'config.operation.parameters.batchSize="1"'` ("1" !=
+  1, type mismatch)
```
VassilisVassiliadis, line 308: Couple of comments here, we don't cover the case for `boolean` scalars. Those are somewhat special because the value is case sensitive and it has to be either `true` or `false` otherwise it gets intepreted as a string.

For example,  this returns results

>  ado get operations -q 'config.operation.parameters.singleMeasurement=true'

However, this doesn't:

>  ado get operations -q 'config.operation.parameters.singleMeasurement=true'

VassilisVassiliadis: Also booleans are special in a different way. 

Boolean properties (either in experiment `parameterization` or in the entitySpace of a discoverySpace) get rewritten by ado into `1` and `0` values.

This means that this search:

> ado get spaces -q 'config.experiments={"experiments":{"parameterization":[{"property":{"identifier":"bf16"},"value":true}]}}'


returns nothing but this one does match spaces:

> ado get spaces -q 'config.experiments={"experiments":{"parameterization":[{"property":{"identifier":"bf16"},"value":1}]}}'

As a user of `ado` the spaces I create look like this:

```yaml
sampleStoreIdentifier: "972034"

experiments:
- experimentIdentifier: finetune_full_benchmark-v1.0.0
  actuatorIdentifier: SFTTrainer
  parameterization:
    - property:
        identifier: fms_hf_tuning_version
      value: "2.8.2"
    - property:
       identifier: bf16
      value: True

entitySpace:
  - identifier: "model_name"
    propertyDomain:
      values: [ "granite-3.1-2b" ]
  - identifier: "number_gpus"
    propertyDomain:
      values: [ 1, 2, 4, 8]
  - identifier: "model_max_length"
    propertyDomain:
      values: [ 512, 1024, 2048  ]
  - identifier: "batch_size"
    propertyDomain:
      values: [ 1, 2, 4, 8, 16, 32, 64, 128, 256, 512 ]
  - identifier: "gpu_model"
    propertyDomain:
      values: [ 'NVIDIA-A100-SXM4-80GB' ]
```

so when I think about finding spaces in which `bf16` was `true` I will use

> ado get spaces -q 'config.experiments={"experiments":{"parameterization":[{"property":{"identifier":"bf16"},"value":true}]}}'

and the query will not yield any results.

AlessandroPomponio: > Couple of comments here, we don't cover the case for `boolean` scalars. Those are somewhat special because the value is case sensitive and it has to be either `true` or `false` otherwise it gets intepreted as a string.
> 
> For example, this returns results
> 
> > ado get operations -q 'config.operation.parameters.singleMeasurement=true'
> 
> However, this doesn't:
> 
> > ado get operations -q 'config.operation.parameters.singleMeasurement=true'

I assume you mean `True` in one of the two. I'll add a note.

As defined on https://www.json.org/json-en.html, boolean values are `true` and `false`, not any other value/capitalization.

> A value can be a string in double quotes, or a number, or true or false or null, or an object or an array. These structures can be nested.

AlessandroPomponio: https://github.com/IBM/ado/pull/69/commits/e03faddc2589da64670d74a76633567e5400691d and https://github.com/IBM/ado/pull/69/commits/e083fed8287b98c1d02e8746b041217843645ab1 should hopefully address your points

PR #89, opened 2025-10-02
fix: minor issues

michael-johnston:
- `ado show entities operation` did not allow `property-format target` even though it could support it
- `ado show entities space --include matching` returned matching entities with no measurements with `property-format observed` but not with `property-format target`. Both now do not return matching entities with no matching measurements.


AlessandroPomponio (approved):
LGTM thanks

PR #157, opened 2025-10-21
feat: add default sample store and --use-default-sample-store

### LINKED ISSUE #156: feat: create a default sample store on first initialization and add support for --use-default-sample-store
## Issue

We want to remove friction that slows people down when getting started with ado. In all examples people need to start off with a sample store, either creating one manually and using `--set sampleStoreIdentifier=ID` when creating a space, or using `--new-sample-store`.

This can lead to a lot of sample stores being created when blindly following the documentation. In many cases, a simple, "default" sample store can be used for these examples.

We should create it automatically:
- On the first ado initalization (after creating the local context)
- When people use the `--use-default-sample-store` flag in `ado create space`

AlessandroPomponio:
## Context

Resolves #156 

### Added
- Added a new CLI option `--use-default-sample-store` (alias for `--set sampleStoreIdentifier=default`) for the `ado create` command when creating spaces. This flag is only available for the *space* resource type. (`orchestrator/cli/commands/create.py`)
- In the discovery space creation logic (`orchestrator/cli/resources/discovery_space/create.py`):
  - When `use_default_sample_store` is true, set `space_configuration.sampleStoreIdentifier = "default"`.
  - Emit an informational message if the default sample store is requested during dry‑run.
  - Add logic to check if the “default” sample store resource already exists; if not, create it in the SQL store.
- In the command‑parameters model (`orchestrator/cli/models/parameters.py`), added new field `use_default_sample_store: bool`.
- In documentation (`website/docs/getting-started/ado.md`): Updated the `ado create` syntax to include the `--use-default-sample-store` flag.
- In documentation (`website/docs/resources/sample-stores.md`): Added a new section “The default samplestore” outlining how to use or reference the default sample store via:
  1. YAML property `sampleStoreIdentifier: default`
  2. CLI flag `--use-default-sample-store`
  3. CLI flag `--set sampleStoreIdentifier=default`

### Documentation
- In `ado.md` (Getting Started): Enhanced the explanation for the `--use-default-sample-store` flag, clarifying its alias behavior.
- In `sample-stores.md`: Revised the tip block for discovery spaces to properly show the line-break and formatting changes, and added a full description for the default sample store workflow.

### Implementation Details
- Creating a discovery space with `--use-default-sample-store` now triggers automatic creation of a `default` sample store if one doesn’t already exist.

---

disclaimer: changelog generated by copilot

### DIFF
--- a/website/docs/resources/sample-stores.md
+++ b/website/docs/resources/sample-stores.md
@@ -69,13 +69,12 @@ SQLSampleStore in the current context.
 
 ## The default samplestore
 
-When experimenting or when strict separation between sample stores is not
-required, you can create spaces using the **default** sample store (whose
-identifier is `default`). This sample store is created automatically when
-requested, allowing for quick setup without needing to create a sample store
-explicitly.
+`ado` provides a **default** `samplestore` (whose identifier is `default`) per
+project, removing the need to create one explicitly unless necessary. This
+`samplestore` is created **automatically** when it is first required.
 
-There are three ways to use the default sample store:
+There are three ways to use the default `samplestore` - each will create it, if
+it doesn't already exist.
 
 1. **Referencing it in the space configuration** by setting the
    `sampleStoreIdentifier` to `default` in the space YAML:


michael-johnston (changes requested):
We need to update the `metastore` and `samplestore` docs to explain default sample store concept and how to create one, how to identify it etc.

AlessandroPomponio:
I have actually changed the way it works to make it a bit more flexible. Now it will be used in three cases:

1. The user uses `--use-default-sample-store`
2. The `sampleStoreIdentifier` in the space configuration is set to `default`
3. The user uses `--set sampleStoreIdentifier=default`

PR #166, opened 2025-10-27
feat: support --use-latest flag in ado show commands

### LINKED ISSUE #165: feat: support --use-latest flag in ado show commands
## Issue

Most of the time people want to see data (a summary, a description, entities, etc) for the latest resource they have created. Since we have the `--use-latest` flag in `ado create`, we can port it over to the `show` commands as well. Since here users cannot reference other resources, instead of it being an option that requires a resource type, it should just be a flag.

AlessandroPomponio:
## Context

Resolves #165.

### ✨ Features
- Introduced a new `--use-latest` flag to the following CLI commands:
  - `ado show details`
  - `ado show entities`
  - `ado show related`
  - `ado show requests`
  - `ado show results`
  - `ado show summary`

### 🛠 Enhancements
- Added logic to resolve the latest resource identifier when `--use-latest` is used and no explicit ID is provided.
- Improved error handling and user feedback when no latest resource is available.
- Updated help texts and command examples to reflect the new flag usage.

### 📁 New Files
- `orchestrator/cli/utils/generic/common.py`: Contains shared utility `get_effective_resource_id()` for resolving resource IDs.

### 📚 Documentation
- Updated `website/docs/getting-started/ado.md`:
  - Added syntax and examples for using `--use-latest` across supported commands.
  - Clarified behavior when both `RESOURCE_ID` and `--use-latest` are provided.

### 🧪 CLI UX Improvements
- Enhanced console messages to inform users when `--use-latest` is used or ignored.
- Provided warnings when both `RESOURCE_ID` and `--use-latest` are specified, prioritizing explicit IDs.

### 🔧 Internal Refactoring
- Refactored CLI command implementations to support optional `resource_id` and `use_latest` parameters.
- Centralized resource ID resolution logic for consistency and reuse.

---

Disclaimer: changelog generated by copilot

### DIFF
--- a/website/docs/getting-started/ado.md
+++ b/website/docs/getting-started/ado.md
@@ -553,9 +553,9 @@ Where:
 
 - `RESOURCE_ID` is the unique identifier of the resource you want to see details
   for.
-- `--use-latest` will use the identifier of the latest resource of RESOURCE_TYPE
-  in the command. It requires having created a resource of RESOURCE_TYPE with
-  `ado create`. It is ignored if a RESOURCE_ID is provided.
+- `--use-latest` will use the identifier of the latest (i.e. most recent)
+  resource of RESOURCE_TYPE created locally. It is not context aware. It is
+  ignored if a RESOURCE_ID is provided.
 
 ##### Examples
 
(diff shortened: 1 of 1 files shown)


website/docs/getting-started/ado.md
```diff
@@ -553,6 +553,9 @@ Where:
 
 - `RESOURCE_ID` is the unique identifier of the resource you want to see details
   for.
+- `--use-latest` will use the identifier of the latest resource of RESOURCE_TYPE
+  in the command. It requires having created a resource of RESOURCE_TYPE with
+  `ado create`. It is ignored if a RESOURCE_ID is provided.
```
michael-johnston:
```suggestion
- `--use-latest` will use the identifier of the latest, i.e. most recent, resource of RESOURCE_TYPE created locally.  It is not context aware. It is ignored if a RESOURCE_ID is provided.
```

michael-johnston: I think the help for the others could also be updated. The suggested change
- Ties latest and create together in the first sentence - latest only makes sense with create.
- Calls out the creation must have happened locally - not by someone else remotely in same project
- Notes its not context aware.

michael-johnston (commented):
I have a suggestion about the help string

PR #215, opened 2025-11-12
build: create test dependency group in pyproject

michael-johnston:
Instead of having to install all the deps for testing one-by-one can now do

```
uv sync --group test
```

The test group is
- ado-sfttrainer
- ado-ray-tune
- optimization_test_functions
- pfas-generative-models/custom_experiments
- profile_space
- anomalous_series

## Context

-  All packages in the group are already required to install and execute as part of the tests in tox
      - They also include packages required for our examples to work
      - Therefore: we are already implicitly dependent on them being compatible with ado-core
- The test group is optional
     - They are not installed via `uv sync` or `uv pip install`
- The dependencies of the test group will inform what versions of packages are in the lock file and installed via uv sync
     - i.e. when you install ado-core the package versions will be compatible with a future install of ado-ray-tune etc. 
     - However we want this c.f. the first point  - the alternative means we could have the situation where our examples
        will not work on a release

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -6,7 +6,7 @@ i.e. the top directory of the `ado` repo.
 We recommend using `uv sync` to install all test requirements:
 
 ```commandline
-uv sync --group tests
+uv sync --group test
 ```
 
 Note: This will also remove all packages not required for testing and sync


michael-johnston:
> These need to be declared as optional dependencies and we first need to see what kind of consequence this has on the build process.

I've update the issue with how this will influence the build process, also w.r.t what we already do.

AlessandroPomponio (changes requested):
The changes here should be reflected in the tox file as well. 

This means:
- Removing all the `uv pip install` commands in there
- Adding the `tests` dependency group (~~should it maybe be called just `test`?~~ -- it is, it's a typo in the file) to the `dependency_groups` section

I have also noticed that the `vllm_performance` actuator is not in the tests group: do we want it there?

PR #229, opened 2025-11-20
feat(group_samplers): improve performance for group generator sampler type

christian-pinto:
Currently, the group samplers generate all the entities for a space up-front, before they are consumed via an iterator. Fo large spaces this means to: 1) long times before the first entity is sampled, 2) waste of memory as all the entities are generated upfront eve if they are not going to be sampled.

This PR generates entities only when they are fetched via an iterator.

For a space with 64k entities, with the current implementation I measure in second

```
Time to first entity 1872.631384583001
Done getting 46080 entities 1872.807050375006
```

the code in this PR 

```
Time to first entity 14.360848165990319
Done getting 46080 entities 1834.0484777079982
```

### DIFF
--- a/orchestrator/core/discoveryspace/space.py
+++ b/orchestrator/core/discoveryspace/space.py
@@ -667,7 +667,7 @@ class DiscoverySpace:
                     p for p in requestedProperties if p not in definedProperties
                 ]
 
-                if len(filtered) > 1:
+                if len(filtered) > 0:
                     raise ValueError(
                         f"Requested match against constitutive properties not in entity space definition: {filtered}"
                     )
@@ -705,6 +705,21 @@ class DiscoverySpace:
             Raise ValueError if the point is not in the discovery space
         """
 
+        property_identifiers = {
+            cp.identifier for cp in self.entitySpace.constitutiveProperties
+        }
+        point_identifiers = set(point.keys())
+        if diff := point_identifiers - property_identifiers:
+            raise ValueError(
+                f"Point {point} is not in space. It has values for additional properties, {diff}"
+            )
+
+        if diff := property_identifiers - point_identifiers:
+            raise ValueError(
+                f"Point {point} is not in space. It is missing values for properties, {diff}"
+            )
+
+        # Note if point contains additional properties this will just ignore them
         property_values = constitutive_property_values_from_point(
             point=point, properties=self.entitySpace.constitutiveProperties
         )


orchestrator/core/discoveryspace/group_samplers.py
```diff
@@ -151,7 +173,9 @@ def _random_iterator(
 
 
 def _sequential_group_iterator(
-    generator: Generator[list[Entity], None, None], batchsize: int
+    generator: Generator[list[dict], None, None],
+    discoverySpace: DiscoverySpace,
```
AlessandroPomponio: Let's use snake_case

AlessandroPomponio (changes requested):

orchestrator/core/discoveryspace/group_samplers.py
```diff
@@ -485,12 +497,14 @@ def random_iterator() -> AsyncGenerator[list[Entity], None]:
         return await iterator_closure(remoteDiscoverySpace)
 
     def entityIterator(
-        self, discoverySpace: DiscoverySpace, batchsize=1
+        self, discovery_space: DiscoverySpace, batchsize=1
```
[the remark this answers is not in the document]
christian-pinto: Thant's what I did, but I have missed this one. In all the others I have preserved the methods signatures when we inherit from other classes.

tests/core/test_group_samplers.py
```diff
@@ -215,10 +226,20 @@ async def test_group_sampler_remote(
...
+            for e in group
         }
         cpu_value = {
-            e.valueForConstitutivePropertyIdentifier("cpu_family").value for e in group
+            (
+                e["cpu_family"]
+                if isinstance(e, dict)
+                else e.valueForConstitutivePropertyIdentifier("cpu_family").value
+            )
+            for e in group
```
michael-johnston: > However, people might be using the entityGroup iterator directly expecting for it to return a list of entities, as per its original signature. Someone might use it directly expecting a list of entities out of that iterato

To check I understand - your saying that now entityGroupIterator returns groups of Entities (as original)?

If so I don't understand the test changes here

christian-pinto: Yes, and that's just a refuse from a previous commit. Let me remove it.

orchestrator/core/discoveryspace/group_samplers.py
```diff
@@ -20,24 +20,54 @@
...
+    matching_entities = discovery_space.matchingEntities()
+    points = []
+    for entity in matching_entities:
+        point = {
+            v.property.identifier: v.value for v in entity.constitutive_property_values
         }
-    )
+        # we save the entity identifier to retrieve the entity from the store at a later stage,
+        # assuming they are cached locally
+        point["entity_identifier"] = entity.identifier
```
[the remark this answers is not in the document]
christian-pinto: Removed it in the last commit. However, no exception was raised and all the tests were successful.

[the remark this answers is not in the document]
michael-johnston: That's strange - I will have a look at the tests because we must be missing something.

michael-johnston:
@christian-pinto I found the issue. @AlessandroPomponio can you give this a last review?

orchestrator/core/discoveryspace/group_samplers.py
```diff
@@ -20,24 +20,51 @@
...
+    matching_entities = discovery_space.matchingEntities()
+    points = []
+    for entity in matching_entities:
+        point = {
+            v.property.identifier: v.value for v in entity.constitutive_property_values
         }
-    )
+        points.append(point)
+
+    return points
```
AlessandroPomponio:
```suggestion
    return [
        {v.property.identifier: v.value for v in entity.constitutive_property_values}
        for entity in discovery_space.matchingEntities()
    ]
```

PR #232, opened 2025-11-21
feat(core): add vector property domain

### LINKED ISSUE #243: feat: support vector property domain
 Some properties have vector values. This is not supported by PropertyDomain (all domains are assumed to be scalars). This is problematic for constitutive properties as we cannot currently support a constitutive property that is a vector as there is no support in PropertyDomain for it. Each element of the vector has to be declared as a separate constitutive property, meaning they are not grouped in addition to not being scalable for writing. 

**Describe the solution you'd like**

A suggestion is to have a property domain for vectors like 

```yaml
propertyDomain:
    - groupDomain: #Domain for all elements in the group
    - numberElements: #The number of elements in this group. 
    - groupIds: #A list of the domain ids: Optional - otherwise based on element
```

michael-johnston:
A vector property domain is a domain consisting of multiple numeric elements all with same property domain e.g. a vector could have 100 elements each of which has an integer domain with values from 0 to 100. 

It provides a way to succinctly describe such a domain.


DRL-NextGen:
No vulnerabilities found.

PR #240, opened 2025-11-21
fix(vllm_performance): example entity space was incompatible with measurement space

michael-johnston:


AlessandroPomponio (approved):
LGTM thanks

PR #249, opened 2025-11-25
fix(vllm_performance): Avoid starvation of measurements requests

### LINKED ISSUE #247: bug: experiment request starvation in vllm_performance.test-deployment-v1 with random walk
## Issue Description

When the number of entities in the random walk initial batch is > max number of environments in vllm_performance the test-deployment-v1 experiments for the excess entities in the initial batch (initial batch size - max number environments) are very likely not to execute until the very end of the random walk. 

This is because of the environment wait/assignment code https://github.com/IBM/ado/blob/6edcc47d503adcecb7d1d5b11d09681ad51536f2/plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py#L119-L127

When a request finds there are no environments it sleeps for 5s. However the time from the finish of an experiment that is executing to firing a new experiment (continuous batching) is much less than 5. Hence in nearly all cases when the environment becomes available the waiting request is asleep and the next requests comes in a gets the free slot. 

### How to reproduce

1. Create a vllm performance actuator configuration with default max_environments (1)
2. Create a space using test-deployment-v1
3. Create a random walk with batch size 2
4. Launch operation
5. You will see the second request created does not get executed until the end of the walk

## Expected behaviour

Requests should queue on environment i.e. first to be waiting should be the first to be served when an environment becomes available

### Python/ado/system info

vllm-performance 1.2.4

christian-pinto:
This PR fixes #247 by introducing a queuing mechanisms for measurement requests that do not find an environment ready to use. Requests are given access to environments in the order of their  arrival.

### DIFF
--- a/plugins/actuators/vllm_performance/ado_actuators/vllm_performance/env_manager.py
+++ b/plugins/actuators/vllm_performance/ado_actuators/vllm_performance/env_manager.py
@@ -1,6 +1,7 @@
 # Copyright (c) IBM Corporation
 # SPDX-License-Identifier: MIT
 
+import asyncio
 import copy
 import logging
 import time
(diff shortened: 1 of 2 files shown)


michael-johnston (approved):
Checked at it works as advertised 👍

PR #263, opened 2025-11-28
feat(core): allow custom_experiments to execute with or without Ray

michael-johnston:
This PR adds fields to custom_experiment decorator allowing developer to specify if they want the custom_experiment to be executed as remote ray task and to provide ray remote options e.g. runtime env, if so.

### DIFF
--- a/orchestrator/modules/operators/_explore_orchestration.py
+++ b/orchestrator/modules/operators/_explore_orchestration.py
@@ -57,6 +57,8 @@ def graceful_explore_operation_shutdown(
     if not orchestrator.modules.operators._cleanup.shutdown:
         import time
 
+        from rich.console import Console
+
         moduleLog.info("Shutting down gracefully")
 
         orchestrator.modules.operators._cleanup.shutdown = True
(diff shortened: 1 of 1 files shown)


orchestrator/modules/actuators/custom_experiments.py
```diff
@@ -724,19 +737,38 @@ async def submit(
...
+        use_ray = getattr(fn, "_use_ray", True)
+        ray_options_model = getattr(fn, "_ray_options", None)
+        if use_ray:
+            remote_kwargs = (
+                ray_options_model.model_dump(exclude_none=True)
+                if getattr(fn, "_ray_options", None)
+                else {}
+            )
+            # Dispatch as Ray task. Pass ray options if present.
+            ray.remote(custom_experiment_executor, **remote_kwargs).remote(
```
AlessandroPomponio: Just double checking that the double `remote` call is correct

website/docs/actuators/creating-custom-experiments.md
```diff
@@ -284,6 +284,66 @@ This is illustrated in the above example.
...
+experiment can execute at a time.
+In addition, once an instance of the custom experiment
+is started no _other_ custom experiments can run
+until it is finished
+
+#### `ray_options`
+
+> [!NOTE]
+>
+> If `use_ray` is False the value of this parameter is ignored
```
AlessandroPomponio:
```suggestion
> If `use_ray` is False, the value of this parameter is ignored
```

orchestrator/modules/actuators/custom_experiments.py
michael-johnston: Yes one wraps the function in `ray.remote` decorator and the second is calling the `remote()` method on the decorated function.

website/docs/actuators/creating-custom-experiments.md
```diff
@@ -284,6 +284,66 @@ This is illustrated in the above example.
...
+If `True` (the default), the custom experiment will be run
+as a Ray remote function.
+This allows multiple instances to execute in parallel
+across a Ray cluster.
+
+If `False`, only one instance of the custom
+experiment can execute at a time.
+In addition, once an instance of the custom experiment
+is started no _other_ custom experiments can run
+until it is finished
```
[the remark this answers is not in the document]
michael-johnston: > Other instances of the same experiment can execute (which feels in contrast with what's said at the start of the paragraph)
The sentence starts with `In addition,` which means this is in addition to the first restriction. 

> No custom experiment can run at all while this is running

It's not exactly this which is why I can't use the other phrasing. The reason is, if the "other" experiment is submitted first and it uses ray it will be started remotely and immediately another experiment can be submitted.  Then if the next experiment does not use ray, it will start synchronously and nothing else can be submitted until it finished. However the first experiment may still be running - so there can be two+ custom_experiments executing at same time.

The key is that when a `use_ray=False` one is submitted it "blocks the queue"  until it is finished.

michael-johnston:
@AlessandroPomponio I found a small issue so don't merge this until I push another change which will be a bit later.

PR #277, opened 2025-12-03
docs(website): clarify wheel build output location

VassilisVassiliadis:
After a recent update to uv, running `uv build` inside nested Python packages places the wheel in the top-level dist directory.

Previously, wheels were generated in the dist directory located within the specific package that was being built.


AlessandroPomponio (approved):
LGTM thanks

PR #278, opened 2025-12-03
fix(custom_experiments): required_properties parameter of decorator was required instead of optional

michael-johnston:
Also add a test that the website example works.


AlessandroPomponio (approved):
LGTM thanks

VassilisVassiliadis (dismissed):

Issue #200, opened 2025-11-06
bug: nested operations cause the outer operation to think a signal has been sent


michael-johnston:
## Issue Description

When an operation launches another operation, and both finish successfully, you will find that ado records the inner operation as having been successful and the outer operation as having "exited normally but a signal was sent" and the overall status set to error 

### How to reproduce

To reproduce currently you need to use the trim operator. Details TBA

## Expected behaviour

Both operations should have status successful

### Python/ado/system info

Please include the output of:

python 12.7
ado: 1.1.1.dev7
os: Mac

PR #281, opened 2025-12-03
refactor(core): separate cleanup logic from signal handling and fix nested-operation shutdown

michael-johnston:
Operation cleanup logic was mixed together (a single function for cleaning up everything was required/was different for different scenarios) and also mixed with signal handling (a shutdown could be successful operation or a signal).

This made it impossible to correctly handle operation cleanup when there were nested operations (see #200)

In this PR
- cleanup code required by different operations/functions is separated
- A single signal handler exists that is generic
   - operations register their cleanup requirements with the handler so they are cleaned-up
- shutdown global now only indicates if a signal was received NOT if an operation finished
- resource cleaner is per-operation
- Actuators can access operation namespace via MeasurementQueue

New behaviour: 

- When an explore operation finishes it cleans up the ray actors it has created including its own ResourceCelaner. It does not
        - shutdown ray
        - set shutdown global (which is gone)
- When orchestrate() (called via ado create operation)) exit it cleans up ray
- When general operation finishes there is no set cleanup as there are no ray resources created. It does not:
         - shutdown ray
        - set shutdown global (which is gone)
- If operations are nested there should no longer be issues with shutdown being set or resources deleted
- If SIGTERM is received the cleanup methods of all active operations will be called via the signal handler
- If SIGINT is received during an operation, the operation catches it, cleans its self and raises InterruptedOperationError operation id
        - previously nothing was raised and the operation exited via the normal route
        
 Details:
 - Ray actors in an operation need to access the ResourceCleaner in the operation ray namespace. For this they need to be passed the ray namespace name: there is no ray method for this information to be retrieved from within the actor.

### DIFF
--- a/orchestrator/modules/operators/_explore_orchestration.py
+++ b/orchestrator/modules/operators/_explore_orchestration.py
@@ -55,7 +55,7 @@ def graceful_explore_operation_shutdown(
     timeout=60,
 ):
 
-    from rich.console import Console
+    from rich.status import Status
 
     moduleLog.info(f"Shutting down operation {identifier} gracefully")
 
@@ -66,9 +66,7 @@ def graceful_explore_operation_shutdown(
     # 3. Send graceful __ray_terminate__ to metric_server, operation and actuators
 
     # This should not return until the metric server has processed all updates.
-
-    console = Console()
-    with console.status(
+    with Status(
         f"Shutdown ({identifier}) - waiting on all samples to be stored", spinner="dots"
     ) as status:
 
--- a/orchestrator/modules/operators/orchestrate.py
+++ b/orchestrator/modules/operators/orchestrate.py
@@ -51,10 +51,9 @@ def graceful_orchestrate_shutdown():
 
     import time
 
-    from rich.console import Console
+    from rich.status import Status
 
-    console = Console()
-    with console.status("Shutdown - shutting down Ray", spinner="dots") as status:
+    with Status("Shutdown - shutting down Ray", spinner="dots") as status:
         ray.shutdown()
         status.update("Shutdown - waiting for logs to flush")
         moduleLog.info("Waiting for logs to flush ...")


orchestrator/modules/operators/_cleanup.py
```diff
@@ -2,56 +2,50 @@
...
+        if not shutdown_signal_received:
+            shutdown_signal_received = True
+            moduleLog.debug("Cleanup custom actors")
+            try:
+                cleaner_handle = ray.get_actor(name=CLEANER_ACTOR)
+                ray.get(cleaner_handle.cleanup.remote())
+                # deleting a cleaner actor. It is detached one, so has to be deleted explicitly
+                ray.kill(cleaner_handle)
+            except ValueError:
+                moduleLog.debug("No ray actor")
```
AlessandroPomponio:
```suggestion
                moduleLog.debug("No Ray cleaner actors were found")
```

orchestrator/modules/operators/_cleanup.py
```diff
@@ -2,56 +2,50 @@
...
 
 import ray
 from ray.actor import ActorHandle
 
-shutdown = False
+shutdown_signal_received = False
 CLEANER_ACTOR = "resource_cleaner"
 
 moduleLog = logging.getLogger("orchestration_cleanup")
+signal_cleanup_callbacks: dict[str, Callable[[], None]] = OrderedDict()
```
AlessandroPomponio: I'd call these `cleanup_callback_functions` so that it's clear that they're functions

orchestrator/modules/operators/_explore_orchestration.py
```diff
@@ -48,112 +50,72 @@
...
+    with console.status(
+        f"Shutdown ({identifier})- waiting on all samples to be stored", spinner="dots"
+    ) as status:
 
-            moduleLog.debug("Shutting down state")
-            promise = state.shutdown.remote()
-            ray.get(promise)
+        moduleLog.debug("Shutting down state")
+        promise = state.shutdown.remote()
+        ray.get(promise)
```
AlessandroPomponio: I guess this could be joined together

orchestrator/modules/operators/_explore_orchestration.py
```diff
@@ -48,112 +50,72 @@
...
 
-            moduleLog.debug("Cleanup custom actors")
-            try:
-                cleaner_handle = ray.get_actor(name=CLEANER_ACTOR)
-                ray.get(cleaner_handle.cleanup.remote())
-                # deleting a cleaner actor. It is detached one, so has to be deleted explicitly
-                ray.kill(cleaner_handle)
-            except Exception as e:
-                moduleLog.warning(f"Failed to cleanup custom actors {e}")
+        wait_graceful = [
```
AlessandroPomponio: I'd call these `terminating_actors`

orchestrator/modules/operators/orchestrate.py
```diff
@@ -42,6 +44,37 @@
...
+    from rich.console import Console
+
+    console = Console()
+    with console.status("Shutdown - cleanup", spinner="dots") as status:
+
+        moduleLog.debug("Cleanup custom actors")
+        try:
+            cleaner_handle = ray.get_actor(name=CLEANER_ACTOR)
+            ray.get(cleaner_handle.cleanup.remote())
+            # deleting a cleaner actor. It is detached one, so has to be deleted explicitly
```
AlessandroPomponio:
```suggestion
            # The cleaner actor is detached, so it requires explicit deletion
```

AlessandroPomponio (commented):
Just naming suggestions

orchestrator/modules/operators/_explore_orchestration.py
```diff
@@ -48,112 +46,69 @@
...
+        lookup = dict(zip(terminating_actors, actors))
 
-            actors = [operator]
-            actors.extend(actuators)
+        moduleLog.debug(f"Shutdown waiting on {lookup}")
+        moduleLog.debug(
+            f"Gracefully stopping actors - will wait {timeout} seconds  ..."
+        )
+        terminated, active = ray.wait(
+            ray_waitables=terminating_actors, num_returns=n_actors, timeout=60.0
```
AlessandroPomponio:
```suggestion
            ray_waitables=terminating_actors, num_returns=n_actors, timeout=timeout
```

orchestrator/modules/operators/setup.py
```diff
@@ -55,16 +54,17 @@ def load_secrets_from_env(vars_to_load, env_var_dict):
...
 ) -> dict[str, "ActuatorActor"]:
     """
+    Creates all the actuators required by discovery_space
+
     Params:
-        namespace: The namespace to set up in
-        config: Configuration of the orchestrator
-        queue: the update queue
+        discovery_space: The discovery space to create the actuators for
+        actuator_configurations: A set of (optional) configurations for actuators in the discoveryspace
```
AlessandroPomponio: These aren't ActuatorConfigurations anymore, they're identifiers (the variable name has changed as well)

orchestrator/modules/operators/_explore_orchestration.py
```diff
@@ -41,119 +38,90 @@
...
+        try:
+            cleaner_handle = ray.get_actor(name=CLEANER_ACTOR, namespace=namespace)
+            moduleLog.debug(f"Calling cleanup on {cleaner_handle}")
+            ray.get(cleaner_handle.cleanup.remote())
+            ray.kill(cleaner_handle)
+        except Exception as e:
+            moduleLog.warning(f"Failed to cleanup custom actors {e}")
+
+        status.update(
+            f"Shutdown ({identifier}) - waiting for actors to terminate (max 60s)"
```
AlessandroPomponio:
```suggestion
            f"Shutdown ({identifier}) - waiting for actors to terminate (max {timeout}s)"
```

orchestrator/modules/operators/_cleanup.py
```diff
@@ -2,56 +2,39 @@
...
-        graceful_operation_shutdown()
+        moduleLog.critical(f"Got signal {sig}")
+        global shutdown_signal_received
+        global cleanup_callback_functions
+
+        if shutdown_signal_received:
+            moduleLog.info("Graceful shutdown already completed")
+
+        shutdown_signal_received = True
+        moduleLog.warning("Calling cleanup callbacks")
```
michael-johnston:
```suggestion
        moduleLog.info("Calling cleanup callbacks")
```

orchestrator/modules/operators/orchestrate.py
```diff
@@ -146,7 +174,9 @@ def orchestrate(
                 )
             )  # type: OperationOutput
     except KeyboardInterrupt:
-        moduleLog.warning("Caught keyboard interrupt - initiating graceful shutdown")
+        moduleLog.warning(
+            "Caught keyboard interrupt at top level - initiating graceful shutdown"
```
michael-johnston:
```suggestion
            "Caught keyboard interrupt - initiating graceful shutdown"
```

orchestrator/modules/operators/setup.py
michael-johnston:
```suggestion
        actuator_configuration_identifiers: A set of (optional) identifiers of configurations for actuators in the discoveryspace
```

PR #345, opened 2026-01-07
docs(website): fix typo in "Target v observed property formats"

danielelotito:
"Target v observed property formats" to "Target vs observed property formats"


AlessandroPomponio (approved):
LGTM thanks

PR #379, opened 2026-01-12
fix(core): add relationships to interrupted nested operations 

michael-johnston:
This PR ensures that if a nested operation is interrupted by SIGINT that the parent operation adds the relationships to the metastore.

It also adds mechanisms to allow an operation to ensure all non-operation resources it has already created are stored in db on keyboard interrupt.

### DIFF
--- a/orchestrator/modules/operators/_orchestrate_core.py
+++ b/orchestrator/modules/operators/_orchestrate_core.py
@@ -122,7 +122,7 @@ def _run_operation_harness(
             message="Operation exited due to SIGINT propagated from nested operation",
         )
 
-        # Add the nested operation identifier to the list
+        # Record the identifier of the interrupted nested operation
         interrupted_nested_operation = error.operation_identifier
         if error.resources:
             # Create an OperationOutput to hold the resources created before interrupt
--- a/website/docs/operators/creating-operators.md
+++ b/website/docs/operators/creating-operators.md
@@ -230,21 +230,20 @@ return OperationOutput(resources=[DataContainerResource(config=data_container)])
 
 ### Storing returned resources
 
-All resources returned by the operation will automatically
-be stored in the project the operation was created in.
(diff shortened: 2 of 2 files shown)


danielelotito (dismissed):
This makes sense to me

orchestrator/modules/operators/_orchestrate_core.py
```diff
@@ -107,10 +108,35 @@ def _run_operation_harness(
...
+            f"during operation {operation_resource.identifier}."
+        )
+
+        operationStatus = OperationResourceStatus(
+            event=OperationResourceEventEnum.FINISHED,
+            exit_state=OperationExitStateEnum.ERROR,
+            message="Operation exited due to SIGINT propagated from nested operation",
+        )
+
+        # Add the nested operation identifier to the list
```
AlessandroPomponio: Not a list anymore

website/docs/operators/creating-operators.md
```diff
@@ -229,9 +226,25 @@ data_container = DataContainer(tabularData={"main_dataframe":tabular_data},
...
 
+### Storing returned resources
 
-```
+All resources returned by the operation will automatically
+be stored in the project the operation was created in.
+In addition, the relationships between the operation and the resources
+it creates are also automatically added.
+This means ```ado show related operation $OPERATIONID``` will list the
+resources the operation created.
```
AlessandroPomponio: Formatting + changing triple backtick to single backtick + OPERATIONID -> OPERATION_ID

```suggestion
All resources returned by the operation will automatically be stored in the
project the operation was created in. In addition, the relationships between the
operation and the resources it creates are also automatically added. This means
`ado show related operation $OPERATION_ID` will list the resources the operation
created.
```

AlessandroPomponio (approved):
LGTM thanks

PR #455, opened 2026-01-22
feat(core): allow specifying actuator in csv sample store

michael-johnston:
Adds  ability to specify an actuator when importing data from a CSV. This allows data obtained from elsewhere (maybe from ado) to appear as ado data (can be used as memoized data in a space with that experiment)

The actuator/experiment combination will be validated against existing actuators and if passed imported. 

Notes:
- Changes the schema of CSVSampleStoreDescriptor to simplify the code - upgrade path and warnings provided. 
- column name are no longer converted to lower case automatically. User must do this in YAML if desired. 
- importing data from parameterized experiments (experiments with default options overridden) is not supported .


website/docs/resources/sample-stores.md
```diff
@@ -213,6 +215,51 @@ copyFrom:
...
+For non-replay actuators `ado` will verify that:
+
+1. The specified actuator exists in the current instance
+2. The experiment exists in that actuator's catalog
+3. The CSV contains all required constitutive properties for the experiment
+
+If any validation fails, a detailed error message will indicate what's wrong.
+
+**Backward Compatibility**: If `actuatorIdentifier` is not specified for an
+experiment, it defaults to `"replay"`, maintaining full compatibility with
```
AlessandroPomponio:
```suggestion
experiment, it defaults to `replay"`, maintaining full compatibility with
```

AlessandroPomponio (changes requested):

PR #460, opened 2026-01-23
refactor: use pathlib instead of os.path

AlessandroPomponio:


AlessandroPomponio:
This became stale. We can do it another time

PR #472, opened 2026-01-28
docs(vllm_performance): update examples

michael-johnston:
- Update instances of old experiment names
- Include YAML for the examples directly from the example files


plugins/actuators/vllm_performance/yamls/random_walk_operation_grouped.yaml
```diff
@@ -16,7 +16,7 @@ operation:
     samplerConfig:
       mode: 'sequentialgrouped'
       samplerType: 'generator'
-      grouping: #These are properties entities compatible with vllm-performance-full
+      grouping: #These are entity properties compatible with vllm_performance.test-deployment-v1
```
christian-pinto:
```suggestion
      grouping: #These are entity properties compatible with vllm_performance.test-deployment-v1 and vllm_performance.test-deployment-guidellm-v1
```

christian-pinto (changes requested):
Just a minor change. Looks good otherwise

PR #474, opened 2026-01-28
refactor: use rich instead of IPython's pretty

### LINKED ISSUE #469: refactor: replace IPython's _repr_pretty with rich's __rich__ for describe commands

AlessandroPomponio:
### Changes
- Migrated from IPython's pretty printing to Rich library for all console output
- Replaced `_repr_pretty_()` methods with `__rich__()` methods
- Removed need for pandas import in describe commands
- Removes dependency on jupyter and transitive packages

Resolves #469 

---

#### Before
image
image

#### After
image
image

#### Before
image

#### After
image

#### Before

image

#### After
image

### DIFF
--- a/orchestrator/core/samplestore/sql.py
+++ b/orchestrator/core/samplestore/sql.py
@@ -112,9 +112,8 @@ class SQLSampleStore(ActiveSampleStore):
         from orchestrator.utilities.rich import get_rich_repr
 
         return Group(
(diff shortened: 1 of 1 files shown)


AlessandroPomponio:
I've updated the outputs - I have noticed that a bit of output might get truncated, I think this might be due to the renderables being nested. I have however not been able to understand how to fix it :\

PR #481, opened 2026-01-30
feat(core): improve sql sample store initialisation and allow for efficient refresh

### LINKED ISSUE #477: feat(core): improve sql sample store initialisation and allow for efficient refresh
## Problem

The current implementation of the `entities` property in `SQLSampleStore`
(`orchestrator/core/samplestore/sql.py`, lines 356-417) has several
inefficiencies and limitations:

### 1. No Refresh Mechanism

- Entities and measurements are cached at initialization (line 333-336)
- No way to refresh the cache when another distributed process updates the store
- The `_entities` dictionary is populated once and never updated

### 2. Inefficient Data Fetching

The current query uses a LEFT OUTER JOIN that returns the entity representation
for **every** measurement result:

```sql
SELECT ent.identifier, ent.representation, res.data
FROM {tablename} ent
LEFT OUTER JOIN {tablename}_measurement_results res ON res.entity_id = ent.identifier
```

This means:

- If an entity has 100 measurements, its representation is returned 100 times
- The entity JSON is parsed and validated on line 381-383, but only stored if
  not already in `_entities` (line 378)
- Significant waste of network bandwidth and CPU cycles

### 3. No Incremental Update Support

- Cannot efficiently fetch only new measurements added since last refresh
- Must re-fetch all data even if only a few new results were added
- The `insert_id` column in `measurement_results` table is not utilized for
  tracking

## Proposed Solution

Refactor the `entities` property to use a two-phase fetch strategy:

### Phase 1: Fetch All Entities

```sql
SELECT identifier, representation
FROM {tablename}
```

- Fetch entities once, without duplication
- Build the `_entities` dictionary efficiently

### Phase 2: Fetch Measurement Results with Metadata

```sql
SELECT insert_id, entity_id, data
FROM {tablename}_measurement_results
ORDER BY insert_id
```

- Fetch results with their `insert_id` for tracking
- Use `entity_id` as key to look up the entity in `_entities`
- Store the maximum `insert_id` seen for incremental updates

### Phase 3: Add Refresh Capability

Add a new method to support refreshing:

```python
def refresh(self) -> None:
    """Refresh entities and fetch new measurement results."""
```

```sql
SELECT insert_id, entity_id, data
FROM {tablename}_measurement_results
WHERE insert_id > :last_insert_id
ORDER BY insert_id
```

Also fetch any new entities that don't exist in the cache yet.

## Benefits

1. **Reduced network overhead**: Entity representations fetched once instead of
   N times
2. **Faster initialization**: No redundant JSON parsing
3. **Incremental updates**: Can efficiently sync with distributed processes
4. **Better scalability**: Handles large datasets with many measurements per
   entity
5. **Backward compatible**: Existing code continues to work, refresh is opt-in

## Implementation Notes

- Track `_last_insert_id` as instance variable
- Consider thread-safety if multiple processes access the same store
- Add logging for refresh operations
- Update tests to verify incremental refresh behavior

AlessandroPomponio:
Resolves #477


AlessandroPomponio:
I need to check one of the tests again - it can sometime fail

orchestrator/core/samplestore/sql.py
```diff
@@ -350,66 +353,270 @@ def location(self) -> orchestrator.utilities.location.SQLStoreConfiguration:
...
+                ):
+                    self.log.info(
+                        f"Result with uid {measurement_result.uid} was already present "
+                        f"in entity with identifier {entity_id}"
+                    )
+                    continue
+
+                self._entities[entity_id].add_measurement_result(
                     result=measurement_result
                 )
```
michael-johnston: The check is already in add_measurement_result - avoid duplication

```suggestion
               try:

                    self._entities[entity_id].add_measurement_result(
                        result=measurement_result
                    )
               except DuplicateMeasurementResultError:
                   pass
```

michael-johnston (approved):

PR #494, opened 2026-02-03
fix(modules): avoid displaying empty spinner panel

### LINKED ISSUE #476: bug: empty panel is displayed when running an operation that has no spinners
This can be seen with `ado create operation -f operation_bayesopt.yaml --use-latest space`

Image

AlessandroPomponio:
Resolves #476


christian-pinto (approved):
LGTM

PR #498, opened 2026-02-03
feat(core): optimize entity retrieval with targeted SQL queries

michael-johnston:
## Summary

This PR introduces significant performance optimizations for entity retrieval operations by replacing inefficient "load-all-then-filter" patterns with targeted SQL queries. These changes dramatically reduce memory usage and improve performance when working with large sample stores.

Previously it might take 10+ mins to complete the following command that now takes 13s 

```commandline
time ado show details operation randomwalk-1.3.4.dev46+d6bf32a.dirty-d4a8d3
──────────────────────────────────────────────────────────────────────────────────────── DETAILS ─────────────────────────────────────────────────────────────────────────────────────────
                                                                  
 Total entities sampled                                      1200 
 Total entities with no successful measurements              0    
 Total entities with only partially successful measurements  37   
 Total entities with all successful measurements             1163 
─────────────────────────────────────────────────────────────────────────────────── RELATED RESOURCES ────────────────────────────────────────────────────────────────────────────────────
actuatorconfiguration
  - actuatorconfiguration-SFTTrainer-d1bb837f
discoveryspace
  - space-1fb50e-7cdc8e
ado show details operation randomwalk-1.3.4.dev46+d6bf32a.dirty-d4a8d3  2.52s user 0.34s system 21% cpu 13.330 total
```

## Key Performance Improvements:

- sampledEntities(): Now fetches only required entities instead of loading all entities from the sample store
- show_details CLI: Uses SQL aggregation to compute statistics instead of loading all measurement results into Python
- experimentCatalog(): Queries single entity directly instead of loading all entities

## New SQL Methods

- entities_with_identifiers(): Efficiently fetch specific entities by their identifiers without loading all entities
- entities_in_operation(): Get entities directly from a single operation in one SQL query
- operation_entity_statistics(): Compute entity-level statistics using SQL aggregation (counts, success rates)

## Additional Improvements

- Added metadata_store parameter to DiscoverySpace methods to allow reusing SQLStore instances and avoid repeated instantiation
- Implemented table existence caching in SQLStore to reduce redundant database queries
- Added comprehensive test coverage for all new methods


orchestrator/core/samplestore/sql.py
```diff
@@ -1126,6 +1359,72 @@ def measurement_results_count_for_operation(
...
+        """
+        try:
+            with self.engine.begin() as connectable:
+                # Step 1: Get per-entity statistics (total and valid measurement counts)
+                # Valid measurement: JSON_EXTRACT(res.data, '$.reason') IS NULL
+                # Invalid measurement: JSON_EXTRACT(res.data, '$.reason') IS NOT NULL
+                query_text = f"""
+                    SELECT
+                        res.entity_id,
+                        COUNT(*) as total_measurements,
```
AlessandroPomponio: This could probably be simplified to be `COUNT(res.uid)` (to be double checked that it's accurate). Retrieving the count of an index is simpler than counting all rows

orchestrator/core/discoveryspace/space.py
AlessandroPomponio: The change in how DiscoverySpaces are initialised has effects on other CLI commands which now feel unresponsive because they had assumptions on how things were fetched. 

One of these examples is `ado show details space`

AlessandroPomponio (approved):

PR #500, opened 2026-02-04
feat(autoconf): Introduce a new recommender for per_device_train_batch_size

### LINKED ISSUE #499: feat: introduce an autoconf recommender for per_device_train_batch_size

The current min_gpu_recommender in autoconf uses effective batch size as a feature. This is in contrast to current tuning workflows where only. the per device batch size is available

We need a recommender that can provide the min gpus for a particular per GPU batch size. We can achieve this through a new experiment that uses the same underlying AutoGluon model. 

Updating the min gpu recommender will only overload the existing experiment.

srikumar003:
Fixes #499

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+## [1.4.1](https://github.com/ibm/ado/compare/1.4.0..1.4.1) - 2026-01-30
+#### Features
+- (**schema**) add function for generating entity id (#471) - ([5412418](https://github.com/ibm/ado/commit/5412418a94d7558da39850b37176d29ec93c2135)) - Michael Johnston
+#### Documentation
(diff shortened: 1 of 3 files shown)


VassilisVassiliadis (changes requested):
We need a record of the model version that this new experiment is using.

I can see 2 ways:
- Direct: have an optional constitutive property with a default value
- Indirect: Document in a file the model version that this version of the experiment is using

srikumar003:
I have added  an optional constitutive property for model_version with a default value "3.1.0"

danielelotito:
this is fine for me as well.
I just +1 on your comment in `config_mapper.py`
```
# TODO(srikumarv): Need to make this formal somewhere so that it is resilient and adaptable to model and training data updates
```
having that script in the util folders looks like a good enough place to keep our agreed ground truth for the mapping

PR #513, opened 2026-02-06
fix(cli): validate experiments against entity space in create space dry run

### LINKED ISSUE #507: bug: ado create space --dry-run does not validate the properties and domains against the experiment
## Issue Description

`ado create space -f SPACE --dry-run` does not fully validate a space. In particular it does not check the constitutive properties are compatible with the experiment. 

This is particularly important as this is the main way for generated YAML to be validated without creation. 

### How to reproduce

Install the custom experiments in examples/optimization_test_functions. 

Save this yaml, which uses the experiment, as space.yaml 
```yaml
entitySpace:
  - identifier: made_up_name
    propertyDomain:
      domainRange: [-10, 10] 
      variableType: CONTINUOUS_VARIABLE_TYPE
  - identifier: x1
    propertyDomain:
      domainRange: [-10, 10] 
      variableType: CONTINUOUS_VARIABLE_TYPE
  - identifier: x0
    propertyDomain:
      domainRange: [-10, 10] 
      variableType: CONTINUOUS_VARIABLE_TYPE
  - identifier: name
    propertyDomain:
        values: ['random_string']
experiments:
metadata:
  name: 'rosenbrock_3d'
```

One property is non-existent (made_up_name), a required property is missing ("x2") and the value of "name" is not valid. However if you validate it
```
ado create space -f space.yaml --dry-run
The configuration passed is valid!
```

## Expected behaviour

The -dry-run should emit an error that the yaml is not valid

### Python/ado/system info

Please include the output of:

```terminaloutput
python --version - Python 3.10.16
ado version - 1.4.2.dev8+174fcb1d
Your OS - Mac
```

## Additional information

This is because the validation does not happen until the DiscoverySpace object is created. The configuration does not validate the entity space versus the experiments as this requires the registry and it deliberately does not try to use this (as we want to be able to load and display YAML even if the experiment is missing from the registry). 

We need to either let dry-run initialise the space fully (have to consider if this is too expensive) or to do the test manually e.g. add a validate method to DiscoverySpaceConfiguration which you have to call explicitly.

michael-johnston:
Fixes #507

### DIFF
--- a/orchestrator/cli/resources/discovery_space/create.py
+++ b/orchestrator/cli/resources/discovery_space/create.py
@@ -191,7 +191,7 @@ def create_discovery_space(parameters: AdoCreateCommandParameters) -> str | None
             raise typer.Exit(1) from error
         except UnknownExperimentError as error:
             console_print(
-                f"{ERROR}Unknown experiment in configuration. This can be due to an actuator not being installed of if the referenced experiment is external: {error}",
+                f"{ERROR}Unknown experiment in configuration. This can be due to an actuator not being installed or if the referenced experiment is external: {error}",
                 stderr=True,
             )
             raise typer.Exit(1) from error


AlessandroPomponio (approved):
LGTM thanks

PR #517, opened 2026-02-06
fix(actuators): use correct typing and classmethod decorators 

michael-johnston:
There were a number of mistyped functions and instance variables which then hid an issue where what appeared to be instance methods were actually classemethods.


orchestrator/modules/actuators/registry.py
AlessandroPomponio: I don't think we support AnyStr here, we can simplify the type annotation as well

```suggestion
        self.actuatorIdentifierMap: dict[str, type["ActuatorBase"]] = {}
        # Maps actuator ids to ExperimentCatalog instances
        self.catalogIdentifierMap: dict[str, ExperimentCatalog] = {}
```

PR #538, opened 2026-02-13
refactor(cli): add validation message for operation yaml in create --dry-run

### LINKED ISSUE #534: feat: --dry-run clarifies that if it fails on invalid id that the YAML is valid (create operation)

When creating an operation YAML you must eventually put a correct space id and optional and actuator id. 
However when writing example YAML for documentation you may only have a "fake" id. 

The problem is ado create --dry-run will attempt to validate the fake id and fail, which is fine, but then it will not validate the parameters (or if it does, it does not say). 

To have the validation emit Success/Error separately for the resource and the yaml format/parameters. This way even with a fake id the --dry-run will tell if the YAML structure and parameter names are ok

YAML with valid parameters but invalid id
```
Operation YAML: Success 
Resource Identifiers: FAIL (Usual error)
```

Wrong identifiers and wrong parameters
```
Operation YAML: Fail  (Usual Pydantic Error)
Resource Identifiers: Not tested as YAML failed to load
```

I imagine the YAML and the operation parameters are validated first. Then once that is valid the ids are tested.

AlessandroPomponio:
Resolves #534

### DIFF
--- a/orchestrator/cli/resources/operation/create.py
+++ b/orchestrator/cli/resources/operation/create.py
@@ -141,7 +141,8 @@ def create_operation(parameters: AdoCreateCommandParameters) -> str | None:
             except ValueError as e:
                 status.stop()
                 console_print(
-                    f"{ERROR}The actuator configuration validation failed: {e}",
+                    f"{ERROR}The provided actuator configurations are "
+                    f"not compatible with the discovery space: {e}",
                     stderr=True,
                 )
                 raise typer.Exit(1) from e


michael-johnston:
```
(ado-core) michaelj@Michaels-MacBook-Pro-2 3b-model-pareto % ado create operation -f operation.yaml --dry-run
INFO:   The operation YAML is syntactically valid.
ERROR:  The actuator configuration validation failed:
There is no resource of kind CoreResourceKinds.DISCOVERYSPACE with the requested id, SPACE_ID_PLACEHOLDER, in the project
```

Why is the error from actuator configuration failed that the space did not exist?

YAML used (install optuna) 

```yaml
metadata:
  name: 3b-model-pareto-optimization
  description: Multi-objective optimization to find pareto front of latency vs throughput for 3B models
# Will be populated after creating the space
spaces:
  - SPACE_ID_PLACEHOLDER # Replace with actual space ID from: uv run ado create space -f space.yaml
# Reference to actuator configuration (optional, can be created separately)
actuatorConfigurationIdentifiers: []
# - ACTUATOR_CONFIG_ID  # Uncomment and add actuator config ID if needed

operation:
  module:
    operatorName: "ray_tune"
    operationType: "search"
  parameters:
    tuneConfig:
      # Multi-objective optimization: minimize latency, maximize throughput
      metric:
        - mean_e2el_ms # End-to-end latency in milliseconds (minimize)
        - total_token_throughput # Total token throughput (maximize)
      mode:
        - 'min' # Minimize latency
        - 'max' # Maximize throughput
      num_samples: 100 # Number of trials to explore the pareto front
      max_concurrent_trials: 2 # Concurrent deployments (adjust based on cluster capacity)
      search_alg:
        name: optuna # Optuna supports multi-objective optimization
        params: {}
        # Optuna will automatically use NSGA-II algorithm for multi-objective optimization
        # This evolutionary algorithm is designed to find pareto-optimal solutions
    runtimeConfig:
      stop:
        - name: MaxSamplesStopper
          keywordParams:
            max_samples: 100 
    orchestratorConfig:
      metric_format: target
      single_measurement_per_property: true
      result_dump: none
```

AlessandroPomponio:
> Why is the error from actuator configuration failed that the space did not exist?

Because the function to validate actuator configuration validates them against a space, so failing to fetch a space will fail the validation

https://github.com/IBM/ado/blob/e6cbb71447fd098f5deaf885059e2c88378476d9/orchestrator/core/operation/config.py#L82-L173

michael-johnston:
No this message 
```
ERROR:  The actuator configuration validation failed:
```

Is misleading as it implies the validation of an actuator configuration failed: but there is no actuator configuration so no validation of such a thing failed. 

The error happened during a call to validate_actuator_configuration passing an empty list of configurations. 

One side effect is that is misleads  agents into thinking there is something wrong with the actuator configuration YAML.

PR #546, opened 2026-02-16
fix(core): use raw dictionary instead of RuntimeEnv to initialise ray

michael-johnston:
Fixes some cases missed in #506 

- Adds a session scoped fixture to the tests which ensure ray is initialised with working_dir None
   - Otherwise any test require ray initialisation would attempt to upload the working directory when run via `uv run`
- Works around a bug in ray where RuntimeEnv(working_dir=None) does not result in working_dir being set to None
   - The changes in #506 did not ensure `uv run ado create operation` would work

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,8 +2,10 @@
 # SPDX-License-Identifier: MIT
 
 import uuid
+from collections.abc import Callable, Generator
 
 import pytest
+import ray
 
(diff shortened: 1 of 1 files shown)


AlessandroPomponio:
Do we need to apply this fix everywhere we have `ray.init` in the code? Because I see we have multiple that have no args, for example

image

PR #611, opened 2026-02-25
refactor(examples): update CSVSampleStores with old keys

AlessandroPomponio:


michael-johnston (commented):
Is anything using [sample_store_resource.json‎](https://github.com/IBM/ado/pull/611/changes#diff-6ae9983566d6e780a306057b3183ffa4c6f565a104f12fb46e62573827d22c63)?

Seems strange it could be wrong and used?

I see this shows a problem with how we test ml_multi_cloud. 
At one stage we ran tests using the files in the example folder - however this meant that we need to use absolute path for the CSV file in the example, which mean the user had to change it to run the example. 

To fix this we create a separate version of the ml_multicloud_sample_store.yaml for use in unit-tests -however also update the end-to-end test in tox to use this file (as the end to end text was running from route of repo). This meant the actual example was no longer tested at all. 

I think we need to also change the tox test to use the actual example file - either by running in the ml_multi_cloud example directory or by adding an override of the CSV path.

PR #630, opened 2026-02-27
fix(cli): correctly handle virtual properties in show entities

michael-johnston:
Note: At the moment its expected a single property value that's a list,  `[[1,2,3]]` and 3 property separate values [1,2,3] would aggregate in the same way i.e. the mean would be 2 in both cases 

This PR fixes the following: 

- experimentSeries aggregation behaviour did not aggregate [[1,2,3]] although seriesRepresentation did 
   - this meant `ado show entities space --aggregate mean` had different behaviour based on column header format
- `ado show entities --property [Virtual Identifier]` did not work.  
- ado-ray-tune raised validation error on a virtual target identifier e.g. `wallclocktime-mean` even though this has always been supported. 
   - this was due to recent use of MeasurementSPace.propertyIdentifierInSpace as a guard which did not handle virtual target identifiers

### DIFF
--- a/orchestrator/schema/entity.py
+++ b/orchestrator/schema/entity.py
@@ -719,11 +719,11 @@ class Entity(pydantic.BaseModel):
                         expProperties[vop.virtualTargetPropertyIdentifier] = (
                             aggregate_value.value
                         )
-                        
+
                         # Continue iterating on success - in other cases (no aggregation
                         # and failure to aggregate), we replace lists with 1 element with
                         # the element itself.
-                        continue         
+                        continue
                     except (
                         ValueError,
                         TypeError,


orchestrator/schema/entity.py
```diff
@@ -701,38 +701,46 @@ def experimentSeries(
...
-                            log.debug(
-                                f"Unable to calculate mean of value for {o} due to {error}. Will not aggregate"
-                            )
-                    else:
-                        pass
-                else:
+                # No aggregation was requested or it failed
+                if len(expProperties[o]) == 1:
+                    # If no aggregation and there is only 1 value we remove the list
+                    expProperties[o] = expProperties[o][0]
```
AlessandroPomponio: A duplicate line somehow slipped through
```suggestion
```

PR #634, opened 2026-03-01
fix(cli): ensure run_experiment result is fully printed

michael-johnston:
This was the case at some point but it had reverted in latest main

### DIFF
--- a/orchestrator/utilities/run_experiment.py
+++ b/orchestrator/utilities/run_experiment.py
@@ -19,7 +19,6 @@ from orchestrator.cli.utils.output.prints import (
     ERROR,
     WARN,
     console_print,
-    set_pandas_display_options,
 )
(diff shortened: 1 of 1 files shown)


orchestrator/utilities/run_experiment.py
```diff
@@ -319,6 +324,7 @@ def run(
                     )
                 else:
                     console_print("Result:")
+                    set_pandas_display_options()
                     console_print(
                         f"{request.series_representation(output_format='target')}\n",
```
AlessandroPomponio: I think the actual fix would be to remove the fstring here

The `console_print` function, when `has_pandas_content` is True, already runs `set_pandas_display_options`.

But by having the fstring here, the content is already rendered by pandas before the function is run

Could you try it by doing this change?

PR #669, opened 2026-03-07
perf(core): cache SQLAlchemy engine per database URL to eliminate redundant TCP handshake

michael-johnston:
Without caching, every component that calls engine_for_sql_store for the same database URL (the metastore SQLResourceStore and the samplestore SQLSampleStore being the primary pair) each calls sqlalchemy.create_engine, which creates a separate connection pool and therefore opens a separate TCP+TLS+MySQL auth connection.

Over a high-latency link the full handshake can const 1-3. With two independent engines, both the metastore and samplestore each paid this cost, adding an addition 1-3s to every CLI command.

_engine_cache (a process-level dict keyed by connection URL) ensures that the second call for the same URL returns the already-initialised Engine and reuses its warm connection pool, reducing the connection cost from 2x to 1x handshake per process lifetime.

Local measured saving on `ado show requests/results/entities`: ~2,360ms.

### DIFF
--- a/orchestrator/metastore/sql/utils.py
+++ b/orchestrator/metastore/sql/utils.py
@@ -5,10 +5,29 @@ import sqlalchemy
 
 from orchestrator.utilities.location import SQLStoreConfiguration
 
+# Process-level cache: reuse the same SQLAlchemy Engine (and its connection pool)
+# for every call with the same database URL.  This means the metastore and the
+# samplestore — which both point at the same MySQL server — share one pool and
+# avoid the overhead of opening a second TCP connection.
+_engine_cache: dict[str, sqlalchemy.Engine] = {}
+
 
 def engine_for_sql_store(
     configuration: SQLStoreConfiguration, database: str | None = None
 ) -> sqlalchemy.Engine:
+    """Return a SQLAlchemy Engine for the given store configuration.
+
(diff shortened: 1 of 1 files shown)


AlessandroPomponio (approved):
LGTM

This is in line with what SQLAlchemy's documentation says: https://docs.sqlalchemy.org/en/20/core/connections.html#basic-usage

> The typical usage of [create_engine()](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine) is once per particular database URL, held globally for the lifetime of a single application process. A single [Engine](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Engine) manages many individual [DBAPI](https://docs.sqlalchemy.org/en/20/glossary.html#term-DBAPI) connections on behalf of the process and is intended to be called upon in a concurrent fashion. The [Engine](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Engine) is not synonymous to the DBAPI connect() function, which represents just one connection resource - the [Engine](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Engine) is most efficient when created just once at the module level of an application, not per-object or per-function call.

PR #698, opened 2026-03-16
refactor(core): remove ray.remote decorator from actuators

### LINKED ISSUE #564: refactor: remove remote decorate from actuators
As part of #185 the registry automatically de-decorates actuators when registering them. 

We should now remove the decorators from the actuators and then we can also remove the de-decoration code.

michael-johnston:
Closes #564


AlessandroPomponio:
> The difference is some of the classes which had their decorator removed by this function no longer have it.

I was wondering if we could get rid of it, but I'm gathering that we can't

PR #703, opened 2026-03-16
fix(schema): ensure experiment_id is string in series representation

michael-johnston:
The experiment_id field in ValidMeasurementResult.series_representation and Entity.experiment_series was being set to the ExperimentReference instance.

This rendered as the id when outputting as CSV. However, when outputting as json it was the pydantic JSON dump.

The field properly should just be the string id c.f. entity "identifier" field. This commit makes this the case.


michael-johnston:
@AlessandroPomponio I simplified one change.

PR #720, opened 2026-03-20
fix(vllm_performance): Replace binary type with discrete type for properties in geospatial experiments

christian-pinto:
This PR replaces the `BINARY_VARIABLE_TYPE` with `DISCRETE_VARIABLE_TYPE` for the `skip_tokenizer_init` and `enforce_eager` properties of the `performance_testing-geospatial-full` and `performance_testing-geospatial-full-custom-dataset` experiments in the vllm-performance  actuator. 

This change is needed because those properties are to be forcibly set to 1 (True) for loading the Geospatial models with vLLM. Using a binary type would instead lead to an entity space where both values in the domain (True, False) are tested, causing the failure of the vLLM engine for all the entities with one (or both) of the above properties set to False.


AlessandroPomponio (approved):
LGTM thanks

PR #735, opened 2026-03-24
build: refactor to achieve proper automatic package discovery

danielelotito:
Fixes the CI problem below

```terminal
/root/.cache/uv/builds-v0/.tmpc8qm1z/lib/python3.10/site-packages/setuptools/command/build_py.py:215: _Warning: Package 'no_priors_characterization.utils' is absent from the `packages` configuration.
!!
        ********************************************************************************
        ############################
        # Package would be ignored #
        ############################
        Python recognizes 'no_priors_characterization.utils' as an importable package[^1],
        but it is absent from setuptools' `packages` configuration.
        This leads to an ambiguous overall configuration. If you want to distribute this
        package, please make sure that 'no_priors_characterization.utils' is explicitly added
        to the `packages` configuration field.
        Alternatively, you can also rely on setuptools' discovery methods
        (for example by using `find_namespace_packages(...)`/`find_namespace:`
        instead of `find_packages(...)`/`find:`).
        You can read more about "package discovery" on setuptools documentation page:
        - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html
        If you don't want 'no_priors_characterization.utils' to be distributed and are
        already explicitly excluding 'no_priors_characterization.utils' via
        `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,
        you can try to use `exclude_package_data`, or `include-package-data=False` in
        combination with a more fine grained `package-data` configuration.
        You can read more about "package data files" on setuptools documentation page:
        - https://setuptools.pypa.io/en/latest/userguide/datafiles.html
        [^1]: For Python, any directory (with suitable naming) can be imported,
              even if it does not contain any `.py` files.
              On the other hand, currently there is no concept of package data
              directory, all directories are treated like packages.
        ********************************************************************************
!!

```

### DIFF
--- a/plugins/operators/trim/trim/tests/test_high_dimensional_sampling.py
+++ b/plugins/operators/trim/trim/tests/test_high_dimensional_sampling.py
@@ -16,7 +16,7 @@ import pytest
 from no_priors_characterization.utils.high_dimensional_sampling import (
     concatenated_latin_hypercube_sampling,
 )
-from trim.test.test_data_documentation import TEST_DATAFRAMES
+from trim.tests.test_data_documentation import TEST_DATAFRAMES
 
 
 class TestConcatenatedLatinHypercubeSampling:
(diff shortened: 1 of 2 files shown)


AlessandroPomponio:
Superseded by #740

PR #805, opened 2026-04-08
refactor(core)!: use class decorators for explore operators and add OperatorMetadata

michael-johnston:
Introduce ability to decorate explore operator classes directly avoiding need for a separate function and related issues due to different sources of operator metadata. 

For example with this change the operator class defines the operators name;  this name is what it is registered as (ado get operators); it is used to generate the operator identifier and operation identifier field values;  and it is what appears in operation resource yaml . Previously these could all be set differently.  

## Changes
- Extend `@explore_operation` to decorate classes directly. No separate function is needed (its generated automatically)
- Migrate `RandomWalk` and `RayTune` to the class-decorator pattern; delete the now-redundant functions for both
- Centralise operator metadata in a new `OperatorMetadata` model (moved to `core/operation/config.py`) as the single source of truth for each operator's name, version, type, configuration model, and implementation class.
- Rename `OperatorFunctionConf` → `OperatorReference` (deprecated alias kept for backward compatibility).
- Introduce OperatorFunction a protocol based type which can be used to type and validate operator function signatures
- Explore operations now store an OperatorReference instance in the operation resource i.e. the registered operator name rather than class name
- orchestration functions (orchestrate_general_operation, orchestrate_explore_operation, _run_operation_harness) updated to use OperatorMetadata

## Breaking

The following are not supported any more

- **explore operator classes without an implementation of the operator_metadata() class method**:  Explore operator classes must be updated to implement `operator_metadata()`.  
      - Instantiating an existing class that does not implement this method will cause `TypeError: Can't instantiate abstract class OPERATOR_CLASS_NAME with abstract method operator_metadata`
      - Calling this class method on the class (without instantiation) will cause `NotImplementedError` to be raised 
      - Note: Via ado cli user will encounter one of the following two errors first if such a class exists. 
- **decorating functions with explore_operation decorator**: This decorator can only be used on classes. 
      - Existing functions decorated with @explore_operation will cause `TypeError: `explore_operation` takes 1 positional argument but N were given` to be raised when the operator registry is accessed via the cli (e.g. ado get operators, ado create op) 
- **Passing an operation yaml with OperatorModuleConf fields for the module field to `ado create op`**: This was allowed but not documented anywhere as viable approach (existing only in some test files and one old example file). It's no longer allowed. 
- **Not supplying values for version, configuration_model or configuration_model_default to operation function decorators other than `explore_operation`**:  These fields are now required.
    -  Existing uses of these decorators that don't provide values for these fields will cause `TypeError: `$DECORATOR_NAME` takes 4 positional arguments but (number less than 4) were given` to be raised when the operator registry is accessed via the cli (e.g. ado get operators, ado create op) 

## Deprecation
- **`OperatorFunctionConf`**: still importable as a deprecated alias for `OperatorReference`.
- operatorIdentifier, operationType, defaultOperationParameters, validateOperationParameters of DiscoveryOperationBase - replaced by operator_metadata


orchestrator/modules/operators/collections.py
```diff
@@ -142,114 +111,201 @@ def __getattr__(
...
 
 
 def explore_operation(
-    name: str,
+    target: "type[DiscoveryOperationBase] | None" = None,
+    *,
+    name: str | None = None,
     description: str | None = None,
     configuration_model: type[pydantic.BaseModel] | None = None,
     version: str | None = "v0.1",
```
AlessandroPomponio:
```suggestion
    version: str | None = "0.1.0",
```

orchestrator/modules/operators/collections.py
```diff
@@ -302,18 +365,49 @@ def export_operation(
...
+                discovery_space=discoverySpace,
+                operation_info=operationInfo or FunctionOperationInfo(),
+                operation_type=orchestrator.core.operation.config.DiscoveryOperationEnum.EXPORT,
+            )
+
+        validate_operator_function_signature(wrapper)
+        export.operators[name] = OperatorMetadata(
+            name=name,
+            function=typing.cast("OperatorFunction", wrapper),
+            version=version or "v0.1",
```
AlessandroPomponio:
```suggestion
            version=version,
```

AlessandroPomponio (dismissed):

orchestrator/modules/operators/orchestrate.py
```diff
@@ -56,6 +58,24 @@ def graceful_orchestrate_shutdown() -> None:
...
+        moduleLog.warning(
+            "The supplied operation configuration uses the unsupported OperatorModuleConf format "
+            "(moduleName/moduleClass). Use OperatorReference (operatorName/operationType) "
+            "instead. See the documentation for migration guidance."
+        )
+        raise ValueError(
+            "The supplied operation configuration uses the unsupported OperatorModuleConf format "
+            "(moduleName/moduleClass). Use OperatorReference (operatorName/operationType) "
+            "instead. See the documentation for migration guidance."
+        ) from None
```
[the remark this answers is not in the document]
AlessandroPomponio: You get a ruff error if you catch an exception and raise a different one not using `from`

orchestrator/core/operation/config.py
```diff
@@ -182,19 +182,89 @@ def operationType(self) -> DiscoveryOperationEnum:
...
+                "operator_metadata() before the decorator injects it."
+            ),
+        ),
+    ] = None
+    version: Annotated[
+        str,
+        pydantic.Field(
+            description="Semantic version string for the operator (e.g. '0.1.0').",
+        ),
+    ] = "0.1.0"
```
AlessandroPomponio, line 219: We didn't add the function to validate this is a valid semver string

orchestrator/modules/operators/base.py
```diff
@@ -43,49 +41,152 @@
...
+        :class:`~orchestrator.core.operation.config.OperatorMetadata` instance
+        that describes the operator's name, version, type, and configuration
+        model.  The ``@explore_operation`` decorator fills in the ``function``
+        and ``cls`` fields before registering::
+
+            @classmethod
+            def operator_metadata(cls) -> OperatorMetadata:
+                return OperatorMetadata(
+                    name="my_op",
+                    version="v1.0",
```
AlessandroPomponio:
```suggestion
                    version="0.1.0",
```

orchestrator/modules/operators/base.py
```diff
@@ -43,49 +41,152 @@
...
+        proto_hints = typing.get_type_hints(OperatorFunction.__call__)
+        proto_sig = inspect.signature(OperatorFunction.__call__)
+    except Exception:  # noqa: BLE001
+        return
+
+    try:
+        hints = typing.get_type_hints(fn)
+        sig = inspect.signature(fn)
+    except Exception:  # noqa: BLE001
+        return
```
AlessandroPomponio: If we can't inspect the signature or get the type hints we don't raise any Exception - could this be a problem?

orchestrator/modules/operators/orchestrate.py
```diff
@@ -56,6 +58,24 @@ def graceful_orchestrate_shutdown() -> None:
...
+    ):
+        moduleLog.warning(
+            "The supplied operation configuration uses the unsupported OperatorModuleConf format "
+            "(moduleName/moduleClass). Use OperatorReference (operatorName/operationType) "
+            "instead. See the documentation for migration guidance."
+        )
+        raise ValueError(
+            "The supplied operation configuration uses the unsupported OperatorModuleConf format "
+            "(moduleName/moduleClass). Use OperatorReference (operatorName/operationType) "
+            "instead. See the documentation for migration guidance."
```
AlessandroPomponio: Might be worth to add a link or tell people where to look (what doc?)

orchestrator/modules/operators/randomwalk.py
```diff
@@ -965,54 +947,19 @@ def numberMeasurementsRequested(self) -> int:
...
-    import orchestrator.modules.module
+    def operator_metadata(cls) -> OperatorMetadata:
+        """Returns operator metadata for the random_walk explore operator."""
+        return OperatorMetadata(
+            name="random_walk",
+            version=version("ado-core"),
+            description=cls.description(),
+            configuration_model=RandomWalkParameters,
+            example_configuration=RandomWalkParameters(),
+            type=DiscoveryOperationEnum.SEARCH,
```
AlessandroPomponio, line 960: RandomWalk is a superclass of Characterize but its type is SEARCH. It seems a bit counterintuitive, given that the characterize type exists

plugins/operators/ray_tune/ado_ray_tune/operator.py
```diff
@@ -996,16 +980,27 @@ def numberMeasurementsRequested(self) -> int:
...
+            description=cls.description(),
+            configuration_model=RayTuneConfiguration,
+            example_configuration=RayTuneConfiguration(
+                tuneConfig=OrchTuneConfig(
+                    metric="wallclock_time",
+                    search_alg=OrchSearchAlgorithm(name="bayesopt"),
+                ),
+                runtimeConfig=OrchRunConfig(),
+            ),
+            type=DiscoveryOperationEnum.SEARCH,
```
AlessandroPomponio, line 1001: We still have the dissonance between search/explore. We should fix it

orchestrator/modules/operators/collections.py
```diff
@@ -9,125 +9,93 @@
...
+    operators: Annotated[
+        dict[str, OperatorMetadata], pydantic.Field(default_factory=dict)
+    ]
 
-    def default_configuration_model_for_operation(
-        self, name: str
-    ) -> pydantic.BaseModel:
-        if name not in self.function_operation_models:
-            raise ValueError(f"Unknown operator {name}")
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
AlessandroPomponio: why is `arbitrary_types_allowed` needed?

orchestrator/modules/operators/collections.py
```diff
@@ -142,178 +110,250 @@ def __getattr__(
...
+    classmethod.  The decorator generates an :class:`OperatorFunction`, validates
+    its signature, and registers it in the explore collection::
+
+        @explore_operation
+        class MyOp(Search):
+            @classmethod
+            def operator_metadata(cls) -> OperatorMetadata:
+                return OperatorMetadata(
+                    name="my_op",
+                    version="v1.0",
```
AlessandroPomponio:
```suggestion
                    version="0.1.0",
```

AlessandroPomponio (dismissed):

orchestrator/modules/operators/randomwalk.py
michael-johnston: This is an older problem - characterise at very start had different meaning - I can fix it here I suppose

plugins/operators/ray_tune/ado_ray_tune/operator.py
michael-johnston: Let's do that separately - can you open an issue?

orchestrator/modules/operators/base.py
michael-johnston: Probably ... will change

orchestrator/modules/operators/collections.py
michael-johnston: I think it was because of the now deleted field object_operations, (old line 30). 

Will remove it

orchestrator/modules/operators/orchestrate.py
michael-johnston: I'll change this so we point to any operation example as they all show operatorName/operationType.

orchestrator/modules/operators/randomwalk.py
michael-johnston: The fix is to change its superclass to Search and remove Characterise class.

orchestrator/core/operation/config.py
michael-johnston: Added

orchestrator/modules/operators/_cleanup.py
```diff
@@ -29,6 +29,7 @@ def handler(sig: int, frame: typing.Any | None) -> None:  # noqa: ANN401
 
         if shutdown_signal_received:
             moduleLog.info("Graceful shutdown already completed")
+            return
```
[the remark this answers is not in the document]
michael-johnston: I'll deal with this once the rest is ok

orchestrator/core/operation/config.py
```diff
@@ -182,19 +182,117 @@ def operationType(self) -> DiscoveryOperationEnum:
...
+        """
+        from packaging.version import InvalidVersion, Version
+
+        try:
+            Version(value)
+        except InvalidVersion as exc:
+            raise ValueError(
+                f"Operator version {value!r} is not a valid PEP 440 version string: {exc}"
+            ) from exc
+        return value
```
AlessandroPomponio: I would definitely move this after the other fields - once I start seeing the validators I assume there aren't any more fields.

It might also be worth to move the function to `orchestrator/utilities` (maybe versioning.py or something along the lines), so that it can be imported here (and wherever needed) and used according to the Annotated pattern https://pydantic.dev/docs/validation/latest/concepts/validators/#field-validators

orchestrator/core/operation/config.py
```diff
@@ -182,19 +182,117 @@ def operationType(self) -> DiscoveryOperationEnum:
...
+        ),
+    ] = None
+    type: Annotated[
+        DiscoveryOperationEnum,
+        pydantic.Field(
+            description="The discovery operation type this operator belongs to."
+        ),
+    ]
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
AlessandroPomponio: Was it this one that needed to be removed?

orchestrator/modules/operators/collections.py
```diff
@@ -7,127 +7,104 @@
...
+def _warn_if_operator_name_reused(
+    collection_label: str, name: str, operators: dict[str, OperatorMetadata]
+) -> None:
+    """Log a warning when registering under a name that is already in use."""
+    if name in operators:
+        moduleLog.warning(
+            "Operator %r is already registered in %s; replacing the existing entry",
+            name,
+            collection_label,
+        )
```
AlessandroPomponio, line 42: This seems like a programming error, it should probably not be just a warning

AlessandroPomponio (dismissed):
I still need to look at tests and documentation, but I noticed these things

orchestrator/modules/operators/collections.py
michael-johnston: It's a user error most likely - installing same package twice. Or perhaps a developer error naming their operator the same as someone else's. Not a programming error though.

orchestrator/core/operation/config.py
michael-johnston: No but this can also be removed. The original field it was required from is removed in this refactor - and also that field I just noticed was mistyped in the first place 😆

michael-johnston:
@AlessandroPomponio last comments addressed

AlessandroPomponio (approved):
LGTM thanks

PR #822, opened 2026-04-10
build(hooks): add default stage for hooks

### LINKED ISSUE #820: bug(build): pre-commit hooks are run twice
## Issue Description

We have hooks that run at two different times:
- pre-commit
- commit-msg

The only hook that (should) run at `commit-msg` time correctly specifies this, but no other hook does specify when they should be run, and no default is present. This causes all the `pre-commit` hooks to be run again once the `commit-msg` hook runs.

AlessandroPomponio:
Resolves #820

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,7 @@
 default_install_hook_types:
   - pre-commit
   - commit-msg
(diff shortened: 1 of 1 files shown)


christian-pinto (approved):
LGTM

PR #833, opened 2026-04-13
fix(cli): ensure we find space ids in delete operation

AlessandroPomponio:
When targeting operations that used actuator configurations, we might find their identifiers first.

Fixes #810


christian-pinto (approved):
LGTM

PR #846, opened 2026-04-14
feat(core): hash entity identifiers if too long for db engine

### LINKED ISSUE #382: feat: handle case where entity identifier is too long for the DB
## Issue

Currently, the entity identifier is created by appending constitutive properties into a long string, which may end up being too long for the database schema.

## Potential solution

- Handle on Python side: Generate a shorter, unique identifier (e.g., using hashing or UUID) while preserving a mapping to the original properties; substring the identifier. This ensures consistency between user-facing operations and stored values.
- Handle on SQL side:  Use functions like SUBSTRING (MySQL) or SUBSTR (SQLite) to truncate identifiers before insertion. This approach requires dialect-specific handling and still risks losing uniqueness.

ref: https://github.ibm.com/Discovery-Orchestrator/ad-orchestrator/issues/389

AlessandroPomponio:
# Summary

This PR adds automatic hashing for long entity identifiers to prevent database field length violations. The entity identifier is used as a primary key in SQLSampleStore. When an entity identifier exceeds 700 characters, it is now hashed using SHA256 to ensure it stays within the 768-character INNODB (MySQL) database limit for columns used as indexes while maintaining uniqueness and determinism.

**NOTE**: this issue stems from a limitation of the INNODB engine on MySQL when used with the utf8mb4 charset. Attempting to create a VARCHAR column with an index on it (such as a primary key) that is above 768 characters would fail with the following error:

> SQL Error [1071] [42000]: Specified key was too long; max key length is 3072 bytes

3072/4 = 768 is the maximum length possible

Resolves #382

# Files Changed

📄 `orchestrator/schema/entity.py`

Modified the `entity_identifier_from_properties_and_values` function to handle long identifiers. The function now checks if the generated identifier exceeds 700 characters (safe threshold below the 768-character database limit). If it does, the identifier is hashed using SHA256 and prefixed with "hash-" to indicate it's a hashed value. Short identifiers remain human-readable for debugging purposes.

📄 `tests/schema/test_entity.py`

Added comprehensive test coverage for the new identifier hashing functionality:
- `test_entity_identifier_short_not_hashed`: Verifies short identifiers remain unchanged and human-readable
- `test_entity_identifier_long_hashed`: Confirms long identifiers (4539+ chars) are properly hashed with the "hash-" prefix and stay within database limits (69 chars total)
- `test_entity_identifier_different_points_different_identifiers`: Ensures different input points produce unique identifiers for both short and long cases
- `test_entity_identifier_threshold_boundary`: Tests edge cases at exactly the 700-character threshold to verify the cutoff works correctly

### DIFF
--- a/orchestrator/schema/entity.py
+++ b/orchestrator/schema/entity.py
@@ -45,15 +45,32 @@ def entity_identifier_from_properties_and_values(point: dict[str, typing.Any]) -
     """
     Creates an entity identifier based on a set of constitutive property ids and values for those properties
 
+    If the identifier exceeds safe length (700 characters), it is hashed to ensure database compatibility.
+
     Parameters:
         point: A dictionary of constitutive property id, value pairs
 
     Returns:
-        An entity identifier
+        An entity identifier (human-readable if short, hashed if long)
     """
 
     parts = [f"{key}.{point[key]}" for key in sorted(point.keys())]
-    return "-".join(parts)
+    full_identifier = "-".join(parts)
+
+    # Use 700 as threshold to leave margin below 768 limit
+    MAX_SAFE_LENGTH = 700
+
+    if len(full_identifier) > MAX_SAFE_LENGTH:
+        import hashlib
+
+        # Hash long identifiers to ensure they fit in database
+        hash_hex = hashlib.sha256(
+            full_identifier.encode("utf-8"), usedforsecurity=False
+        ).hexdigest()
+        # Prefix to indicate it's a hash
+        return f"hash-{hash_hex}"
+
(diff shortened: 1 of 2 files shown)


michael-johnston:
We should probably note in title and description the dB that has the restriction e.g MySQL has a x limit in primary keys. Since entity id is primary … 

should also note if/if not SQLite has similar issue

AlessandroPomponio:
@michael-johnston let me know if this is better
Feel free to make suggestions otherwise

michael-johnston (approved):

PR #852, opened 2026-04-14
docs(agents): add examining project skill

michael-johnston:
Adds a new skill for building a project-wide picture of an ado context: activity volume, spaces and operations created over time, experiments and actuators in use, space relationship analysis, and a structured report
template. 

Cross-links the new skill from the existing examining-ado-operations and examining-discovery-spaces skills.

### DIFF
--- a/.cursor/skills/examining-ado-project/SKILL.md
+++ b/.cursor/skills/examining-ado-project/SKILL.md
@@ -123,7 +123,7 @@ batch / sample** configuration for explore operations.
 
 ## 3. Space relationships and entity-space shape
 
-From step 2, pick a handful of **high-traffic or recent** spaces. For each,
+From step 2, pick a handful of **commonly operated on or recent** spaces. For each,
 inspect its fragment in `spaces.yaml`, focusing on **entity space** structure
 (dimensions, bounds, representation).
 
--- a/.cursor/skills/examining-ado-project/SKILL.md
+++ b/.cursor/skills/examining-ado-project/SKILL.md
@@ -1,0 +1,179 @@
+---
+name: examining-ado-project
+description: >-
+  Builds a picture of work in an ado project: activity volume,
+  spaces and operations created over time, experiments and operation configs 
+  used etc. Use to create a project/context overview report,
+  summarize what the team has been doing in an ado project, report trends across
+  spaces/operations, or to onboard onto an ado project.
+---
+
+# Examining an ado Project
+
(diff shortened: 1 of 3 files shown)


.cursor/skills/examining-ado-project/SKILL.md
```diff
@@ -0,0 +1,192 @@
+---
+name: examining-ado-project
+description: >-
+  Builds a picture of work in an ado project: activity volume,
```
AlessandroPomponio, line 4: In CLI terms this i a context, it might be better to stick to that word everywhere. Or, if we believe project is better, we can change the cli to at first add it and then eventually deprecate context/ctx

.cursor/skills/examining-ado-project/SKILL.md
```diff
@@ -0,0 +1,192 @@
...
+
+- What is being **measured** and with which experiments (domain of the project).
+- Which **explore** (or similar) operations drove the largest submitted entity
+  sets and on **which spaces**.
+- Whether **experiment definitions or parameters** shift over time (versions,
+  configs).
+
+## 3. Space relationships and entity-space shape
+
+From step 2, pick a handful of **high-traffic or recent** spaces. For each,
```
AlessandroPomponio: With "high-traffic" you mean "commonly used", right?

.cursor/skills/examining-ado-project/SKILL.md
michael-johnston: I think context/project is valid - same as context/namespace in k8s. One is client side, one is server side. We need two terms as they are separate concepts - however since they are closely associated (a context is associated to a project by definition) they can often be used interchangeably in natural language without being ambiguous. 


An important factor is that this isn't docs: it's agent skill. One purpose is to help agent map the skill to what a human might type. So for example sometimes I might type project or others context due to my human brain being fuzzy. I need the skill to trigger on either. The skill will help the agent understand the nuances of the terminology

.cursor/skills/examining-ado-project/SKILL.md
michael-johnston: Sure - the agent picked high-traffic here as the term it liked. Maybe commonly used will be better for it.

michael-johnston:
```suggestion
From step 2, pick a handful of **commonly operated on or recent** spaces. For each,
```

AlessandroPomponio (approved):

PR #857, opened 2026-04-15
docs(vllm_performance): use optuna in vLLM performance endpoint example

### LINKED ISSUE #847: bug: Unable to run Ray Tune operations using hyperopt
## Issue Description

When using hyperopt with the ray_tune operator I get the below error 

```
(ado-core) christian@cp-dev-pod:~/ado$ LOGLEVEL=DEBUG ado create operation -f operation.yaml --use-latest ac --use-latest space | tee ado.logs
2026-04-14 11:55:42,312 DEBUG     MainThread           operation_collections: load_operators      : Loaded plugin: ado-ray-tune from ado_ray_tune.operator_function
2026-04-14 11:55:42,328 DEBUG     MainThread           operation_collections: load_operators      : Loaded plugin: rifferla from ado_ray_tune.rifferla
ERROR:  The operation configuration provided was not valid:
1 validation error for DiscoveryOperationResourceConfiguration
operation.tuneConfig
  Assertion failed, HyperOpt must be installed! Run `pip install hyperopt`. 
    For further information visit https://errors.pydantic.dev/2.13/v/assertion_error
```

This is happening even if `hyperopt` is installed.

### Screenshots/Logs

If applicable, add screenshots and logs (e.g., Ray Logs) to help explain your problem.

### Python/ado/system info

Please include the output of:

```terminaloutput
python --version
ado version
Your OS
```

christian-pinto:
This PR migrates the vLLM performance endpoint example from using HyperOpt to Optuna as the search algorithm, mainly because Optuna is more actively maintained and released. 

Using HyperOpt users might incur in runtime errors due to to using deprecated python libraries.

The example is slightly different as in optuna it is not possible to set the gamma value that control the number of top observations to be selected. The rest of the discussion, explaining how the optimizer behaves, is mostly related to the actual sampling algorithm (TPE) and therefore it is valid with both libraries.

This PR fixes #847

### DIFF
--- a/website/docs/examples/vllm-performance-endpoint.md
+++ b/website/docs/examples/vllm-performance-endpoint.md
@@ -132,13 +132,13 @@ operation:
     tuneConfig:
       metric: "request_throughput" # The metric to optimize
       mode: "max"
-      num_samples: 16
+      num_samples: 20
       search_alg:
         name: optuna
         params:
           sampler: TPESampler
           sampler_parameters:
-            n_startup_trials: 8 # Number of random samples before optimization
+            n_startup_trials: 10 # Number of random samples before optimization
 ```
 
 Create the operation with:
@@ -151,9 +151,9 @@ Results will appear as they are measured.
 
 > [!NOTE]
 >
-> Optuna's TPESampler may sample the same point multiple times, especially as
-> the search converges on promising regions. The likelihood increases as the
-> number of points in the space decreases.
+> Optuna's TPESampler samples with replacement and it may sample the same point
+> multiple times. The likelihood increases as the number of points in the space
+> decreases.
 
 ### Monitor the optimization
 
@@ -215,7 +215,7 @@ and the best region is unlikely to be visited.
 >
 > By default, Optuna's TPESampler uses the top 10% of observations (capped at 25
 > samples maximum) as the "good" region. With `n_startup_trials: 8`, the first 8
-> samples are random, then optimization begins. After 80+ samples, the "good"
+> samples are random, then optimization begins. After 250+ samples, the "good"
 > region will be capped at the best 25 observations.
 
 ## Next steps
--- a/website/docs/examples/vllm-performance-endpoint.md
+++ b/website/docs/examples/vllm-performance-endpoint.md
@@ -27,7 +27,7 @@
 > - Install the following Python packages:
 >
 > ```bash
-> pip install hyperopt
+> pip install optuna
 > pip install ado-ray-tune
 > pip install ado-vllm-performance
 > ```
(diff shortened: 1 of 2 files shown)


website/docs/examples/vllm-performance-endpoint.md
```diff
@@ -135,24 +134,26 @@ operation:
...
 ```
 
 Results will appear as they are measured.
 
 > [!NOTE]
 >
-> Hyperopt samples with replacement so you may see the same points sampled
-> twice. The likelihood increases as the number of points in the space decreases
-> The likelihood increase as number of points in the space decreases
+> Optuna's TPESampler may sample the same point multiple times, especially as
```
michael-johnston: Do you have reference to why Optuna samples multiple times as search converges?

website/docs/examples/vllm-performance-endpoint.md
```diff
@@ -215,9 +213,10 @@ and the best region is unlikely to be visited.
 
 > [!TIP]
 >
-> The number of samples hyperopt will use for first guess of good region is
-> (n_initial_points)\*gamma -> in above case 2, the other will be used for the
-> "bad" region
+> By default, Optuna's TPESampler uses the top 10% of observations (capped at 25
+> samples maximum) as the "good" region. With `n_startup_trials: 8`, the first 8
+> samples are random, then optimization begins. After 80+ samples, the "good"
```
michael-johnston: I don't get the "After 80+ samples statement."

If  `n_startup_trials: 8` I would expect top 10% to round to 1. Then at 80 samples it will be using 8 samples for top.

From 250 samples onwards it will start using just 25. So if you e.g. did 500 samples it would be top 5%.

website/docs/examples/vllm-performance-endpoint.md
```diff
@@ -135,24 +134,26 @@ operation:
...
       num_samples: 16
       search_alg:
-        name: hyperopt
-        n_initial_points: 8 #Number of points to sample before optimizing
-        gamma: 0.25 #The top gamma fraction of measured values are considered "good"
+        name: optuna
+        params:
+          sampler: TPESampler
+          sampler_parameters:
+            n_startup_trials: 8 # Number of random samples before optimization
```
michael-johnston: If gamma is 10% with optuna then n_startup_trials maybe should be 20? This would give 2 trials to start using for good region, matching what hyperopt had (it was using 25% of 8). 

or maybe for the example 10 is better for speed - although it only will use one sample for top %

michael-johnston (changes requested):

website/docs/examples/vllm-performance-endpoint.md
christian-pinto: It's just a matter of wrong rewording that I have missed,. The re-sampling is a by product of the sampler, TPE in this case, and not of optuna vs hyperopt. In fact we had a similar statement already with hyperopt. 

The faulty statement is:

`especially as the search converges on promising regions`

that I am about to fix. Thanks for spotting it.

website/docs/examples/vllm-performance-endpoint.md
christian-pinto: yes 80+ should be 250+

michael-johnston (approved):

PR #858, opened 2026-04-15
feat(utilities): enable auto width detection for rich renderables

AlessandroPomponio:
# Summary

Enhanced the `render_to_string` utility function to support automatic width detection for renderables, preventing content truncation when rendering to files. This change simplifies the API by allowing the function to automatically use the renderable's width instead of requiring manual width specification.

# Files Changed

📄 `orchestrator/utilities/rich.py`

Added an `auto_width` parameter to the `render_to_string` function. When enabled, the function automatically detects and uses the renderable's width attribute (if available), preventing content truncation. This parameter takes precedence over the manual `width` parameter. The implementation checks if the renderable has a `width` attribute and uses it when `auto_width=True`.

📄 `tests/utilities/cli_rendering.py`

Updated the `render_ado_resources_to_cli_output` function to use the new `auto_width=True` parameter instead of manually passing `width=table.width`. This simplifies the function call and demonstrates the intended usage of the new feature.

### DIFF
--- a/orchestrator/utilities/rich.py
+++ b/orchestrator/utilities/rich.py
@@ -200,16 +200,27 @@ def dataframe_to_rich_table(
     return table
 
 
-def render_to_string(renderable: "RenderableType", width: int | None = None) -> str:
+def render_to_string(
+    renderable: "RenderableType",
+    width: int | None = None,
+    auto_width: bool = False,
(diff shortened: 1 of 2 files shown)


christian-pinto (approved):
LGTM

PR #859, opened 2026-04-15
feat(cli): allow using --output-file with all output types 

### LINKED ISSUE #788: feat: unify CLI output format flag and make it follow conventions established in other tools
## Is your feature request related to a problem? Please describe.

The `--output-format` flag (and related output flags) are inconsistent and confusing across commands:

- The flag is named differently across commands: `--output` (`ado get`), `--output-format` (`ado show entities/requests/results`), `--format` (`ado show summary`).
- `ado show entities` has no short form `-o`, unlike the other show commands — (minor bug)
- The format value for "rich terminal table" is called `default` in `ado get`, `console` in show commands, and `table` in `ado show summary`.
- Most frustratingly, passing `--output-format csv` or `--output-format json` **writes to an auto-generated file in the current directory** rather than printing to stdout. This makes piping impossible (`ado show results ... -o csv | python analyse.py` does not work), creates files without explicit user intent, and conflates two orthogonal concerns — output format and output destination.
    - This behaviour is very complicated to explain to agents via SKILLS who assume existing conventions established by other similar tools (e.g. see below) hold

## Describe the solution you'd like

1. **Unify the flag name** to `--output` / `-o` across all commands.
2. **Standardize format names**: use `table` consistently for the rich terminal rendering (currently `default`/`console`/`table`). Other formats (`json`, `yaml`, `csv`, `md`) stay as-is.
3. **Send all data to stdout**: `--output csv` should print CSV to stdout, not write a file. Users redirect with the shell as expected:
   ```sh
   ado show results operation <id> -o csv > results.csv
   ado show results operation <id> -o json | jq '.[]'
   ```
4. **Add an explicit `--output-file <path>` flag** as a separate, optional destination override for users who want a named file without shell redirection. It works with any format and prints `✔ Output saved as <path>` to stderr. 
5. **Strict stdout/stderr discipline**: stdout carries all data output; stderr carries only diagnostics (errors, warnings, progress spinners, hints, empty-data notices, and the `--output-file` confirmation).

## Describe alternatives you've considered

- **Keep auto file-writing but make the filename configurable** (e.g. `--output-file`). Rejected because it still breaks piping — the fundamental Unix composability issue remains unless data goes to stdout by default.
- **Keep separate flag names per command**. Rejected because it adds unnecessary cognitive overhead and is inconsistent with tools users already know (`kubectl`, `helm`, `gh` all use `--output`/`-o` uniformly).

## Additional context

This aligns the CLI with established conventions from comparable tools:

- `kubectl`: `--output`/`-o` everywhere; all data to stdout; users redirect with `>`; stderr for errors/warnings only.
- `helm`: `--output` with `table`/`json`/`yaml` to stdout.
- `gh`: `-F`/`--format` with machine formats to stdout.

Key files involved in the implementation:

- `orchestrator/cli/models/types.py` — format enums
- `orchestrator/cli/commands/show_*.py` and `get.py` — flag definitions
- `orchestrator/cli/utils/output/dataframes.py` — `df_to_output`, the central place where file-writing currently happens
- `orchestrator/cli/resources/discovery_space/show_summary.py`
- `website/docs/getting-started/ado.md` — documentation

### LINKED ISSUE #839: refactor(cli): remove arbitrary restrictions to output type when using --output-file
At the moment we don't allow using --output-file with formats like TABLE, NAME, etc. but there's no real reason to do so

AlessandroPomponio:
# Summary

This pull request removes redundant output format validation logic from CLI commands and enhances output file handling to support writing table format to files. Previously, validation prevented using `--output-file` with certain display formats (like `table`, `console`, `default`, and `name`), but this restriction has been removed to allow more flexible output options.

Resolves #839 and closes #788

# Files Changed

📄 `orchestrator/cli/commands/get.py`

Removed validation that prevented using `--output-file` with `DEFAULT` and `NAME` output formats. This restriction is no longer necessary as the output handlers now properly support file output for all formats.

📄 `orchestrator/cli/commands/show_entities.py`

Removed validation that prevented using `--output-file` with `TABLE` output format. File output is now supported for table format.

📄 `orchestrator/cli/commands/show_requests.py`

Removed validation that prevented using `--output-file` with `TABLE` output format for the show requests command.

📄 `orchestrator/cli/commands/show_results.py`

Removed validation that prevented using `--output-file` with `TABLE` output format for the show results command.

📄 `orchestrator/cli/commands/show_summary.py`

Removed validation that prevented using `--output-file` with `MARKDOWN` and `TABLE` output formats for the show summary command.

📄 `orchestrator/cli/utils/output/dataframes.py`

Enhanced the `df_to_output` function to support writing table format to files. When outputting to a file, tables are now rendered to string format using `render_to_string` with automatic width. Column truncation is automatically disabled when writing to files. Also added proper handling for empty JSON output.

📄 `orchestrator/cli/utils/resources/handlers.py`

Refactored output handling in multiple format handlers (`_handle_name_format`, `_handle_table_format`) to use the centralized `_write_or_print_output` helper function. Enhanced `_write_or_print_output` to accept both string content and rich renderables, automatically converting renderables to strings when writing to files. This ensures consistent file output behavior across all format types.

### DIFF
--- a/orchestrator/cli/utils/resources/handlers.py
+++ b/orchestrator/cli/utils/resources/handlers.py
@@ -8,6 +8,7 @@ import pydantic
 import rich.rule
 import typer
 import yaml
+from rich.console import RenderableType
 from rich.status import Status
 
 from orchestrator.cli.models.types import (
@@ -106,17 +107,19 @@ def _handle_name_format(
             else "IDENTIFIER"
         )
         if identifier_column in dataframe.columns:
-            for identifier in dataframe[identifier_column]:
-                console_print(identifier)
+            output = "\n".join(
+                str(identifier) for identifier in dataframe[identifier_column]
+            )
+            _write_or_print_output(output, parameters.output_file)
         return
 
     # If resources provided, extract identifiers
     if resources is not None:
         if isinstance(resources, list):
-            for resource in resources:
-                console_print(resource.identifier)
+            output = "\n".join(resource.identifier for resource in resources)
         else:
-            console_print(resources.identifier)
+            output = resources.identifier
+        _write_or_print_output(output, parameters.output_file)
         return
 
     # Otherwise use efficient DB query
@@ -141,7 +144,7 @@ def _handle_name_format(
                     resource_id=parameters.resource_id, kind=resource_type
                 )
             status.stop()
-            console_print(parameters.resource_id)
+            _write_or_print_output(parameters.resource_id, parameters.output_file)
         else:
             # Multiple resources: use efficient getResourceIdentifiersOfKind
             identifiers_df = sql_store.getResourceIdentifiersOfKind(
@@ -154,8 +157,10 @@ def _handle_name_format(
                 console_print(ADO_INFO_EMPTY_DATAFRAME, stderr=True)
                 return
             # Output one identifier per line
-            for identifier in identifiers_df["IDENTIFIER"]:
-                console_print(identifier)
+            output = "\n".join(
+                str(identifier) for identifier in identifiers_df["IDENTIFIER"]
+            )
+            _write_or_print_output(output, parameters.output_file)
 
 
 def _handle_table_format(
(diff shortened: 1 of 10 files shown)


orchestrator/cli/utils/resources/handlers.py
```diff
@@ -106,17 +107,19 @@ def _handle_name_format(
             else "IDENTIFIER"
         )
         if identifier_column in dataframe.columns:
```
michael-johnston: What is expected to happen if identifier_column is NOT in dataframe,columns

In current code it's silently ignored - is this correct?

AlessandroPomponio: You're right, it's better to raise an error - if it isn't, it's likely a programming error

michael-johnston (approved):

orchestrator/cli/utils/resources/handlers.py
AlessandroPomponio: https://github.com/IBM/ado/pull/859/commits/777c6a150b2d33ebb94994df92a9737fe6247c18

PR #865, opened 2026-04-16
docs(agents): use updated output flags in skills

michael-johnston:
- Document -o/--output as format vs --output-file for get/show, and template -o as path
- Fix query-ado-data: show entities/requests/results use csv|json|table, not console
- Prefer --output-file in examining-* workflows; fix datacontainer get argument order
- Fix examining-discovery-spaces typo (corrupted)"


michael-johnston:
@AlessandroPomponio I noticed that `ado template` now is different. It has --output instead of --output-file, and autogenerated file names. It's probably worth changing it so everything is the same unless strong reason why it should be different?

.cursor/skills/using-ado-cli/SKILL.md
```diff
@@ -29,6 +29,19 @@ uv run ado [COMMAND] [SUBCOMMAND1] [SUBCOMMAND2] --help
...
+
+For `ado get` and `ado show` subcommands:
+
+- `-o` / `--output` selects the **output format** (for example `yaml`, `table`,
+  `csv`, or `json`; allowed values depend on the command — use `--help`).
+- `--output-file PATH` writes that formatted output to **PATH** instead of
+  stdout. Prefer this over shell redirection for large output so encoding and
+  table rendering stay consistent.
+
+For `ado template` only, `-o` / `--output` is the **path for the generated
```
michael-johnston:
```suggestion
```

.cursor/skills/using-ado-cli/SKILL.md
```diff
@@ -29,6 +29,19 @@ uv run ado [COMMAND] [SUBCOMMAND1] [SUBCOMMAND2] --help
...
+template file**. That command does not use `--output-file`.
```
michael-johnston:
```suggestion
```

AlessandroPomponio (approved):
LGTM

My only observation is whether agents get tripped by referring to --output-file as a path (basically, if they provide a folder instead of a file). But that can be easily fixed later if that's the case

PR #868, opened 2026-04-17
build: include timestamp in development wheels

michael-johnston:
Avoid issues especially with local uncommitted edits and using `ado --remote`


michael-johnston:
@AlessandroPomponio core pyproject updated

AlessandroPomponio:
We should probably update the examples we have for the pyprojects, such as https://github.com/IBM/ado/blob/e1abfe871a42f71ac9439e28ded6bcc4ed48b6a5/.cursor/rules/plugin-development.mdc#L34

AlessandroPomponio (approved):

PR #878, opened 2026-04-21
refactor(core): remove support for actuators decorated with ray.remote

### LINKED ISSUE #704: refactor: remove support for actuators decorated with ray.remote
#185 introduced dynamic decorator of Actuators - meaning actuators did not have to have @ray.remote. 
It also added a function to remove the decoration from any actuators that still had it for compatibility. 

#564 removed this decorator from all in-repo actuators. 

Hence it's likely this extraction function is no longer needed and should be removed by a TBD date.

AlessandroPomponio:
Closes #704


michael-johnston (commented):
Does this not also affect operators? 

in the explore decorator in operators/collections.py

AlessandroPomponio:
> Does this not also affect operators?
> 
> in the explore decorator in operators/collections.py

This is in the ActuatorRegistry, and it's the only instance where that function was called. If you have anything in particular you'd like me to try, I can

michael-johnston:
ok seems like I deprecated support  for operators decorated with ray.remote in #805

michael-johnston (approved):

Issue #884, opened 2026-04-21
bug: Unable to run vllm performance in uv native environment


christian-pinto:
## Issue Description

When running the vllm-performance actuator, a ray task is generated for executing the benchmark and  dependencies (e.g., vllm) are installed on the fly.

However, due to a bug in Ray, even though uv is installed in the venv it is not found and Ray tries to install it with pip. Pip is in turn not found and the execution hangs indefinitely.

Only inspecting the ray logs one can figure it out

```
2026-04-21 10:27:16,112 ERROR runtime_env_agent.py:404 -- Failed to create runtime env {"uv": {"packages": ["ado-vllm-performance[vllm]"], "uv_check": false, "uv_pip_install_options": ["--no-cache"]}}.
Traceback (most recent call last):
  File "/workspace/venvs/test/lib/python3.12/site-packages/ray/_private/runtime_env/agent/runtime_env_agent.py", line 396, in _create_runtime_env_with_retry
    runtime_env_context = await asyncio.wait_for(
                          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/tasks.py", line 520, in wait_for
    return await fut
           ^^^^^^^^^
  File "/workspace/venvs/test/lib/python3.12/site-packages/ray/_private/runtime_env/agent/runtime_env_agent.py", line 354, in _setup_runtime_env
    await create_for_plugin_if_needed(
  File "/workspace/venvs/test/lib/python3.12/site-packages/ray/_private/runtime_env/plugin.py", line 255, in create_for_plugin_if_needed
    size_bytes = await plugin.create(uri, runtime_env, context, logger=logger)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/venvs/test/lib/python3.12/site-packages/ray/_private/runtime_env/uv.py", line 316, in create
    uv_dir_bytes = await task
                   ^^^^^^^^^^
  File "/workspace/venvs/test/lib/python3.12/site-packages/ray/_private/runtime_env/uv.py", line 296, in _create_for_hash
    await UvProcessor(
  File "/workspace/venvs/test/lib/python3.12/site-packages/ray/_private/runtime_env/uv.py", line 206, in _run
    await self._install_uv_packages(
[...]

PR #886, opened 2026-04-21
fix(operators): prevent infinite recursion when calling non-explore operators

michael-johnston:
Bug reported in #884, bug created in #805

The operator function decorators wrap an operator function, so that when the wrapper is called,  the general operator orchestration function is called to execute the wrapped function (the operator function the developer writes and decorates)

Pre #805 an internal function was receiving the un-decorated, original operator function and calling it. 
in #805 this functions interface was changed so it received the decorated function, however the call was not updated to account for this (it should have accessed the wrapped function and called it).

As a result that internal call called the wrapper, which called the outermost orchestration function again and a recursive loop kicks off. 

operator_wrapper() -> general_orchestration(operator metadata) -> ... -> Internal Function(operator wrapper) -> operator_wrapper() 

Instead of 

operator_wrapper() -> general_orchestration(operator metadata) -> ... -> Internal Function(operator wrapper)  -> unwrap(operator_wrapper)()

### DIFF
--- a/orchestrator/modules/operators/_general_orchestration.py
+++ b/orchestrator/modules/operators/_general_orchestration.py
@@ -1,6 +1,7 @@
 # Copyright IBM Corporation 2025, 2026
 # SPDX-License-Identifier: MIT
 
+import inspect
 import logging
 import typing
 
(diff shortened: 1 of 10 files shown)


AlessandroPomponio (approved):
LGTM thanks

PR #894, opened 2026-04-22
build(core): add entry point for random walk operator

michael-johnston:
Prior to use random walk operator from collection, users would have to import the random walk module to trigger its addition to the collection


AlessandroPomponio (approved):
LGTM thanks

PR #896, opened 2026-04-22
fix(vllm_performance): do not override env if existing in deployment template

christian-pinto:
The actuator code is currently overriding the container env configuration even is something is already present int the deployment template. Instead, if something is the template we should just extend with any additional variable.

This PR fixes the above issue.


christian-pinto:
@mgazz please confirm it now works for you

mgazz:
Env vars processed as expected. 👍

mgazz (approved):

PR #900, opened 2026-04-22
fix(metastore): correctly handle intermediate fields with underscores on sqlite

### LINKED ISSUE #897: bug(metastore): simulated JSON_CONTAINS on SQLite doesn't handle fields in snake case
## Issue Description

SQLite's `json_tree` function quotes field names containing underscores in the
`path` column, but the current implementation in
`orchestrator/metastore/sql/statements.py` does not account for this behavior.
This causes queries with snake_case field names to fail to match any results,
blocking the PEP8 naming migration.

### How to reproduce

1. Create a resource with nested snake_case field names:

   ```python
   {
       "config": {
           "outer_field": {
               "inner_field": {"value": "test"}
           }
       }
   }
   ```

2. Query for the nested field using the CLI:

   ```bash
   ado get operations -q config.outer_field.inner_field.value=test
   ```

3. Observe that **no results are returned** even though the data exists

**Root Cause:** The generated pattern `$.config%.outer_field%.inner_field%`
looks for `.outer_field` but SQLite's `json_tree` uses `."outer_field"` (with
quotes).

## Expected behaviour

Queries with snake_case field names should successfully match nested structures
containing underscored fields.

### SQLite json_tree Behavior

When querying a JSON document with underscored field names, SQLite's `json_tree`
quotes them:

```
Actual paths in json_tree:
  - $.config."outer_field"
  - $.config."outer_field"."inner_field"

Current pattern (no quoting): $.config%.outer_field%.inner_field%
Result: 0 matches (fails to find the data)
```

### Affected Code

**File:** `orchestrator/metastore/sql/statements.py`  
**Function:** `check_field_in_sqlite_json_document` (line 52)  
**Specific location:** Lines 195-203

```python
if isinstance(candidate[field], list | dict):
    fragments.extend(
        check_field_in_sqlite_json_document(
            candidate[field], f"{path}%.{field}"  # <-- No quoting here
        )
    )
```

### Impact

This issue affects:

- Querying operations with nested snake_case field names
- Any code using `simulate_json_contains_on_sqlite` for complex queries
- PEP8 naming migration - currently blocked until resolved
- CLI queries using `-q` flag with snake_case fields

### Proposed Solution

Apply a simple quoting fix that quotes field names containing underscores:

```python
if isinstance(candidate[field], list | dict):
    # SQLite quotes field names containing underscores in json_tree paths.
    # Quote the field name if it contains an underscore to match SQLite's behavior.
    if "_" in field:
        field_pattern = f'"{field}"'
    else:
        field_pattern = field

    fragments.extend(
        check_field_in_sqlite_json_document(
            candidate[field], f"{path}%.{field_pattern}"
        )
    )
```

Similarly for scalar fields at line 212:

```python
if isinstance(candidate[field], _ScalarType):
    # Quote field names with underscores
    if "_" in field:
        field_key = f'"{field}"'
    else:
        field_key = field

    fragments.append(
        f"{preamble} F.path LIKE '{path}%' AND "
        f"F.key = '{field_key}' AND "
        f"F.value {_searchable_scalar_value_for_query_string(candidate[field])}"
    )
```

### Known Limitation

This fix may produce false positives in complex nested structures where the same
field name appears at different nesting levels. For example, the pattern
`$.config%."outer_field"%."inner_field"%` will match both:

- `$.config."outer_field"."inner_field"` (intended)
- `$.config.something."outer_field".somethingelse."inner_field"` (false
  positive)

This is acceptable as it's similar to the existing wildcard behavior for arrays,
and false positives are filtered by the INTERSECT logic in
`simulate_json_contains_on_sqlite`.

### Follow-up

A more robust solution using exact pattern matching with OR clauses can be
implemented in a future enhancement if the false positive rate becomes
problematic.

AlessandroPomponio:
# Summary

This PR fixes a bug in SQLite JSON querying where field names containing underscores were not being properly handled. SQLite's `json_tree` function quotes field names with underscores in paths, which was causing query mismatches. The fix adds logic to detect and quote such field names when constructing SQL queries.

Fixes #897 

# Files Changed

📄 `orchestrator/metastore/sql/statements.py`

Added handling for SQLite's quoting behavior of field names containing underscores in `json_tree` paths. The `check_field_in_sqlite_json_document` function now:
- Detects field names with underscores and wraps them in quotes when building path patterns
- Adds documentation explaining the quoting behavior and potential edge cases
- Clarifies that the `key` column in `json_tree` is never quoted, only intermediate fields in the `path` column

📄 `tests/ado/get/test_ado_get.py`

Adds comprehensive test coverage for querying nested JSON fields with underscores. The tests verify:
- Querying deeply nested fields with underscores (e.g., `outer_field.inner_field.test_value`)
- Querying with JSON objects containing underscore fields
- Querying mixed underscore and simple field names
Imports `ActuatorConfigurationResource` and creates a test fixture from the new YAML file.

📄 `tests/resources/actuatorconfiguration/mock-ac-with-snake-case.yaml`

New test fixture file containing an `ActuatorConfigurationResource` with nested fields using snake_case naming (underscores). Includes fields like `outer_field`, `inner_field`, `test_value`, `another_field`, and `simple_field` to test various underscore scenarios.

### DIFF
--- a/orchestrator/metastore/sql/statements.py
+++ b/orchestrator/metastore/sql/statements.py
@@ -60,6 +60,11 @@ def check_field_in_sqlite_json_document(
     and constructs SQL subqueries that can be used to filter rows produced by SQLite's json_tree
     function based on whether the specified fields and values exist at the given JSON path.
 
+    Note: SQLite's json_tree quotes field names containing underscores. This function handles
+    that by quoting such field names in the generated patterns. This may produce false positives
+    in complex nested structures where the same field name appears at different nesting levels,
+    but these are filtered by the INTERSECT logic in simulate_json_contains_on_sqlite.
+
     Args:
         candidate (dict | list | str | int | float): The JSON structure or scalar value to match against.
             - If a scalar (str, int, float), generates a simple query checking for value presence.
@@ -196,9 +201,14 @@ def check_field_in_sqlite_json_document(
             # The use of % in the path is because json_tree will add list items in the path.
             # (e.g., $.config.entitySpace[2].propertyDomain). As we can't know for sure
             # whether a field is a list or not, we use the LIKE operator and a wildcard (%)
+
+            # SQLite quotes field names containing underscores in json_tree paths.
+            # Quote the field name if it contains an underscore to match SQLite's behavior.
+            field_pattern = f'"{field}"' if "_" in field else field
+
             fragments.extend(
                 check_field_in_sqlite_json_document(
-                    candidate[field], f"{path}%.{field}"
+                    candidate[field], f"{path}%.{field_pattern}"
                 )
             )
             continue
@@ -207,6 +217,9 @@ def check_field_in_sqlite_json_document(
         # with an array field, for which the path would contain
         # the index.
         if isinstance(candidate[field], _ScalarType):
+            # Note: We do NOT quote the field name in F.key because the 'key' column
+            # in json_tree is never quoted. Only intermediate fields in the 'path'
+            # column are quoted when they contain underscores.
             fragments.append(
                 f"{preamble} F.path LIKE '{path}%' AND "
                 f"F.key = '{field}' AND "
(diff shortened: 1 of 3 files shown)


christian-pinto (approved):
LGTM

PR #906, opened 2026-04-23
feat(cli): support --patch and --patch-file flags in ado edit for non-interactive metadata patches

### LINKED ISSUE #825: feat(cli): add support for a --patch flag in ado edit
If you want an agent to edit metadata of created resources (add labels, enhance description etc.) it can't use `ado edit` as this command only support interactive editing of the fields. 

As a result on an instruction to update resource metadata an agent will likely develop a once-off few hundred line python script to do the job

It would be good if `ado edit` supported setting the fields via CLI e.g. providing a patch YAML or exposing field like --set-name.

michael-johnston:
Resolves #825


orchestrator/cli/commands/edit.py
```diff
@@ -32,6 +35,30 @@
...
+    for member in AdoEditSupportedEditors:
+        if member.value == token:
+            return member
+    valid = ", ".join(sorted(m.value for m in AdoEditSupportedEditors))
+    console_print(
+        f"{ERROR}Invalid --editor (or $ADO_EDITOR) value {raw!r}. "
+        f"Use one of: {valid}.",
+        stderr=True,
+    )
+    raise typer.Exit(1)
```
AlessandroPomponio: Why is this function here? This is typer's default behaviour

orchestrator/cli/commands/edit.py
```diff
@@ -52,9 +79,32 @@ def edit_resource(
...
+                "metadata (non-interactive; cannot be combined with --editor)."
+            ),
+            file_okay=True,
+            dir_okay=False,
+            exists=True,
+            readable=True,
+            resolve_path=True,
+            show_default=False,
+        ),
+    ] = None,
```
AlessandroPomponio: We should behave like `oc` and have these two flags:

```
    -p, --patch='':
        The patch to be applied to the resource JSON file.

    --patch-file='':
        A file containing a patch to be applied to the resource.
```

The behaviour we should support is the default strategic merge (basically update the dictionary) and require patches in input by default, not in a file

orchestrator/cli/utils/pydantic/metadata_merge.py
```diff
@@ -0,0 +1,38 @@
...
+                if left is None or not isinstance(left, dict):
+                    left = {}
+                out["labels"] = {**left, **value}
+            else:
+                out["labels"] = value
+        elif isinstance(value, dict) and isinstance(out.get(key), dict):
+            out[key] = {**out[key], **value}
+        else:
+            out[key] = value
+    return out
```
AlessandroPomponio: i don't think we need this function, we can just call dict.update - the only bit we need to handle differently is labels, where we need to do dict['labels'].update separately and pop the key out before updating the rest

AlessandroPomponio (dismissed):

.cursor/skills/using-ado-cli/SKILL.md
```diff
@@ -113,6 +112,24 @@ uv run ado create operation -f operation.yaml
...
+### ado edit
+
+Updates **metadata** (name, description, labels, etc.) for metastore resources.
+
+```bash
+# Interactive (default editor: nano, or $ADO_EDITOR)
+uv run ado edit space SPACE_ID
+
+# Non-interactive: default is an inline patch (-p / --patch), like oc
+uv run ado edit space SPACE_ID -p "labels: { team: research }"
```
AlessandroPomponio, line 125: This looked a bit weird to me at the beginning, because I assume it's inline YAML rather than the inline JSON that I was expecting

.cursor/skills/using-ado-cli/SKILL.md
```diff
@@ -143,6 +160,30 @@ uv run ado describe space SPACE_ID
...
+  a potentially long data fetch, avoiding a silent failure at the end.
+- **Stdout pollution**: if any log lines or warnings are mixed into stdout (e.g.
+  when another tool in the pipeline writes to stdout), a redirect captures that
+  noise alongside the data. `--output-file` writes only the formatted output to
+  the file; logs continue to go to stderr.
+
+```bash
+uv run ado show entities operation OPERATION_ID -o csv --output-file entities.csv
+```
+
```
AlessandroPomponio, line 190: Is this worth having? 
I would suggest asking it to always use --output-file no matter what

orchestrator/cli/commands/edit.py
```diff
@@ -32,6 +35,17 @@
...
 
+def _parse_ado_edit_editor_name(value: str) -> AdoEditSupportedEditors:
+    """Map *value* from :class:`HiddenPluralChoice` to the editor enum member."""
+    token = value.removesuffix("s")
+    for m in AdoEditSupportedEditors:
+        if m.value == token:
+            return m
+    raise RuntimeError(
+        "HiddenPluralChoice should have already validated the editor"
+    )  # pragma: no cover
```
AlessandroPomponio: This shouldn't be using hiddenpluralchoice

orchestrator/cli/commands/edit.py
```diff
@@ -52,9 +66,43 @@ def edit_resource(
...
+        str,
+        typer.Option(
+            "--editor",
+            envvar="ADO_EDITOR",
+            help=(
+                "The editor to use to edit metadata (interactive mode only; "
+                "not with --patch or --patch-file)."
+            ),
+            show_default=AdoEditSupportedEditors.NANO.value,
+            click_type=HiddenPluralChoice(AdoEditSupportedEditors),
```
AlessandroPomponio: This shouldn't use HiddenPluralChoice
```suggestion
            click_type=AdoEditSupportedEditors,
```

orchestrator/cli/commands/edit.py
```diff
@@ -52,9 +66,43 @@ def edit_resource(
...
-        typer.Option(envvar="ADO_EDITOR", help="The editor to use to edit metadata"),
+        str,
+        typer.Option(
+            "--editor",
+            envvar="ADO_EDITOR",
+            help=(
+                "The editor to use to edit metadata (interactive mode only; "
+                "not with --patch or --patch-file)."
+            ),
+            show_default=AdoEditSupportedEditors.NANO.value,
```
AlessandroPomponio:
```suggestion
```

orchestrator/cli/commands/edit.py
```diff
@@ -68,12 +116,48 @@ def edit_resource(
...
+    if non_interactive and ctx.get_parameter_source("editor") in (
+        ParameterSource.COMMANDLINE,
+        ParameterSource.PROMPT,
+    ):
+        console_print(
+            f"{ERROR}The options --patch / -p and --patch-file "
+            "may not be used with an explicit --editor flag.",
+            stderr=True,
+        )
+        raise typer.Exit(1)
```
AlessandroPomponio: I wonder if it's even worth doing this instead of just saying "--editor is ignored if --patch or --patch-file are specified" or something along the lines

orchestrator/cli/utils/resources/handlers.py
```diff
@@ -444,27 +470,77 @@ def handle_edit_resource_metadata(
...
-            suppress_success_message=True,
-        )
-
+    if metadata_path is not None or metadata_patch is not None:
+        if metadata_path is not None and metadata_patch is not None:
+            console_print(
+                f"{ERROR}Use only one of --patch / -p and --patch-file.",
+                stderr=True,
+            )
+            raise typer.Exit(1)
```
AlessandroPomponio: This is already checked in orchestrator/cli/commands/edit.py L125-130

orchestrator/cli/utils/resources/handlers.py
```diff
@@ -426,11 +426,37 @@ def print_related_resources(
...
+            out["labels"] = None
+        elif isinstance(v, dict):
+            left = out.get("labels")
+            if not isinstance(left, dict):
+                left = {}
+            out["labels"] = {**left, **v}
+        else:
+            out["labels"] = v
+    out.update(rest)
+    return out
```
AlessandroPomponio: To me this is way more understandable:


```suggestion
    merged = dict(base)
    overrides = dict(patch)
    if "labels" in overrides:
        new_labels = overrides.pop("labels")
        old_labels = merged.get("labels", {})
        if new_labels is None:
            merged["labels"] = None
        elif old_labels is None:
            merged["labels"] = new_labels
        else:
            merged["labels"] = old_labels | new_labels

    return merged | overrides
```

orchestrator/cli/utils/resources/handlers.py
```diff
@@ -444,27 +470,77 @@ def handle_edit_resource_metadata(
...
+        except (OSError, yaml.YAMLError) as e:
+            source: pathlib.Path | str
+            if metadata_path is not None:
+                source = metadata_path
+            else:
+                source = "inline --patch / -p"
+            console_print(
+                f"{ERROR}Could not read metadata from {source}:\n{e}",
+                stderr=True,
+            )
```
AlessandroPomponio: I would try to handle all of these uniformly and do some additional validation at the start (e.g., ensure that `labels` isn't a list


```suggestion
            raw = yaml.safe_load(
                metadata_patch
                if metadata_patch is not None
                else metadata_path.read_text()
            )
            _ = ConfigurationMetadata.model_validate(raw)
        except (OSError, yaml.YAMLError, ValueError) as e:
            console_print(
                f"{ERROR}The provided metadata was invalid: {e}",
                stderr=True,
            )
```

orchestrator/cli/utils/resources/handlers.py
```diff
@@ -444,27 +470,77 @@ def handle_edit_resource_metadata(
...
-            new_metadata = ConfigurationMetadata.model_validate(
-                yaml.safe_load(file.read_text())
+            merged = strategic_merge_configuration_metadata(
+                base=base_dict, patch=patch_dict
             )
+            new_metadata = ConfigurationMetadata.model_validate(merged)
         except pydantic.ValidationError as e:
-            console_print(f"{ERROR}The updated metadata was invalid: {e}", stderr=True)
+            console_print(f"{ERROR}The merged metadata was invalid: {e}", stderr=True)
             raise typer.Exit(1) from e
```
AlessandroPomponio: With the additional validation that we had we can simplify this
```suggestion
        if raw is None:
            new_metadata = {}
        else:
            base_dict = resource.config.metadata.model_dump()
            new_metadata = strategic_merge_configuration_metadata(
                base=base_dict, patch=raw
            )
```

orchestrator/cli/utils/resources/handlers.py
```diff
@@ -444,27 +470,77 @@ def handle_edit_resource_metadata(
...
+                new_metadata = ConfigurationMetadata.model_validate(
+                    yaml.safe_load(file.read_text())
+                )
+            except pydantic.ValidationError as e:
+                console_print(
+                    f"{ERROR}The updated metadata was invalid: {e}", stderr=True
+                )
+                raise typer.Exit(1) from e
 
     resource.config.metadata = new_metadata
```
AlessandroPomponio: The final validation is common to both paths so might as well reuse it
```suggestion
            new_metadata = yaml.safe_load(file.read_text())

    try:
        resource.config.metadata = ConfigurationMetadata.model_validate(
            new_metadata
        )
    except pydantic.ValidationError as e:
        console_print(f"{ERROR}The updated metadata was invalid: {e}", stderr=True)
        raise typer.Exit(1) from e
```

AlessandroPomponio (dismissed):

.cursor/skills/using-ado-cli/SKILL.md
michael-johnston: Is there anything to be done?

.cursor/skills/using-ado-cli/SKILL.md
michael-johnston: It's little problematic since the models heavily prefer to pipe everything. So it seems  easier to let them do what they want and provide this for fallback if they encounter issues.

AlessandroPomponio: Okay then we should probably implement pipe recognition because the real reason to use `--output-file` instead of `--output something > file` is that I make sure table outputs aren't truncated
Of course, this is not a problem for non-table output formats, but I can imagine it could lead to weird failures

michael-johnston: Ok so then this means --output-format table is different when --output-file is specified?

AlessandroPomponio: To an extent, yes: by default we request to not truncate only identifier columns (e.g., `IDENTIFIER` in ado get operations), while when using --output-file I override that so no column gets truncated

orchestrator/cli/commands/edit.py
michael-johnston: What is the change that's required?

orchestrator/cli/utils/resources/handlers.py
michael-johnston: Can we just remove the if block?

.cursor/skills/using-ado-cli/SKILL.md
michael-johnston: I will add a line for now about the truncation.

.cursor/skills/using-ado-cli/SKILL.md
AlessandroPomponio: I would maybe add "inline YAML/JSON patch"

orchestrator/cli/commands/edit.py
AlessandroPomponio: Completely remove the function

orchestrator/cli/commands/edit.py
```diff
@@ -52,9 +66,42 @@ def edit_resource(
...
+        str,
+        typer.Option(
+            "--editor",
+            envvar="ADO_EDITOR",
+            help=(
+                "The editor to use to edit metadata (interactive mode only; "
+                "not with --patch or --patch-file)."
+            ),
+            click_type=AdoEditSupportedEditors,
+        ),
```
AlessandroPomponio:
```suggestion
        AdoEditSupportedEditors,
        typer.Option(
            "--editor",
            envvar="ADO_EDITOR",
            help=(
                "The editor to use to edit metadata (interactive mode only; "
                "not with --patch or --patch-file)."
            ),
        ),
```

orchestrator/cli/commands/edit.py
AlessandroPomponio: This would also get rid of the difference between the methods that this PR introduces (where editor must be set to None if "non interactive"

orchestrator/cli/utils/resources/handlers.py
AlessandroPomponio: I'd say so, yes

orchestrator/cli/commands/edit.py
michael-johnston: So for non interactive editor is ignored if set instead of error. No message?

michael-johnston: This also would change the CLI help as a note

AlessandroPomponio: Yeah, I'd just add a note there and in ado.md like for other commands

.cursor/skills/using-ado-cli/SKILL.md
AlessandroPomponio: I guess another option is to also add --no-trunc to the normal ado get call. This should probably be the case for all commands issued by the agent I'm guessing

michael-johnston: Let's see how it gets on. If it's frequently doing `ado get ... >` and then failing due to truncation and falling back to --output-file, we should change it

AlessandroPomponio:
https://github.com/IBM/ado/pull/906/changes#r3148142007 is still missing. I'll address it before it gets merged so we don't lose track.

AlessandroPomponio (approved):
LGTM

christian-pinto (approved):
LGTM

PR #915, opened 2026-04-28
docs: add note to issue template about editable installs

### LINKED ISSUE #851: bug: ado version is misleading in editable install mode
## Issue Description

If ado is installed from source in editable mode, `ado version` reports the version set ad install time. However, if a git pull is performed at a later stage `ado version` will keep reporting the same version, while the code executing might be that of a more recent version. 

### How to reproduce

1. Clone and install ado in editable mode
2. `ado version`
3.  after some time do a `git pull` without reinstalling
4. `ado version` reports the same value as in step 2 while the code executing might be at a newer version.

AlessandroPomponio:
Resolves #851

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,10 +1,9 @@
 ---
 name: Bug report
 about: Create a report to help us improve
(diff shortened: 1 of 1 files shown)


christian-pinto (approved):
LGTM

PR #922, opened 2026-04-28
docs(skills): add .agents

michael-johnston:
The Skills docs recommend .agents (anthropic) and its supported by copilot and codex from their docs.


AlessandroPomponio (approved):
LGTM thanks

PR #931, opened 2026-05-01
feat(vllm_performance): make deployment parameters optional and update to v2

mgazz:
## Summary

Updates vLLM performance actuator experiments to v2 with optional deployment parameters.

## Key Changes

- **Optional parameters**: `max_batch_tokens`, `gpu_memory_utilization`, `cpu_offload`, `max_num_seq` now optional (only passed to vLLM if explicitly set)
- **Smart tensor-parallel-size**: Only set when `n_gpus > 1` (allows vLLM default for single GPU)
- **Version bump**: All experiments updated from v1 to v2
- **Documentation**: Added comprehensive docs explaining property value selection and optional parameters
- **Tests**: Added unit tests for deployment parameter handling


michael-johnston:
Although you will still be able to `get` and `show` spaces with the v1 experiments you won't be able to describe those experiments. It's probably better to keep the old experiments and mark them as deprecated. 

Also it would be good here to understand what changed between v1 and v2 - it's difficult to understand from the description e.g. if  I run a v1 and v2 experiment with same parameters what changes?

michael-johnston:
> Optional parameters: max_batch_tokens, gpu_memory_utilization, cpu_offload, max_num_seq now optional (only passed to vLLM if explicitly set)

I also don't get this - if we don't pass them unless explicitly set how can we ever know what they are? or that they are the same between two runs of a v2 experiment? Or are we talking about experiment parameters here at all?

mgazz:
I think we can close this PR for now. After discussing it, a better approach would be to handle the complexity of changing behaviours and parameters of the benchmark tools across versions within the actuator itself.

With this approach, there is no need to change the experiment version, as these differences will be managed directly by the actuator. In this context, it will be important to clearly communicate any such changes to users with proper documentation and logging.

AlessandroPomponio:
Closing as per the comment above - feel free to re-open if needed @mgazz

PR #935, opened 2026-05-04
feat(cli): support --use-latest in ado get

### LINKED ISSUE #853: feat(cli): add --use-latest to ado get

Fetching the full resource YAML for the most recently created space or operation requires two commands a step to get latest id, then a step to retrieve the YAML e.g.  `ado show related RESOURCE_TYPE --use-latest`,  `ado get RESOURCE_TYPE ID -o yaml` to fetch the YAML. The first step acts as an ID-lookup step and adds friction to a very common workflow (this is regardless if you do `ado get RESOURCE_TYPE` and grep the last line etc.). 

**Describe the solution you'd like.**

Add --use-latest to ado get, consistent with its behaviour on all ado show subcommands and ado describe space. When the flag is provided without an explicit resource ID, it would resolve the most recently created resource of the given type and return its YAML

```
uv run ado get space --use-latest -o yaml
uv run ado get operation --use-latest -o yaml
```

Behaviour should match the existing convention: --use-latest is ignored if an explicit resource ID is also supplied.

**Additional context.**

The --use-latest flag is already supported on `ado show` commands and `ado describe space`. Adding it to `ado get` completes the pattern and removes the only command where a two-step ID-lookup is necessary for the common "examine the latest resource" workflow. The interaction with --query / --label filters could follow the same precedence rule used elsewhere: an explicit ID takes precedence over --use-latest, and --use-latest with filters could either be disallowed or treated as "latest among matching resources".

AlessandroPomponio:
# Summary

This PR adds a `--use-latest` flag to the `ado get` command that allows users to retrieve the most recently created resource of a specified type without needing to know its identifier. When both an explicit resource ID and `--use-latest` are provided, the explicit ID takes precedence.

Resolves #853 

# Files Changed

📄 `orchestrator/cli/commands/get.py`

Added a new `--use-latest` boolean option to the `get_resource` command. When enabled, the flag resolves to the actual resource ID of the most recently created resource of the specified type. Updated help text to reflect the new option, added documentation examples showing usage with spaces and operations, and integrated the resolution logic before the main command execution. The flag is ignored if an explicit resource identifier is also provided.

📄 `orchestrator/cli/models/parameters.py`

Added the `use_latest` boolean field to the `AdoGetCommandParameters` Pydantic model to support passing the new flag through the command parameter structure.

📄 `tests/ado/get/test_ado_get.py`

Added comprehensive test coverage for the new `--use-latest` functionality including:
- `test_get_space_with_use_latest`: Verifies that the latest space is retrieved when using the flag
- `test_get_operation_with_use_latest`: Verifies that the latest operation is retrieved when using the flag
- `test_get_with_use_latest_and_explicit_id`: Ensures explicit IDs take precedence over `--use-latest`
- `test_get_with_use_latest_table_format`: Tests the flag with default table output format
- `test_get_with_use_latest_name_format`: Tests the flag with name-only output format

### DIFF
--- a/.cursor/skills/using-ado-cli/SKILL.md
+++ b/.cursor/skills/using-ado-cli/SKILL.md
@@ -39,7 +39,6 @@ For `ado get` and `ado show` subcommands:
   stdout. Prefer this over shell redirection for large output so encoding and
   table rendering stay consistent.
 
-
 ## Commands That do not exist
 
 These plausible-sounding commands do not exist in ado. Do not write them:
@@ -97,6 +96,12 @@ uv run ado get space SPACE_ID -o yaml
 
 # Or write the same YAML to a file
 uv run ado get space SPACE_ID -o yaml --output-file space.yaml
+
+# Get the latest space as YAML
+uv run ado get space --use-latest -o yaml
+
+# Get the latest operation as YAML
+uv run ado get operation --use-latest -o yaml
 ```
 
 ### ado create
--- a/orchestrator/cli/commands/context.py
+++ b/orchestrator/cli/commands/context.py
@@ -114,6 +114,7 @@ def list_contexts(
         resource_type=AdoGetSupportedResourceTypes.CONTEXT,
         show_deprecated=False,
         show_details=False,
+        use_latest=False,
     )
 
     # NOTE: there will always be at least one context (local)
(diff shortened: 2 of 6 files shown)


christian-pinto (approved):
LGTM

PR #952, opened 2026-05-15
feat(vllm-performance): Install dependencies in benchmark execution worker

christian-pinto:
## High-Level Summary

This PR improves the Ray runtime environment handling for the vLLM Performance actuator by implementing intelligent package inheritance for experiment tasks.

**Key Changes:**

- **Added `_build_experiment_runtime_env()` function**: Constructs experiment-specific runtime environments that inherit packages from the parent job's environment while adding required experiment tools (vllm or guidellm)

- **Smart package management**: Detects if `ado-vllm-performance` is already installed in the job environment and intelligently merges extras (e.g., `[vllm]`, `[guidellm]`) rather than creating conflicting installations

- **Improved isolation**: Ensures experiment tasks have access to both the base job dependencies and their specific tool requirements without package conflicts

This change addresses runtime environment isolation issues when running vLLM performance experiments on Ray clusters, ensuring experiments can access both shared dependencies and their specialized tools.

### DIFF
--- a/orchestrator/utilities/environment.py
+++ b/orchestrator/utilities/environment.py
@@ -2,146 +2,94 @@
 # SPDX-License-Identifier: MIT
 import logging
 import os
-from importlib.metadata import requires
 
 import ray
+from packaging.requirements import Requirement
 
 logger = logging.getLogger(__name__)
 
 
-def _get_dependency_from_package_metadata(
-    package_name: str, extra_name: str
-) -> str | None:
+def extract_package_specs_from_job_env(
+    package_names: list[str],
+) -> dict[str, dict[str, str | None]]:
     """
-    Extract dependency specification from installed package metadata.
+    Parse package specifications from the Ray job's uv environment.
+
+    Given a list of package names, returns information about those packages
+    if they are present in the worker job's uv venv.
 
     Args:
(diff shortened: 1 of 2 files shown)


christian-pinto:
@michael-johnston this fixes the dependency inheritance issue we have discovered with the vllm-performance actuator.

michael-johnston (changes requested):
Can we move the build runtime env function to orchestrator/utilities ? Seems generally useful

 it just needs a param for the packages to search for in the job env. 

Also I think we should use “package” instead of “tool” e.g in the param

christian-pinto:
> Can we move the build runtime env function to orchestrator/utilities ? Seems generally useful
> 
> it just needs a param for the packages to search for in the job env.
> 
> Also I think we should use “package” instead of “tool” e.g in the param

This function is implemented under the assumption that the "tool" to be installed is also the name of an extra dependency in the actuator, package:

Example, for `vllm` I can install `ado-vllm-performance[vllm]` but this is not necessarily the case for the other actuators. I will extract the dependency from the actuator package installed and then intsall it separate from the actuator package.

christian-pinto:
@michael-johnston have another look please.

michael-johnston (approved):

PR #955, opened 2026-05-18
feat(core): add StandardActuator base class for simple creation of actuator plugins

### LINKED ISSUE #169: feat(actuators): improve the process of using an actuator from within another actuator

Currently, to safely re-use the logic of an actuator you have to use the following exitCode:

```python
# pseudocode
actuator_class = registry.actuatorForIdentifier(actuator_identifier)
queue = get_measurement_queue()
actuator = actuator_class.remote(queue=queue, params=actuator_params)

actuator.submit.remote(
    entities=[entity],
    experimentReference=experiment.reference,
    requesterid="run_experiment",
    requestIndex=0,
)

measurementRequest = queue.get()
# inspect measurementRequest.measurements to find the results of the measurement
```

This is a bit awkward as you have to instantiate a new MeasurementRequestQueue and Actuator instance per call to the Actuator that you would like to reuse.

Otherwise, you risk having multiple instances of the wrapped Actuator pushing their results into the same MeasurementRequestQueue. For example, that would be the case if you were to run your Actuator using the RandomWalk operator with a batchSize of 2.

**Describe the solution you'd like**

I would like to augment the Actuator API with an async function that returns the array of measurements with the arguments:

- entity
- experiment

**Additional context**

This would also simplify the code that I would write as a developer of an Actuator. For example, ActuatorBase could have a default implementation of submit() which invokes this new method.

michael-johnston:
## What Changed

Introduces `StandardActuator`, a new base class that makes writing actuator plugins significantly simpler. Plugin authors no longer need to manage Ray actors or `MeasurementQueue` instances directly.

**Two ways to use it:**
- **Simple** — override `_experiment_implementations()` and return a dict of plain Python functions; the base class handles everything else.
- **Custom** — override `_get_request_executor()` for full control when shared state or external resources are needed.

Both paths support synchronous (`execute`) and asynchronous (`submit`) execution out of the box.

**Supporting changes:**
- `NullQueue` made public — a no-op queue that allows `execute()` to work outside of an ado operation context without requiring Ray.
- Actuator registry now skips abstract classes, so `StandardActuator` itself is never registered as a concrete actuator.
- The `RoboticLab` example actuator is refactored to use the new pattern; `experiment_executor.py` removed.
- 15 new tests and updated actuator authoring documentation.


michael-johnston:
TBA: handling of ray helper actors for actuators - like cleanup and console queue that won't be available running outside ray.

orchestrator/modules/actuators/standard.py
```diff
@@ -0,0 +1,451 @@
...
+                    entity.identifier, values, request.experimentReference
+                )
+            )
+    request.measurements = results
+    request.status = compute_measurement_status(results)
+    return request
+
+
+@ray.remote
+def _run_execute_fn(
```
AlessandroPomponio, line 183: The name of this function seems a bit too generic

orchestrator/modules/actuators/standard.py
```diff
@@ -0,0 +1,451 @@
...
+        experiment = ParameterizedExperiment(
+            parameterization=request.experimentReference.parameterization,
+            **experiment.model_dump(),
+        )
+    results: list[MeasurementResult] = []
+    for entity in request.entities:
+        try:
+            input_values = experiment.propertyValuesFromEntity(entity)
+            result_dict = fn(**input_values)
+            values = dict_to_measurements(result_dict, experiment)
```
AlessandroPomponio, line 82: This method should probably be renamed to `observed_property_values_from_dicts`, since it doesn't return Measurements

orchestrator/modules/actuators/standard.py
michael-johnston: It is a pretty generic function though.

orchestrator/modules/actuators/standard.py
```diff
@@ -0,0 +1,451 @@
...
+Subclasses customise experiment logic through one of two hooks:
+
+* _experiment_implementations() — simple path: return a mapping from experiment
+  identifier to a callable ``fn(**kwargs) -> dict[str, Any]``.  The caller
+  supplies constitutive property values as kwargs; the callee returns a dict
+  whose keys are observed property identifiers.
+
+* _get_request_executor() — custom path: override to return any zero-argument
+  callable that runs the batch and returns a completed MeasurementRequest.
+  Use functools.partial or a closure to capture any actuator state required.
```
AlessandroPomponio, line 18: These seem to be two separate options. If so, it could be worth to have this be an abstract class that can't be instantiated and have two separate subclasses where each implements one option and prevents the other from being used.

It would also be good to have descriptive names for them, although I wouldn't really know what to suggest

orchestrator/modules/actuators/standard.py
```diff
@@ -0,0 +1,451 @@
...
+
+Provides both synchronous (execute) and asynchronous (submit) execution paths
+without requiring callers to manage Ray actors or MeasurementQueues.
+
+Subclasses customise experiment logic through one of two hooks:
+
+* _experiment_implementations() — simple path: return a mapping from experiment
+  identifier to a callable ``fn(**kwargs) -> dict[str, Any]``.  The caller
+  supplies constitutive property values as kwargs; the callee returns a dict
+  whose keys are observed property identifiers.
```
AlessandroPomponio, line 14: This seems sort of similar to `custom_experiments`, no?

michael-johnston: Yes it is similar.

orchestrator/modules/actuators/standard.py
michael-johnston: I tried the two class route but you end up with more code to maintain, including functions that are very similar but slightly different. 

To be clear:

_get_request_executor() default implementation-> calls _experiment_implementations. 

You can either override _experiment_implementations in a subclass and keep the default _get_request_executor (simple case) or override _get_request_executor (and optionally _experiment_implementation) if you want something more custom (where your actuator has some state/ivars it needs to access while processing a request)

As to the names this is the best I could up with
_experiment_implementations -> return python function that executes an experiment
_get_request_executor -> return a function that process a MeasurementRequest and returns a MeasurementResult

orchestrator/modules/actuators/standard.py
michael-johnston: You mean the function dict_to_measurements?

AlessandroPomponio: yep

orchestrator/modules/actuators/standard.py
AlessandroPomponio: Is there a way to merge them together in a sensible way?

michael-johnston: Maybe in future but its a much bigger change - custom_experiment uses the custom_experiment decorator. 

Basically evolution of CustomExperiments and the other actuators show common boilerplate code, this new class extracts it. 

Then in future we can evolve other actuators to be base classes of this, but doing it all one shot is too much.

AlessandroPomponio: I opened https://github.com/IBM/ado/issues/1005 from this comment

orchestrator/modules/actuators/standard.py
AlessandroPomponio: I opened https://github.com/IBM/ado/issues/1006 to track this since it's unrelated to these changes

AlessandroPomponio:
@michael-johnston there are merge conflicts that need resolving

AlessandroPomponio:
What worries me here is https://github.com/IBM/ado/pull/955/commits/f0a7fe04750c17cccb8686135f5be9d7fa39b2a6 - adding the `cloudpickle` dependency to the tests.

I see it's "Required in tests to send functions via ray (testing StandardActuator mechanisms)" and the comment in the tests says:

```
# Register this module for pickle-by-value so that test-defined functions
# (_double_x, _custom_execute, etc.) are serialised by bytecode when passed
# to Ray remote tasks.  Without this, cloudpickle would try to import
# tests.actuators.test_standard_actuator on the Ray worker, which fails
# because the tests directory is not installed.
cloudpickle.register_pickle_by_value(sys.modules[__name__])
```

Could this happen in normal usage when someone is developing a StandardActuator? In short: should we depend directly on cloudpickle?

michael-johnston:
> Could this happen in normal usage when someone is developing a StandardActuator? In short: should we depend directly on cloudpickle?

No this is just for the tests. In reality the actuator would be installed in the venv of the ray-worker so it can find the reference.

AlessandroPomponio (commented):
Please checkout the lockfile from main and re-create it

AlessandroPomponio (dismissed):

christian-pinto (approved):

PR #956, opened 2026-05-18
feat(core): add script operation support for DiscoverySpace

michael-johnston:
This PR adds first-class support for inline script operations on DiscoverySpace instances, so  custom sampling code can store measurements under a proper OperationResource without registering an operator plugin or going through orchestrate().

* Introduces ScriptOperatorConf - specifies a "dynamic" operator that doesn't require the operator registry
* Adds DiscoverySpace.operation_context() to create and lifecycle-manage a script operation (CREATED → STARTED → FINISHED/SUCCESS or FAIL), yielding an operation_id for use with Actuators
* Ensures script-run measurements appear in sampledEntities(), space summaries, and ado show operation commands
* Can be used to record any type of operation

## Use

Combined with #169 you can now do the following to iterate over a space, make measurements synchronously, store them without ray objects. 

```python
  space = DiscoverySpace.from_stored_configuration(
                      project_context=ctx,
                      space_identifier=space_id,
                      )

  actuators = {}
  for experiment in space.measurementSpace.experiments:
      aid = experiment.actuatorIdentifier
      if aid in actuators:
          continue
      actuator_cls = registry.actuatorForIdentifier(aid)
      actuators[aid] = actuator_cls()

  with space.operation_context(
      name="direct-standard-actuator-grid",
      description="Example of iterating over a discrete space",
  ) as operation_id:
      sampler = ExplicitEntitySpaceGridSampleGenerator(mode=WalkModeEnum.SEQUENTIAL)
      request_index = 0
      for batch in sampler.entityIterator(space, batchsize=1):
          for experiment in space.measurementSpace.experiments:
              ref = experiment.reference
              actuator = actuators[ref.actuatorIdentifier]
              completed = actuator.execute(
                  entities=batch,
                  experimentReference=ref,
                  requesterid=operation_id,
                  requestIndex=request_index,
                  use_ray=False,
              )
              space.addMeasurement(completed)
              request_index += 1
```

## Motivation

Developers want to use DiscoverySpace for sampling and result storage from custom scripts or lightweight inline operator logic, perhaps inside an existing operator.  Previously this required a registered explore operator in order to create the required OperationResource to associate measurements to (measurements stored without a linked OperationResource were invisible to most ADO query paths.)

### DIFF
--- a/orchestrator/core/discoveryspace/space.py
+++ b/orchestrator/core/discoveryspace/space.py
@@ -1,10 +1,11 @@
 # Copyright IBM Corporation 2025, 2026
 # SPDX-License-Identifier: MIT
 
+import contextlib
 import logging
 import os
 import typing
-from collections.abc import Callable
+from collections.abc import Callable, Iterator
 from functools import wraps
 from typing import Any
 
@@ -884,6 +885,98 @@ class DiscoverySpace:
 
         self.log.info(f"Updating run {operationResource.identifier}")
         return self._metadataStore.updateResource(operationResource)
+
+    @contextlib.contextmanager
+    def operation_context(
+        self,
+        name: str,
+        description: str | None = None,
+        metadata: dict | None = None,
+    ) -> Iterator[str]:
+        """Context manager that registers a script operation and manages its lifecycle.
+
+        Creates an OperationResource linked to this space, appends STARTED before
+        yielding the operation_id, and writes FINISHED/SUCCESS or FINISHED/FAIL on exit.
+        The operation_id should be passed as ``requesterid`` to Actuators execute
+        or submit methods
+
+        Args:
+            name: Human-readable script name stored in the operation configuration.
+            description: Optional description for the operation metadata.
+            metadata: Optional extra metadata fields merged into ConfigurationMetadata.
+
+        Yields:
+            The operation resource identifier.
+
+        Raises:
+            RuntimeError: If the discovery space has no metadata store.
+        """
+        if self._metadataStore is None:
+            raise RuntimeError(
+                "DiscoverySpace.operation_context requires a metadata store; "
+                "load the space from stored configuration first."
+            )
+
+        from orchestrator.core.metadata import ConfigurationMetadata
+        from orchestrator.core.operation.config import (
+            DiscoveryOperationConfiguration,
+            DiscoveryOperationEnum,
+            DiscoveryOperationResourceConfiguration,
+            ScriptOperatorConf,
+        )
+        from orchestrator.core.operation.resource import (
+            OperationExitStateEnum,
+            OperationResource,
+            OperationResourceEventEnum,
+            OperationResourceStatus,
+        )
+
+        script_module = ScriptOperatorConf(name=name)
+        config_metadata = ConfigurationMetadata(name=name, description=description)
+        if metadata:
+            for key, value in metadata.items():
+                setattr(config_metadata, key, value)
+
+        operation_payload = DiscoveryOperationResourceConfiguration(
+            operation=DiscoveryOperationConfiguration(
+                module=script_module,
+                parameters={},
+            ),
+            metadata=config_metadata,
+            spaces=[self.uri],
+        )
+
+        operation = OperationResource(
+            operationType=DiscoveryOperationEnum.SCRIPT,
+            operatorIdentifier=script_module.operatorIdentifier,
+            config=operation_payload,
+        )
+
+        self.addOperation(operation)
+        self._verified_operation_ids.add(operation.identifier)
+
+        try:
+            operation.status.append(
+                OperationResourceStatus(event=OperationResourceEventEnum.STARTED)
+            )
+            self.updateOperation(operation)
+            yield operation.identifier
+            operation.status.append(
+                OperationResourceStatus(
+                    event=OperationResourceEventEnum.FINISHED,
+                    exit_state=OperationExitStateEnum.SUCCESS,
+                )
+            )
+        except Exception:
+            operation.status.append(
+                OperationResourceStatus(
+                    event=OperationResourceEventEnum.FINISHED,
+                    exit_state=OperationExitStateEnum.FAIL,
+                )
+            )
+            raise
+        finally:
+            self.updateOperation(operation)
 
     @_perform_preflight_checks_for_sample_store_methods
     def complete_measurement_request_with_results_timeseries(
--- a/orchestrator/core/discoveryspace/space.py
+++ b/orchestrator/core/discoveryspace/space.py
@@ -22,6 +22,7 @@ from orchestrator.core.discoveryspace.config import (
     DiscoverySpaceConfiguration,
     DiscoverySpaceProperties,
 )
+from orchestrator.core.operation.config import DiscoveryOperationEnum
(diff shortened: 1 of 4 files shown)


orchestrator/core/discoveryspace/space.py
```diff
@@ -885,6 +890,109 @@ def updateOperation(
...
+
+        Creates an OperationResource linked to this space, appends STARTED before
+        yielding the operation_id, and writes FINISHED/SUCCESS or FINISHED/FAIL on exit.
+        The operation_id should be passed as ``requesterid`` to Actuators execute
+        or submit methods
+
+        Args:
+            name: Human-readable script name stored in the operation configuration.
+            description: Optional description for the operation metadata.
+            metadata: Optional extra metadata fields merged into ConfigurationMetadata.
```
AlessandroPomponio, line 911: Wouldn't it be clearer to have a ConfigurationMetadata parameter instead of these three? When I read name I assumed it would've been basically the operation identifier, which felt weird to me but that's what I assumed

tests/core/test_space.py
```diff
@@ -319,3 +319,75 @@ def test_matching_entities_table_virtual_property_with_multiple_values(
...
+    mock_store = MagicMock()
+    pfas_space._metadataStore = mock_store
+
+    with pfas_space.operation_context(
+        name="test-script",
+        description="Script operation for testing",
+        metadata={"labels": {"source": "test"}},
+    ) as operation_id:
+        assert operation_id.startswith("operation-script-test-script-")
+        assert operation_id in pfas_space._verified_operation_ids
```
AlessandroPomponio: Why are these things mocked?

tests/core/test_space.py
```diff
@@ -319,3 +319,75 @@ def test_matching_entities_table_virtual_property_with_multiple_values(
...
+
+    with (
+        pytest.raises(RuntimeError, match="boom"),
+        pfas_space.operation_context(name="failing-script"),
+    ):
+        raise RuntimeError("boom")
+
+    finished_operation = mock_store.updateResource.call_args_list[-1].args[0]
+    assert finished_operation.status[-1].event == OperationResourceEventEnum.FINISHED
+    assert finished_operation.status[-1].exit_state == OperationExitStateEnum.FAIL
```
AlessandroPomponio: I assume this serves to test a case where there's an exception after the operation id is returned. First of all, it looks a bit weird with the "boom", so I would maybe change that to be something like "exception during execution"
We also don't need to mock the store, right?

orchestrator/core/discoveryspace/space.py
michael-johnston: 🤷 no real opinion :-)

tests/core/test_space.py
michael-johnston: Agents

tests/core/test_space.py
michael-johnston: sure

DRL-NextGen:
**Checks Summary**

[Last run](https://cloud.ibm.com/devops/pipelines/tekton/b8283408-b2a3-44b7-adfe-aae8183c84bd/runs/93723cfc-518b-4f82-a303-06a1aa3dd05a?env_id=ibm:yp:us-south): 2026-05-29T09:01:22.823Z

Mend Unified Agent vulnerability scan found 2 vulnerabilities:
Severity | Identifier | Package | Details | Fix |
--- | --- | --- | --- | --- |
🔷 Medium | [CVE-2026-48710](https://www.mend.io/vulnerability-database/CVE-2026-48710) | starlette-0.52.1-py3-none-any.whl | <details><summary>Starlette is a lightweight ASGI framework/toolkit. Prior to version 1.0.1, the HTTP "Host" request h...</summary>Starlette is a lightweight ASGI framework/toolkit. Prior to version 1.0.1, the HTTP "Host" request header was not validated before being used to reconstruct "request.url". Because the routing algorithm relies on the raw HTTP path while "request.url" is rebuilt from the "Host" header, a malformed header could make "request.url.path" differ from the path that was actually requested. Middleware and endpoints that apply security restrictions based on "request.url" (rather than the raw "scope" path) could therefore be bypassed. Users should upgrade to a version greater than or equal to version 1.0.1, which validates the "Host" header against the grammar of RFC 9112 §3.2 / RFC 3986 §3.2.2 when constructing "request.url" and falls back to "scope["server"]" for malformed values.</details> | Upgrade to version starlette - 1.0.1,starlette - 1.0.1,https://github.com/Kludex/starlette.git - 1.0.1 |
🔷 Medium | [CVE-2026-45409](https://www.mend.io/vulnerability-database/CVE-2026-45409) | idna-3.14-py3-none-any.whl | <details><summary>This is the same issue as CVE-2024-3651, however the original remediation in 2024 was not a complete...</summary>This is the same issue as CVE-2024-3651, however the original remediation in 2024 was not a complete fix. Payloads such as ""\u0660" * N" or ""\u30fb" * N + "\u6f22"" utilize the "valid_contexto" function prior to length rejection, and for high values of "N" will take a long time to process. Impact A specially crafted argument to the "idna.encode()" function could consume significant resources. This may lead to a denial-of-service. Patches Starting in version 3.14, the function rejects long inputs as soon as practicable prior to any further processing to minimize resource consumption. In version 3.15, this approach was extended to lesser used alternate functions (i.e. per-label conversions and codec support). Workarounds Domain names cannot exceed 253 characters in length, if this length limit is enforced prior to passing the domain to the "idna.encode()" function it should no longer consume significant resources. This is triggered by arbitrarily large inputs that would not occur in normal usage, but may be passed to the library assuming there is no preliminary input validation by the higher-level application.</details> | Upgrade to version idna - 3.15 |

orchestrator/core/discoveryspace/space.py
michael-johnston: The problem on consideration is that name is the required operator name. Everything in ConfigurationMetadata is "operation" focused. 

Anyway I'm making a change

orchestrator/core/operation/config.py
```diff
@@ -351,6 +352,28 @@ def operatorIdentifier(self) -> str:
         return operator.operatorIdentifier if operator else f"{self.operatorName}-None"
 
 
+class ScriptOperatorConf(pydantic.BaseModel):
+    """Identifies an inline script or custom operator not registered in any collection."""
+
+    model_config = ConfigDict(extra="forbid")
+    name: Annotated[str, pydantic.Field(description="Human-readable script name")]
+    version: Annotated[str, pydantic.Field()] = "0.1.0"
```
AlessandroPomponio, line 360: We should have a validator for this in `orchestrator.utilities.pydantic`, but we can do it down the line. I tracked this in https://github.com/IBM/ado/issues/1004

orchestrator/core/operation/config.py
```diff
@@ -351,6 +352,28 @@ def operatorIdentifier(self) -> str:
...
+    operationType: Annotated[
+        DiscoveryOperationEnum,
+        pydantic.Field(
+            description=(
+                "Semantic operation type (e.g. search, characterize). "
+                "Script provenance is recorded separately via operation metadata labels."
```
AlessandroPomponio, line 366: This should probably change with https://github.com/IBM/ado/pull/1002

AlessandroPomponio (approved):
LGTM thanks

PR #977, opened 2026-05-26
fix(core): improve operation shutdown handling when SIGTERM is received

michael-johnston:
- Run cleanup callbacks in reverse (LIFO) order so the most recently registered callback runs first.
- Fix using local copy of signal variable `shutdown_signal_recieved` (due to using from X import Y)
- Ensure SIGTERM reliably records a FINISHED status on operations, including 
      - when shutdown happens without an exception in _run_operation_harness
      - when shutdown happens and the base callback fails

### DIFF
--- a/orchestrator/modules/operators/_orchestrate_core.py
+++ b/orchestrator/modules/operators/_orchestrate_core.py
@@ -22,7 +22,7 @@ from orchestrator.core.operation.resource import (
     OperationResourceEventEnum,
     OperationResourceStatus,
 )
-from orchestrator.modules.operators._cleanup import shutdown_signal_received
+from orchestrator.modules.operators import _cleanup
 from orchestrator.modules.operators.base import (
     InterruptedOperationError,
     add_operation_output_to_metastore,
@@ -31,6 +31,32 @@ from orchestrator.modules.operators.base import (
 
 # Global variable to track if graceful shutdown was called
 moduleLog = logging.getLogger("orchestrate_core")
+
+_SIGTERM_SHUTDOWN_MESSAGE = (
+    "An external event e.g. SIGTERM, initiated shutdown. "
+    "This may have caused the operation to exit early"
+)
+
+
+def _sigterm_finished_status(
+    *, underlying_error: BaseException | None = None
+) -> OperationResourceStatus:
+    """Return a FINISHED/error status for SIGTERM-initiated shutdown.
+
+    Args:
+        underlying_error: Optional underlying exception to append to the message.
+
+    Returns:
+        OperationResourceStatus with the SIGTERM shutdown message.
+    """
+    message = _SIGTERM_SHUTDOWN_MESSAGE
+    if underlying_error is not None:
+        message = f"{message}. Underlying Ray error: {underlying_error}"
+    return OperationResourceStatus(
+        event=OperationResourceEventEnum.FINISHED,
+        exit_state=OperationExitStateEnum.ERROR,
+        message=message,
+    )
 
 
 def log_space_details(discovery_space: "DiscoverySpace") -> None:
@@ -109,6 +135,25 @@ def _run_operation_harness(
         exit_state=OperationExitStateEnum.ERROR,
         message="Operation exited due to uncaught exception)",
     )
+    sigterm_status_callback_key = f"{operation_resource.identifier}_sigterm_status"
+    sigterm_status_was_recorded = False
+
+    # This updates operation status on SIGTERM
+    # in cases where the finally: block is not executed
+    def record_sigterm_shutdown_status() -> None:
+        nonlocal sigterm_status_was_recorded
+        sigterm_status = _sigterm_finished_status()
+        operation_resource.status.append(sigterm_status)
+        discovery_space.metadataStore.updateResource(operation_resource)
+        sigterm_status_was_recorded = True
+        moduleLog.debug(
+            f"Recorded SIGTERM shutdown status for {operation_resource.identifier}"
+        )
+
+    _cleanup.cleanup_callback_functions[sigterm_status_callback_key] = (
+        record_sigterm_shutdown_status
+    )
+
     try:
         operation_resource.status.append(
             OperationResourceStatus(event=OperationResourceEventEnum.STARTED)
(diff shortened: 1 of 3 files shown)


orchestrator/modules/operators/_orchestrate_core.py
```diff
@@ -32,6 +32,32 @@
 # Global variable to track if graceful shutdown was called
 moduleLog = logging.getLogger("orchestrate_core")
 
+_SIGTERM_SHUTDOWN_MESSAGE = (
+    "An external event e.g. SIGTERM, initiated shutdown. "
+    "This may have caused the operation to exit early"
+)
```
AlessandroPomponio: This is used only in one place, not sure if it's worth to have on its own as a variable

orchestrator/modules/operators/_orchestrate_core.py
```diff
@@ -32,6 +32,32 @@
...
+
+
+def _sigterm_finished_status(
```
AlessandroPomponio: Maybe a better name would be `_operation_status_for_sigterm_initiated_shutdown`

orchestrator/modules/operators/_orchestrate_core.py
```diff
@@ -228,13 +278,16 @@ def _run_operation_harness(
                 operation=operation_resource, exitStatus=operationStatus
             )
 
-        # Add the final status to the operation resource
-        moduleLog.info(
-            f"Sending final status for operation {operation_identifier} to metastore"
-        )
-        operation_resource.status.append(operation_output.exitStatus)
+        # If no signal OR sigterm status was not recorded update status
+        # The "not sigterm_status_was_recorded" engages the fallback
```
AlessandroPomponio, line 282: I don't understand this

orchestrator/modules/operators/_orchestrate_core.py
```diff
@@ -109,6 +135,25 @@ def _run_operation_harness(
...
+        sigterm_status_was_recorded = True
+        moduleLog.debug(
+            f"Recorded SIGTERM shutdown status for {operation_resource.identifier}"
+        )
+
+    _cleanup.cleanup_callback_functions[sigterm_status_callback_key] = (
+        record_sigterm_shutdown_status
+    )
+
     try:
```
AlessandroPomponio, line 155: We should maybe revisit all these if/else statements in the except section, it's really tough to follow

orchestrator/modules/operators/_orchestrate_core.py
michael-johnston: Originally it was in three places but then those places were further unified by the below helper. 

I can remove it.

orchestrator/modules/operators/_orchestrate_core.py
michael-johnston: It is but its also very complicated set of things that can happen that all need to be handled in different ways. 

To cover all the possibilities I couldn't find an easier way even with agent - and I tried.

orchestrator/modules/operators/_orchestrate_core.py
michael-johnston: If we got a signal and the status was recorded for the signal we don't enter this block - status is already recorded

If we didn't a signal we enter this block to record status in db. 

If we got a single but sigterm_status_was_recorded is False we enter this block - this is the fallback for a failure in the callback defined above, and why there are prior if/elses creating "fallback" error messages. 

So the flow is through the try/expect:
- Catch exception
- Check if there was a signal
      - If Yes, record a "fallback" status for the exception to be used later if we find the callback didn't store a message already.
- In finally: 
   - check if signal and if sigterm_status_was_recorded is False -> The callback failed. 
        - If so use the fallback status message

orchestrator/modules/operators/_orchestrate_core.py
AlessandroPomponio: Let's remove it, then

orchestrator/modules/operators/_orchestrate_core.py
AlessandroPomponio: Bummer... maybe in the future

DRL-NextGen:
**Checks Summary**

[Last run](https://cloud.ibm.com/devops/pipelines/tekton/b8283408-b2a3-44b7-adfe-aae8183c84bd/runs/c151acb7-566e-45b4-8b2f-65dfb79cb8a0?env_id=ibm:yp:us-south): 2026-05-29T08:36:25.270Z

Mend Unified Agent vulnerability scan found 2 vulnerabilities:
Severity | Identifier | Package | Details | Fix |
--- | --- | --- | --- | --- |
🔷 Medium | [CVE-2026-48710](https://www.mend.io/vulnerability-database/CVE-2026-48710) | starlette-0.52.1-py3-none-any.whl | <details><summary>Starlette is a lightweight ASGI framework/toolkit. Prior to version 1.0.1, the HTTP "Host" request h...</summary>Starlette is a lightweight ASGI framework/toolkit. Prior to version 1.0.1, the HTTP "Host" request header was not validated before being used to reconstruct "request.url". Because the routing algorithm relies on the raw HTTP path while "request.url" is rebuilt from the "Host" header, a malformed header could make "request.url.path" differ from the path that was actually requested. Middleware and endpoints that apply security restrictions based on "request.url" (rather than the raw "scope" path) could therefore be bypassed. Users should upgrade to a version greater than or equal to version 1.0.1, which validates the "Host" header against the grammar of RFC 9112 §3.2 / RFC 3986 §3.2.2 when constructing "request.url" and falls back to "scope["server"]" for malformed values.</details> | Upgrade to version starlette - 1.0.1,starlette - 1.0.1,https://github.com/Kludex/starlette.git - 1.0.1 |
🔷 Medium | [CVE-2026-45409](https://www.mend.io/vulnerability-database/CVE-2026-45409) | idna-3.14-py3-none-any.whl | <details><summary>This is the same issue as CVE-2024-3651, however the original remediation in 2024 was not a complete...</summary>This is the same issue as CVE-2024-3651, however the original remediation in 2024 was not a complete fix. Payloads such as ""\u0660" * N" or ""\u30fb" * N + "\u6f22"" utilize the "valid_contexto" function prior to length rejection, and for high values of "N" will take a long time to process. Impact A specially crafted argument to the "idna.encode()" function could consume significant resources. This may lead to a denial-of-service. Patches Starting in version 3.14, the function rejects long inputs as soon as practicable prior to any further processing to minimize resource consumption. In version 3.15, this approach was extended to lesser used alternate functions (i.e. per-label conversions and codec support). Workarounds Domain names cannot exceed 253 characters in length, if this length limit is enforced prior to passing the domain to the "idna.encode()" function it should no longer consume significant resources. This is triggered by arbitrarily large inputs that would not occur in normal usage, but may be passed to the library assuming there is no preliminary input validation by the higher-level application.</details> | Upgrade to version idna - 3.15 |

AlessandroPomponio (approved):
LGTM thanks

PR #998, opened 2026-06-03
feat(core): detect Ray measurement tasks that never start, and add remote runtime setup options

michael-johnston:
In ado distributed Ray operations hang when ray experiment workers do not post to the update queue **or** if posts to the queue are "lost". This can happen:

1. when measurement tasks are submitted but never actually start on a worker
2. when an unhandled exception is raised from a ray experiment worker
3. when an experiment worker is killed/ or a worker node is lost
4. There are network issues/gcs store issues that mean the sent request object is "lost"  

This PR adds a generic `ExperimentExecutorSupervisor` which monitors ray tasks that execute experiments (via their ray state), and can handle many of these issues. The CustomExperiments actuator is updated to use this. Other actuators can optionally avail of it

The PR also exposes some ray runtime env options that can impact if these errors occur

## Executor supervisor

- Watches if submitted Ray tasks are marked as FAILED by Ray
  - For these it waits taskFailedGracePeriod seconds (default 10mins) in-order for the refs from the failed task to become available (if any). 
     - If the ref is then available it tries to retrieve it (function exited, uncaught exception) 
                 - Success: it does nothing as it means the FAILED was after task exit
                 - Exception: Uncaught exception. It send InvalidMeasurementResult
     - If the ref is not available (task never started, ref got lost to do some ray error) it cancels task and sends InvalidMeasurementResult 
- Watches if submitted Ray tasks are not in a pending/running/fail state within a configurable time limit `taskRunningTimeoutSeconds` - default is 15mins
      - Once a task is seen running this timeout no longer is applied. Subsequent issues are handled by FAILED path
- (OPTIONAL) Watches if submitted RayRay tasks are still pending after a configurable time-limit taskPendingResourceTimeoutSeconds 
      - Turn this on if an autoscaling cluster is expected to handle the load

###  Edge Cases Handled

- It's possible the executor could send the result but then be marked by ray as FAILED AND the ref never become available e.g. node loss,  network issues. 
      - In this case without additional logic it would be possible for two results to be recorded for the entity 
      - To prevent this the supervisor provides a call-back - tasks can notify it that they have sent the result, so it does not put a duplicate if it later sees the task as FAILED and can't retrieve its ref
      
### Edge Cases Not Handled

 - If in previous scenario the result sent by the task never reaches the DiscoverySpaceManager due to ray errors that FAILED it, this result is lost and the operation will hang
 - If some issue kills a task but also stops ray marking it as FAILED, this will not be handled and the operation will hang
- If there is some scenario where tasks are completing successfully but the error lies in transmitting the results to DiscoverySpaceManager (network error, GCS error?) this is not handled. 

### Summary of Ray States Handling 

Task are considered dead after taskRunningTimeoutSeconds is exceeded for these states

State at timeout | Verdict | Reasoning
-- | -- | --
NIL | Definitively stuck | Task has not been accepted by the scheduler at all after 900s. Nothing normal keeps a task in NIL this long.
PENDING_ARGS_AVAIL | Likely stuck | Input object refs are not available. For an executor task submitted with live refs this should be milliseconds. Persisting to timeout implies upstream has failed silently.
PENDING_ARGS_FETCH / GETTING_AND_PINNING_ARGS | Ambiguous | Brief transfer states. Should be sub-second under normal conditions; persisting to 900s suggests a network or object-store issue. Marginal but still a bad sign.
SUBMITTED_TO_WORKER | Ambiguous but suspicious | The task has been accepted by a worker process but not yet started. Normally sub-second. At 900s this implies worker process is hung/overloaded, but the task has been picked up.
OTHER / lookup failure | Unknown | Cannot determine state; fire the running-timeout anyway as a safety net.

if taskPendingResourceTimeoutSeconds is set, tasks are considered dead if they are still in either of these states at the timeout. This is for use with autoscaling cluster, which is expected to scale to the required load. 

State at timeout | Verdict | Reasoning
-- | -- | --
PENDING_NODE_ASSIGNMENT | Likely stuck | No node with matching resources has been assigned. The autoscaler has had taskPendingResourceTimeoutSeconds to provision one. Likely a a permanent resource mismatch.
PENDING_OBJ_STORE_MEM_AVAIL | Likely stuck | Plasma memory pressure. Might eventually resolve if other tasks free memory, but state persisting for taskPendingResourceTimeoutSeconds is considered an issue

Tasks considered dead when taskFailedGracePeriod is exceeded

State at timeout | Verdict | Reasoning
-- | -- | --
FAILED |  Stuck if also no ref after timeout AND task did not inform of completion | If ray has marked as FAILED but no ref has appeared after 10mins AND the task did not inform it has sent results we can assume no result will be sent and we need to act

## Remote execution

- Adds optional runtimeEnv settings to remote execution context files:
    - `setupTimeoutSeconds` — how long Ray may spend setting up the job environment
    - `eagerInstall `— whether to install the environment when the job starts or wait until the first task runs
- Documents these options in the remote run guide.

## Tests

- Unit tests for helper logic and configuration mapping.
- Integration tests against a live Ray cluster for timeout, running-task, completion, and duplicate-result behaviour.


AlessandroPomponio:
@michael-johnston tests are failing

michael-johnston:
Everything passing now @AlessandroPomponio

orchestrator/core/remotecontext/config.py
```diff
@@ -98,6 +98,54 @@ class JobExecutionType(pydantic.BaseModel):
...
+    """
+
+    model_config = pydantic.ConfigDict(extra="forbid")
+
+    setupTimeoutSeconds: Annotated[
+        int,
+        pydantic.Field(
+            description=(
+                "Maximum seconds to create the job runtime environment on a worker. "
+                "Use -1 to disable the timeout."
```
AlessandroPomponio, line 122: Does any negative number work or is it required to be -1? If the latter we should ensure we add `ge=-1`

orchestrator/core/remotecontext/config.py
```diff
@@ -98,6 +98,54 @@ class JobExecutionType(pydantic.BaseModel):
 ]
 
 
+# Ray RuntimeEnvConfig defaults (see Ray handling-dependencies docs).
+RAY_DEFAULT_SETUP_TIMEOUT_SECONDS = 600
```
AlessandroPomponio: Might be worth to record that this isn't in the docs (or I couldn't find it). There's the `DEFAULT_RUNTIME_ENV_TIMEOUT_SECONDS` variable in the source code in `ray._private.ray_constants`

orchestrator/core/remotecontext/config.py
```diff
@@ -98,6 +98,54 @@ class JobExecutionType(pydantic.BaseModel):
...
+RAY_DEFAULT_EAGER_INSTALL = True
```
AlessandroPomponio, line 104: Is this also taken from Ray or is it from us?

orchestrator/core/remotecontext/config.py
```diff
@@ -98,6 +98,54 @@ class JobExecutionType(pydantic.BaseModel):
...
+    @classmethod
+    def validate_setup_timeout_seconds(cls, value: int) -> int:
+        """Validate setup timeout matches Ray rules."""
+        if value == -1:
+            return value
+        if value <= 0:
+            raise ValueError(
+                "setupTimeoutSeconds must be greater than zero or -1 to disable timeout"
+            )
+        return value
```
AlessandroPomponio, line 147: Well, this answers my question. We can get rid of it and just add `gt=-1` in the field since we're using the annotated pattern: https://pydantic.dev/docs/validation/latest/concepts/types/#using-the-annotated-pattern

orchestrator/core/remotecontext/config.py
```diff
@@ -170,6 +218,17 @@ class RemoteExecutionContext(pydantic.BaseModel):
         ),
     ]
 
+    runtimeEnv: Annotated[
+        RuntimeEnvironmentConfiguration | None,
+        pydantic.Field(
+            default=None,
```
AlessandroPomponio: Not needed when using the annotated pattern

```suggestion
```

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+def _verify_supervisor_ray_states_supported() -> None:
+    """Fail fast if Ray's State API no longer exposes RUNNING or FAILED."""
+    api_states = _ray_api_task_state_names()
+    missing = _SUPERVISOR_RAY_STATE_NAMES - api_states
+    if missing:
+        raise RuntimeError(
+            "Ray State API task states no longer include "
+            f"{sorted(missing)} (required by LaunchSupervisor). "
+            f"Available states: {sorted(api_states)}. "
+            "Update orchestrator.modules.actuators.measurement_launch."
```
AlessandroPomponio: This path isn't correct

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+
+
+def _verify_supervisor_ray_states_supported() -> None:
+    """Fail fast if Ray's State API no longer exposes RUNNING or FAILED."""
+    api_states = _ray_api_task_state_names()
+    missing = _SUPERVISOR_RAY_STATE_NAMES - api_states
+    if missing:
+        raise RuntimeError(
+            "Ray State API task states no longer include "
+            f"{sorted(missing)} (required by LaunchSupervisor). "
```
AlessandroPomponio: I assume LaunchSupervisor is a name that was used previously

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+def _ray_api_task_state_names() -> frozenset[str]:
+    """Return task state strings declared on Ray's ``TaskState`` schema."""
+    from ray.util.state.common import TaskState as RayTaskStateRecord
+
+    annotation = RayTaskStateRecord.__annotations__["state"]
+    return frozenset(typing.get_args(annotation))
+
+
+def _verify_supervisor_ray_states_supported() -> None:
+    """Fail fast if Ray's State API no longer exposes RUNNING or FAILED."""
```
AlessandroPomponio, line 50: Will this make it impossible to run anything or will it just disable these features?

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+            or ``OTHER`` (all other states and unavailable lookups).
+        """
+        if raw in (cls.RUNNING.value, "RUNNING_IN_RAY_GET", "RUNNING_IN_RAY_WAIT"):
+            return cls.RUNNING
+        if raw == cls.FAILED.value:
+            return cls.FAILED
+        if raw == cls.PENDING_NODE_ASSIGNMENT.value:
+            return cls.PENDING_NODE_ASSIGNMENT
+        if raw == cls.PENDING_OBJ_STORE_MEM_AVAIL.value:
+            return cls.PENDING_OBJ_STORE_MEM_AVAIL
```
AlessandroPomponio: This is probably cleaner with a `match`/`case`

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+        log = logger or logging.getLogger(__name__)
+        if raw is None:
+            log.debug("Ray task state unavailable; treating as %s", cls.OTHER.value)
+        else:
+            log.debug(
+                "Ray task state %r collapsed to %s for launch supervision",
+                raw,
+                cls.OTHER.value,
+            )
+        return cls.OTHER
```
AlessandroPomponio: Actually, this whole thing could be a match/case, where `None` is the first match, then you have all the states, and a default that maps to other

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+        return ExperimentExecutorSupervisorConfig(
+            taskFailedGraceSeconds=self.taskFailedGraceSeconds,
+            taskRunningTimeoutSeconds=self.taskRunningTimeoutSeconds,
+            supervisorPollIntervalSeconds=self.supervisorPollIntervalSeconds,
+            taskPendingResourceTimeoutSeconds=self.taskPendingResourceTimeoutSeconds,
+        )
+
+
+@dataclass
+class _MonitoredExecutor:
```
AlessandroPomponio: Wouldn't `_SupervisedRayTask` be a more fitting name?

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+    submitted_at: float
+    seen_running: bool = False
+    """True once Ray State API has reported RUNNING for this task."""
+
+
+@dataclass
+class _SupervisorState:
+    """Mutable supervisor state guarded by a lock."""
+
+    monitored_executors: dict[str, _MonitoredExecutor] = field(default_factory=dict)
```
AlessandroPomponio: I'd call this `supervised_tasks`

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+    try:
+        task_id = executor_ref.task_id().hex()
+    except (AttributeError, RuntimeError, ValueError):
+        return RayTaskState.OTHER
+
+    from ray.util.state import list_tasks
+
+    last_error: Exception | None = None
+    tasks: list[object] = []
+    max_attempts = 8
```
AlessandroPomponio, line 249: Wouldn't it be better if this was configurable (keeping 8 as default)?

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+    if isinstance(raw_state, str):
+        return RayTaskState.from_ray_state(raw_state)
+    return RayTaskState.OTHER
+
+
+def notify_executor_supervisor_completed(
+    notifier: object | None,
+    requestid: str,
+) -> None:
+    """Notify executor supervisor that an executor queue a measurement result.
```
AlessandroPomponio: Something's missing here

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+
+    ``notifier`` is usually the hosting actuator (in-process or as a Ray actor
+    handle). It must expose ``mark_launch_completed``;
```
AlessandroPomponio: Since "It must expose ``mark_launch_completed``" this should be a `Protocol` object rather than a generic `object` to help type IDEs

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+        InvalidMeasurementResult(
+            entityIdentifier=entity.identifier,
+            experimentReference=request.experimentReference,
+            reason=reason,
+        )
+        for entity in request.entities
+    ]
+    request.measurements = measurements
+    request.status = compute_measurement_status(measurements)
+    return request
```
AlessandroPomponio, line 296: Should this method be here or is it generic/reusable enough?

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+    ) -> None:
+        """Register an executor task for supervision."""
+        with self._lock:
+            if request.requestid in self._state.completed_request_ids:
+                return
+            self._state.monitored_executors[request.requestid] = _MonitoredExecutor(
+                request=request,
+                executor_ref=executor_ref,
+                submitted_at=time.monotonic(),
+            )
```
AlessandroPomponio: I'd call this `monitor_ray_task` or `monitor_task` if you want to have more emphasis on the ado request side of things

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+        an external problem causes the task to FAIL with no associated exception
+        e.g. raylet failure, node failure, some issue with ray ref retrieval.
+
+        Called from the executor path via :func:`notify_executor_supervisor_completed`
+        as soon as the measurement queue receives a result, before the Ray executor
+        ref becomes ready.
+        """
+        with self._lock:
+            self._state.completed_request_ids.add(requestid)
+            self._state.monitored_executors.pop(requestid, None)
```
AlessandroPomponio: I'd call this `mark_task_complete` or something along the lines. In general, I'd mention `task` everywhere in these methods to make sure people understand they are related to the tasks and not to the supervisor itself

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+
+    def _run_loop(self) -> None:
+        """Poll pending executor tasks until stopped."""
+        while not self._stop.is_set():
+            self._poll_once()
+            time.sleep(self._config.supervisorPollIntervalSeconds)
```
AlessandroPomponio: I'd call this `_task_monitoring_loop`

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+
+    def _poll_once(self) -> None:
+        """Run a single supervision pass over pending launches."""
+        with self._lock:
+            pending_snapshot = list(self._state.monitored_executors.values())
+
+        for pending in pending_snapshot:
+            self._check_pending(pending)
```
AlessandroPomponio: This is only used in the function above, right? If that's the case, do we need it?

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+            ray.get(pending.executor_ref)
+        except Exception as error:
+            self._record_executor_failure(
+                pending,
+                reason=f"Executor task raised: {error}",
+            )
+        else:
+            with self._lock:
+                self._state.completed_request_ids.add(pending.request.requestid)
+                self._state.monitored_executors.pop(pending.request.requestid, None)
```
AlessandroPomponio: These tasks seem to have completed rather than be "ready".

I would:

- Call this `_handle_task_completion` 
- Rename `pending` (the parameter) to `task`
- Replace the "else" part should be a call to `mark_task_complete`

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+                return
+            self._state.completed_request_ids.add(requestid)
+
+        failed_request = add_invalid_measurement_results(
+            pending.request.model_copy(deep=True),
+            reason=reason,
+        )
+        self._queue.put(failed_request, block=False)
+        self._log.warning(
+            "Launch supervision failure for request %s (index=%s): %s",
```
AlessandroPomponio: I think launch supervision is still old terminology

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+            return
+
+        requestid = pending.request.requestid
+        with self._lock:
+            if requestid in self._state.completed_request_ids:
+                self._state.monitored_executors.pop(requestid, None)
+                return
+
+        self._record_executor_failure(pending, reason=reason)
+        self._cancel_executor(pending.executor_ref)
```
AlessandroPomponio: This method is a bit confusing because it's called `_emit_launch_failure` but it also performs a check with the ray tasks to see if the task actually completed before it runs the failure handling.

The method is always called after a timeout check is performed - should the result be valid in the (i assume) off chance it becomes available in the delta time between the checks? Should this be supported? Should we have a different name for this method? 

I think the method above, `_record_executor_failure`, is actually the one that's doing what this method would suggest through its name

orchestrator/modules/actuators/executor_supervisor.py
```diff
@@ -0,0 +1,549 @@
...
+
+    def _cancel_executor(self, executor_ref: ray.ObjectRef) -> None:
+        """Best-effort cancellation of a stuck executor task."""
+        try:
+            ray.cancel(executor_ref, force=True, recursive=True)
+        except (TypeError, ValueError, RuntimeError) as error:
+            self._log.debug("Could not cancel executor ref: %s", error)
```
AlessandroPomponio: To unify naming and help clarify the best-effort nature of the method, I'd call this `_request_task_cancellation`

AlessandroPomponio (dismissed):
We need to think very carefully about the naming we're introducing in this PR. As it stands, a lot of it is confusing for me. I have some suggestions, but we should also have a pass to make sure it's uniform.

orchestrator/core/remotecontext/config.py
michael-johnston: These are just our internal module level vars for holding these values.

orchestrator/core/remotecontext/config.py
michael-johnston: As above - just for setting the default values clearly,

orchestrator/core/remotecontext/config.py
michael-johnston: -1

orchestrator/modules/actuators/executor_supervisor.py
michael-johnston: If Ray state api gets rid of RUNNING or FAILED tests will fail. Code importing the supervisor will fail. 

It's very unlikely but just in case.

orchestrator/modules/actuators/executor_supervisor.py
michael-johnston:
```suggestion
            "Update orchestrator.modules.actuators.executor_supervisor."
```

orchestrator/modules/actuators/executor_supervisor.py
michael-johnston:
```suggestion
            f"{sorted(missing)} (required by ExperimentExecutorSupervisor). "
```

orchestrator/modules/actuators/executor_supervisor.py
michael-johnston: It's not a generic ray task though, it's something executing an experiment c.f. request field.

orchestrator/modules/actuators/executor_supervisor.py
michael-johnston:
```suggestion
    """Notify executor supervisor that an executor queued a measurement result.
```

michael-johnston: Does changing to  `queued` fix the problem?

orchestrator/modules/actuators/executor_supervisor.py
michael-johnston: I think it's just here.

orchestrator/modules/actuators/executor_supervisor.py
michael-johnston: ready in terms of "ray object ref is ready to be retrieved" , but ok in any case

orchestrator/modules/actuators/executor_supervisor.py
michael-johnston:
```suggestion
            "Execution supervision failure for request %s (index=%s): %s",
```

orchestrator/modules/actuators/executor_supervisor.py
michael-johnston: Can change the name. There are two failure modes, that both require InvalidMeasurementResult o be set
- ray task never launched (emit_launch_failure)
- ray task launched but then failed (record_executor_failure)

> - should the result be valid in the (i assume) off chance it becomes available in the delta time between the checks? S

The `if ready_refs` is meant to catch this potential loop whole. 
1. The task has gone passed launch timeout without become reading - give up on it -> call this method
2. Check just in case the ref became available (because the above doesn't check that only ray state)
3. If yes, its ok just go normal path, If no, fail it and emit InvalidMeasurementResult

orchestrator/modules/actuators/executor_supervisor.py
michael-johnston: In general "executor" is supposed to be shorthand to mean "ray task that will execute an experiment according to a MeasurementRequest and it fails in a particular way InvalidMeasurementResult is to be created a queued on state queue"  - hence ExperimentExecutorSupervisor. 

We could call it monitor_executor, monitor_experiment_executor - but it's not a generic ray task.

orchestrator/core/remotecontext/config.py
AlessandroPomponio: Gotcha, I'd still mention where to find them (as above) instead of saying "(see Ray handling-dependencies docs)", which didn't help me too much earlier

orchestrator/modules/actuators/executor_supervisor.py
AlessandroPomponio: Yes, I was wondering if that kind of removal would break the code completely or allow proceeding without this bit, but I'm guessing it's the first one

orchestrator/modules/actuators/executor_supervisor.py
AlessandroPomponio: Yes, I think this fixes the docstring. I would then change the method name to be something like `notify_request_completion` - something that clearly states that it's a notification of a completion event and not that the supervisor has completed

orchestrator/modules/actuators/executor_supervisor.py
AlessandroPomponio: Maybe `monitor_experiment_executor`, then?

orchestrator/modules/actuators/executor_supervisor.py
AlessandroPomponio: > 3\. If yes, its ok just go normal path, If no, fail it and emit InvalidMeasurementResult

It feels weird to me for this function to also have this concern after it's been checked already. In my opinion this is already the "give up" function. If in the milliseconds between the check and this function executing the result has become available, too bad: that's what timeouts are for anyway

orchestrator/modules/actuators/executor_supervisor.py
michael-johnston: It would break

orchestrator/core/remotecontext/config.py
michael-johnston: ge=-1 includes 0 while above disallows 0. 

In any case let's see can we allow 0.

michael-johnston: Ray doesn't allow 0 at python level 


Value | Ray behavior
-- | --
1 | Valid: 1-second timeout (impractical, but literal)
0 | Invalid in Python API; if it reaches the proto layer, silently becomes 600s default
-1 | Valid: timeout disabled


It's not clear what will happen if 0 is passed - could be error, could be 600s - depends on what parses first which I don't think there's any guarantee of. Plus in either case allowing it would be confusing.

orchestrator/modules/actuators/custom_experiments.py
```diff
@@ -615,6 +619,7 @@ def custom_experiment_executor(
     measurement_request: MeasurementRequest,
     target_experiment: Experiment,
     queue: MeasurementQueue,
+    custom_experiments_actuator: typing.Optional["CustomExperimentsActor"] = None,
```
AlessandroPomponio:
```suggestion
    custom_experiments_actor: "CustomExperimentsActor | None" = None,
```

orchestrator/modules/actuators/custom_experiments.py
```diff
@@ -623,6 +628,8 @@ def custom_experiment_executor(
     :param target_experiment: The experiment to execute.
         Required as the measurementRequest only includes an ExperimentReference
     :param queue: The queue to put the result on
+    :param custom_experiments_actuator: Optional handle to CustomExperiments actuator
+        to notify after the result is queued
```
AlessandroPomponio:
```suggestion
    :param custom_experiments_actor: Optional handle to CustomExperiments actuator
        to notify after the result is queued
```

orchestrator/modules/actuators/custom_experiments.py
```diff
@@ -662,12 +669,23 @@ def custom_experiment_executor(
     measurement_request.status = compute_measurement_status(measurement_results)
 
     queue.put(measurement_request, block=False)
+    if custom_experiments_actuator:
+        custom_experiments_actuator.mark_measurement_request_completed.remote(
+            measurement_request.requestid
+        )
```
AlessandroPomponio:
```suggestion
    if custom_experiments_actor:
        custom_experiments_actor.mark_measurement_request_completed.remote(
            measurement_request.requestid
        )
```

tests/actuators/test_executor_supervision_integration.py
AlessandroPomponio, line 1: These tests might slow down the CI by quite a lot - could we use a smaller timeout?

michael-johnston: Lets see how it goes

AlessandroPomponio (approved):
LGTM thanks

AlessandroPomponio:
~~@michael-johnston tests are using too much memory now~~
EDIT: they're unrelated CI issues

PR #1002, opened 2026-06-04
feat(core): record distribution package provenance on resource creation

michael-johnston:
When a DiscoverySpaceResource, OperationResource, or ActuatorConfigurationResource is written to the metastore, the exact PyPI distribution name and version of every plugin involved is now captured and stored alongside it. This makes it straightforward to identify — after the fact — which package versions were used to create a given resource, supporting both replication and debugging.

Previously this was somewhat deducible for some operators as long as the operator name and package name were similar as the identifier contained the operator plugin version. However this breaks down if a package provides multiple operators. For experiments or actuator configurations there was no record.

Changes
- Add model for package provenance
- Add model for collection of package provenance instances (ProvenanceInfo)
- Add field `provenance` to ADOResource - type `ProvenancenInfo` (empty by default)
- Add subclasses of ProvenanceInfo for specific resource type (DiscoverySpaceResourceProvenanceInfo)
- Add functions to plugin catalogs (actuator registry, operator collections) to get relevant data
- Before creation of resource of each type get package data and add it

All fields in ProvenanceInfo subclass are expected to have type dict[str, PackageProvenance]. 

Examples: 

space - dictionaries of `customExperiment` packages and `actuator packages`

```yaml
...
created: '2026-06-04T18:17:16.323098Z'
provenance:
  customExperiments:
    nevergrad_opt_3d_test_func:
      distributionName: optimization_test_functions
       distributionVersion: 1.8.1.dev46+437b3291.dirty
   actuators:
      ....
identifier: space-7b7ae1-default
status:
- event: created
  recorded_at: '2026-06-04T18:17:16.323104Z'
- event: added
  recorded_at: '2026-06-04T18:17:16.324510Z'
 ```
 
 operation - dict of operators. 
 
 ```yaml
 ...
 kind: operation
metadata:
  entities_submitted: 40
  experiments_requested: 40
operationType: search
operatorIdentifier: ray_tune-1.8.1.dev43+g08402ddd3.d20260603144611
provenance: 
    operators:
         ray_tune:
           distributionName: ado-ray-tune
           distributionVersion: 1.8.1.dev43+g08402ddd3.d20260603144611
  ...
 ```
 
 ## Notes
 
 We use `dict[str, PackageProveancen]` for collecting packages even in cases there can be only one (like one operator per operation) to keep the same structure across all resources.

### DIFF
--- a/orchestrator/core/discoveryspace/resource.py
+++ b/orchestrator/core/discoveryspace/resource.py
@@ -8,13 +8,38 @@ import pydantic
 import rich.box
 
 from orchestrator.core.discoveryspace.config import DiscoverySpaceConfiguration
-from orchestrator.core.metadata import PackageProvenance
+from orchestrator.core.metadata import PackageProvenance, ProvenanceInfo
 from orchestrator.core.resources import ADOResource, CoreResourceKinds
 from orchestrator.schema.measurementspace import MeasurementSpaceConfiguration
 from orchestrator.utilities.pydantic import Defaultable
 
 if typing.TYPE_CHECKING:
     from rich.console import RenderableType
+
+
+class DiscoverySpaceProvenanceInfo(ProvenanceInfo):
+    """Plugin provenance for a discovery space resource."""
+
+    actuators: Annotated[
+        dict[str, PackageProvenance],
+        pydantic.Field(
+            default_factory=dict,
+            description=(
+                "Mapping of actuator identifier to the Python distribution that "
+                "provided it at the time this space was created."
+            ),
+        ),
+    ]
+    customExperiments: Annotated[
+        dict[str, PackageProvenance],
+        pydantic.Field(
+            default_factory=dict,
+            description=(
+                "Mapping of custom experiment identifier to the Python distribution "
+                "that provided it at the time this space was created."
+            ),
+        ),
+    ]
 
 
 class DiscoverySpaceResource(ADOResource):
@@ -29,30 +54,11 @@ class DiscoverySpaceResource(ADOResource):
             default_factory=lambda: f"space-{str(uuid.uuid4())[:8]}",
         ),
     ]
-
-    actuatorProvenance: Annotated[
-        dict[str, PackageProvenance],
+    provenance: Annotated[
+        DiscoverySpaceProvenanceInfo,
         pydantic.Field(
-            default_factory=dict,
-            description=(
-                "Mapping of actuator identifier to the Python distribution that "
-                "provided it at the time this space was created. Populated automatically "
-                "from the installed environment; empty for resources created before "
-                "provenance tracking was introduced."
-            ),
-        ),
-    ]
-
-    customExperimentProvenance: Annotated[
-        dict[str, PackageProvenance],
-        pydantic.Field(
-            default_factory=dict,
-            description=(
-                "Mapping of custom experiment identifier to the Python distribution "
-                "that provided it at the time this space was created. Only populated "
-                "for experiments registered via the ``ado.custom_experiments`` entry "
-                "point whose source module can be resolved to a distribution."
-            ),
+            default_factory=DiscoverySpaceProvenanceInfo,
+            description="Plugin package provenance frozen at resource creation time.",
         ),
     ]
 
--- a/orchestrator/core/discoveryspace/resource.py
+++ b/orchestrator/core/discoveryspace/resource.py
@@ -2,11 +2,13 @@
 # SPDX-License-Identifier: MIT
 import typing
 import uuid
+from typing import Annotated
 
 import pydantic
 import rich.box
 
 from orchestrator.core.discoveryspace.config import DiscoverySpaceConfiguration
+from orchestrator.core.metadata import PackageProvenance
 from orchestrator.core.resources import ADOResource, CoreResourceKinds
 from orchestrator.schema.measurementspace import MeasurementSpaceConfiguration
 from orchestrator.utilities.pydantic import Defaultable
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Comment by AlessandroPomponio** [comment]
Wouldn't it be better to have a model called `ProvenanceInfo` which we can then include in all `AdoResource`s?

Something like:

```python
class ProvenanceInfo:
  actuators: dict[str, PackageProvenance] | None
  custom_experiments: dict[str, PackageProvenance] | None
  operators: dict[str, PackageProvenance] | None
```

This would give us a consistent pattern + extensibility for the future.

As a side note, I think we should stop using camelCasing even in Pydantic models, so I would use snake_case everywhere

michael-johnston:
> Wouldn't it be better to have a model called ProvenanceInfo which we can then include in all AdoResources?

Only certain resources have this information though.

Its seems strange to have these fields on datacontainer or samplestore - they would also be in full dump of objects and schema. 

Maybe the object should just be 
```
class ProvenanceInfo:
    #no fields
    #add validator that all fields values are dict of name:PackageProvenance
```

resource subclass can then also subclass this OperationProvenanceInfo etc and put whatever fields are relevant to them.

AlessandroPomponio:
> Its seems strange to have these fields on datacontainer or samplestore

Well, we could always add information on what ado version created the resources. This could help us down the line troubleshoot resource model upgrades. Right now we're doing some magical parsing of the Pydantic errors when it comes to legacy resource upgrades, for example.



> they would also be in full dump of objects and schema.

I think we don't dump nulls by default, right? I still think it's worth having a dedicated provenance section, though.



> Maybe the object should just be
> 
> ```
> class ProvenanceInfo:
>     #no fields
>     #add validator that all fields values are dict of name:PackageProvenance
> ```
> 
> resource subclass can then also subclass this OperationProvenanceInfo etc and put whatever fields are relevant to them.

Sure. We could also validate with `InstanceOf` https://pydantic.dev/docs/validation/latest/concepts/validators/#special-types

orchestrator/core/discoveryspace/resource.py
```diff
@@ -15,18 +17,50 @@
...
+        dict[str, PackageProvenance],
+        pydantic.Field(
+            default_factory=dict,
+            description=(
+                "Mapping of actuator identifier to the Python distribution that "
+                "provided it at the time this space was created."
+            ),
+        ),
+    ]
+    customExperiments: Annotated[
```
AlessandroPomponio, line 33: I think we should stop using camelCase even for our model fields. I would suggest renaming this to `custom_experiments`

orchestrator/core/discoveryspace/space.py
```diff
@@ -532,13 +532,76 @@ def config(self) -> DiscoverySpaceConfiguration:
...
+                                version = dist.metadata.get("Version")
+                                if version is not None:
+                                    custom_experiments[experiment.identifier] = (
+                                        PackageProvenance(
+                                            distributionName=dist_name,
+                                            distributionVersion=version,
+                                        )
+                                    )
+                        except Exception:  # noqa: S110
+                            pass
```
AlessandroPomponio: We can invert the ifs to reduce indentation

orchestrator/core/discoveryspace/space.py
```diff
@@ -532,13 +532,76 @@ def config(self) -> DiscoverySpaceConfiguration:
...
+                        if isinstance(module_conf, dict)
+                        else getattr(module_conf, "moduleName", None)
+                    )
+                    if module_name is not None:
+                        try:
+                            dist_name = distribution_from_module(module_name)
+                            if dist_name is not None:
+                                dist = importlib.metadata.distribution(dist_name)
+                                version = dist.metadata.get("Version")
+                                if version is not None:
```
AlessandroPomponio: We do something similar for the actuator plugins, maybe we can have a function that returns this info (and the summary)

orchestrator/core/metadata.py
```diff
@@ -30,3 +31,60 @@ class ConfigurationMetadata(pydantic.BaseModel):
...
+    replication or debugging.
+
+    Attributes:
+        distributionName: The PyPI distribution name (e.g. ``"ado-ray-tune"``).
+        distributionVersion: The installed version of the distribution (e.g. ``"1.7.1"``).
+    """
+
+    model_config = ConfigDict(frozen=True)
+
+    distributionName: Annotated[
```
AlessandroPomponio, line 52: distribution_name

orchestrator/core/operation/config.py
```diff
@@ -253,6 +253,16 @@ class OperatorMetadata(pydantic.BaseModel):
             description="The discovery operation type this operator belongs to."
         ),
     ]
+    distributionName: Annotated[
```
AlessandroPomponio: Why is it only the distribution name and not the provenance?

orchestrator/core/metadata.py
```diff
@@ -30,3 +31,60 @@ class ConfigurationMetadata(pydantic.BaseModel):
...
+        str,
+        pydantic.Field(
+            description="PyPI distribution name (e.g. 'ado-ray-tune', 'ado-core')."
+        ),
+    ]
+    distributionVersion: Annotated[
```
AlessandroPomponio, line 58: distribution_version

We should add `validate_version_is_pep440` as an AfterValidator (we should take that function out of where it is and put it in `orchestrator/utilities/pydantic.py`

orchestrator/core/metadata.py
```diff
@@ -30,3 +31,60 @@ class ConfigurationMetadata(pydantic.BaseModel):
...
+                    f"{field_name} must be a dict[str, PackageProvenance], "
+                    f"got {type(value)}"
+                )
+            for key, item in value.items():
+                if not isinstance(item, PackageProvenance):
+                    raise ValueError(
+                        f"{field_name}[{key!r}] must be PackageProvenance, "
+                        f"got {type(item)}"
+                    )
+        return self
```
AlessandroPomponio, line 167: Is this even worth having?

michael-johnston: Why not

orchestrator/core/discoveryspace/resource.py
michael-johnston: Let's separate naming from this PR - we shouldn't do this piecemeal.

orchestrator/core/discoveryspace/space.py
michael-johnston: Are you saying "create utility function to do X and use in Y and Z"?

AlessandroPomponio: Yep, using it in the Annotated pattern https://pydantic.dev/docs/validation/latest/concepts/validators/#field-validators

orchestrator/core/metadata.py
AlessandroPomponio: Given that this model has no fields and we control the other models, it felt a bit redundant, but sure

orchestrator/core/discoveryspace/space.py
michael-johnston: Can you be specific on the suggested changes e.g. where in  actuator plugins the similar thing happens and what the function does and example of the validator? 

Its because you have view of required change and if we're expanding this PR to include it, need to be specific as otherwise time is required to deduce what the suggestion means :-)

orchestrator/core/operation/config.py
michael-johnston: Good question - this doesn't look right.

orchestrator/core/metadata.py
michael-johnston: It's to save implementing in all sub-models, since all submodes should follow this structure. Inheritance??

orchestrator/core/discoveryspace/space.py
michael-johnston: All the distribution name/version extraction code has been centralised to PackageProvenance

AlessandroPomponio (approved):
LGTM thanks

PR #1014, opened 2026-06-08
feat(vllm_performance): Add vLLM threadpool support

mgazz:
This pull request adds measurement caching and vLLM threadpool support to the vLLM performance testing actuator. The caching mechanism reuses benchmark results for entities with identical environment and benchmark parameters, significantly reducing redundant measurements. Additionally, it implements version-aware threadpool configuration for vLLM >= 0.20.0, with automatic fallback for older versions.

Key Features

1. **Measurement Caching**
Intelligent Result Reuse: Automatically caches and reuses benchmark results for entities with identical configurations
Composite Cache Keys: Combines environment parameters (model, GPUs, vLLM settings) and benchmark parameters (request rate, token counts) into unique cache keys
Significant Performance Gains: Eliminates redundant benchmarks in parameter sweeps with repeated configurations

2. **Version-Aware Threadpool Support**
Automatic Detection: Parses vLLM version from image property to determine threadpool compatibility
vLLM >= 0.20.0: Enables threadpool rendering with configurable worker counts
Backward Compatible: Automatically disables threadpool for older vLLM versions. 

3. **Enhanced Image Format**
New List Format: ["vllm/vllm-openai:v0.20.1", "0.20.1"] enables version-aware features
Backward Compatible: String format "vllm/vllm-openai:v0.20.1" still works
Version Information: Second element provides explicit version for feature detection

New Parameters:
  - threadpool (0=disabled, 1=enabled, default: 1)
  - renderer_num_workers (1-128, default: 32)

### DIFF
--- a/orchestrator/core/discoveryspace/group_samplers.py
+++ b/orchestrator/core/discoveryspace/group_samplers.py
@@ -60,12 +60,10 @@ def _build_point_group_values(
     """
 
     def make_hashable(
-        value: float | list | str | dict | None,
+        value: float | list | str | None,
     ) -> int | float | tuple | str | None:
         """Convert unhashable types to hashable equivalents"""
         match value:
-            case dict():
-                return tuple(sorted(value.items()))
             case list():
                 return tuple(value)
             case _:
--- a/orchestrator/core/discoveryspace/group_samplers.py
+++ b/orchestrator/core/discoveryspace/group_samplers.py
@@ -63,14 +63,13 @@ def _build_point_group_values(
         value: float | list | str | dict | None,
     ) -> int | float | tuple | str | None:
         """Convert unhashable types to hashable equivalents"""
-        if isinstance(value, dict):
-            # Convert dict to sorted tuple of items
-            return tuple(sorted(value.items()))
-        if isinstance(value, list):
-            # Convert list to tuple
-            return tuple(value)
-        # Return value as-is if already hashable
-        return value  # type: ignore[return-value]
+        match value:
+            case dict():
+                return tuple(sorted(value.items()))
+            case list():
+                return tuple(value)
+            case _:
+                return value  # type: ignore[return-value]
 
     return frozenset({(k, make_hashable(v)) for k, v in point.items() if k in group})
 
--- a/orchestrator/core/discoveryspace/group_samplers.py
+++ b/orchestrator/core/discoveryspace/group_samplers.py
@@ -55,9 +55,24 @@ def _build_point_group_values(
 ) -> frozenset[tuple[str, Any]]:
     """
     :return: A frozen set of (key,value) pairs
+
+    Note: Converts unhashable values (dict, list) to hashable representations
     """
 
-    return frozenset({(k, v) for k, v in point.items() if k in group})
+    def make_hashable(
+        value: float | list | str | dict | None,
+    ) -> int | float | tuple | str | None:
+        """Convert unhashable types to hashable equivalents"""
+        if isinstance(value, dict):
+            # Convert dict to sorted tuple of items
+            return tuple(sorted(value.items()))
+        if isinstance(value, list):
+            # Convert list to tuple
+            return tuple(value)
+        # Return value as-is if already hashable
+        return value  # type: ignore[return-value]
+
+    return frozenset({(k, make_hashable(v)) for k, v in point.items() if k in group})
 
 
 def _build_groups_dict(
--- a/plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing.yaml
+++ b/plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing.yaml
@@ -60,7 +60,9 @@ test-deployment-v1:
         description: "(deployment) Docker image to use to create vllm deployments"
       propertyDomain:
         variableType: "OPEN_CATEGORICAL_VARIABLE_TYPE"
-        values: ["vllm/vllm-openai:v0.14.0"]
+        values:
+          - {image: "vllm/vllm-openai:v0.14.0", vllm_version: "0.14.0"}
+          - {image: "vllm/vllm-openai:latest", vllm_version: "0.21.0"}
     - identifier: n_cpus
       metadata:
         description: "(deployment) the number of CPUs to use"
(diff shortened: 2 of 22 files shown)


orchestrator/core/discoveryspace/group_samplers.py
```diff
@@ -55,9 +55,21 @@ def _build_point_group_values(
...
+        value: float | list | str | None,
+    ) -> int | float | tuple | str | None:
+        """Convert unhashable types to hashable equivalents"""
+        match value:
+            case list():
+                return tuple(value)
+            case _:
+                return value  # type: ignore[return-value]
+
+    return frozenset({(k, make_hashable(v)) for k, v in point.items() if k in group})
```
AlessandroPomponio, line 72: This function only seems to cast lists to tuples and not do anything to dicts

```suggestion
    Note: Converts unhashable values (list) to hashable representations (tuple)
    """
    return frozenset(
        (k, tuple(v) if isinstance(v, list) else v)
        for k, v in point.items()
        if k in group
    )
```

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing.yaml
```diff
@@ -60,7 +60,9 @@ test-deployment-v1:
         description: "(deployment) Docker image to use to create vllm deployments"
       propertyDomain:
         variableType: "OPEN_CATEGORICAL_VARIABLE_TYPE"
-        values: ["vllm/vllm-openai:v0.14.0"]
+        values:
+          - ["vllm/vllm-openai:v0.14.0", "0.14.0"]
+          - ["vllm/vllm-openai:latest", "0.21.0"]
```
AlessandroPomponio, line 65: I'd advise against using latest, as it would make experiments unreliable

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing_geospatial.yaml
```diff
@@ -130,7 +149,9 @@ performance_testing-geospatial-full:
         description: "(deployment) Docker image to use to create vllm deployments"
       propertyDomain:
         variableType: "OPEN_CATEGORICAL_VARIABLE_TYPE"
-        values: ["your/image/with/vllm/and/terratorch:0.1"]
+        values:
+          - ["your/image/with/vllm/and/terratorch:0.1", "0.20.0"]
+          - ["icr.io/drl-nextgen/mgazz/vllm:v0.18.0-tt.v1.2.5", "0.18.0"]
```
AlessandroPomponio, line 154: Either we provide a publicly accessible image or I'd remove this line

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/yaml_support/build_components.py
```diff
@@ -93,6 +93,8 @@ def deployment_yaml(
         skip_tokenizer_init: bool = False,
         io_processor_plugin: str | None = None,
         otlp_traces_endpoint: pydantic.AnyUrl | None = None,
+        threadpool: int = 1,
```
AlessandroPomponio, line 96: Maybe I'd rename this to `use_threadpool` to make it clearer and treat it as a bool in Python (basically just changing the type annotation)

mgazz: sure

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing_geospatial.yaml
mgazz: we could push some images on quay.

mgazz:
Regarding the need for Cached Measurements

For vLLM < 0.20.0 Threadpool is not supported regardless of user settings, therefore `renderer_num_workers` parameter has no effect on performance. As a result, testing with renderer_num_workers=32 vs renderer_num_workers=64 would produce identical results. 

To avoid running the same test with vLLM < 0.20.0 many times, for different values of `rendered_num_workers` that is not used, we cache the first run and reused it. This basically translate in: "For your image with vLLM < 0.20.0 setting different values of rendered_num_workers does not make a difference". 

This is a way to circumvent a limitation of ado where I cannot conditionally select when a property applies given the value of another property.

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing.yaml
```diff
@@ -125,10 +127,23 @@ test-deployment-v1:
...
+        description: "(deployment) Number of renderer workers when threadpool is enabled. Only used when threadpool=1."
+      propertyDomain:
+        variableType: 'DISCRETE_VARIABLE_TYPE'
+        domainRange: [1, 128]
+        interval: 1
   defaultParameterization:
     - property:
         identifier: 'image'
-      value: "vllm/vllm-openai:v0.14.0"
+      value: ["vllm/vllm-openai:v0.14.0", "0.14.0"]
```
christian-pinto, line 146: Isn't ado going to treat these values as two different images and therefore the second will fail? THis comment applies to all instances with the same style.

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing.yaml
christian-pinto: +1

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/yaml_support/build_components.py
```diff
@@ -174,6 +178,8 @@ def deployment_yaml(
 
         vllm_serve_args = [
             model,
+            "--max-num-seqs",
+            "256",
```
christian-pinto, line 182:
```suggestion
            "--max-num-seqs",
            str(max_num_seq),
```

christian-pinto (changes requested):
In addition to the comments below, these are three different PRs in my opinion:

- Adding support to threadpool
- measurements cache
- vLLM version checking

please split the three contributions otherwise the PR is not straightforward to review

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing.yaml
mgazz: no, it's a list of lists.

mgazz:
Splitting in different PRs. I will fix comments in each PR.

PR #1016, opened 2026-06-09
feat(vllm): add version checking utility for vLLM compatibility

mgazz:
- VLLMVersionChecker class to parse vLLM versions from image values
- supports_threadpool method to check version compatibility (>= 0.20.0)
- Comprehensive tests for version parsing and threadpool support detection

### DIFF
--- a/plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
+++ b/plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
@@ -1,0 +1,41 @@
+# Copyright IBM Corporation 2025, 2026
+# SPDX-License-Identifier: MIT
+
+"""Utilities for vLLM version checking and threadpool support detection."""
+
+from packaging import version
+
+
+class VLLMVersionChecker:
+    """Utility class for checking vLLM version and threadpool support."""
+
+    THREADPOOL_MIN_VERSION = "0.20.0"
+
+    @staticmethod
+    def parse_version(image_value: list | str) -> str | None:
+        """Extract vLLM version from image property value."""
+        if isinstance(image_value, list) and len(image_value) > 1:
+            return image_value[1]
+        return None
+
+    @classmethod
+    def supports_threadpool(
+        cls, image_value: list | str, threadpool_requested: int
+    ) -> bool:
+        """Check if threadpool is supported and should be enabled."""
+        if threadpool_requested == 0:
+            return False
+
+        vllm_version_str = cls.parse_version(image_value)
+        if vllm_version_str is None:
+            return True
+
+        try:
+            vllm_ver = version.parse(vllm_version_str)
+            min_ver = version.parse(cls.THREADPOOL_MIN_VERSION)
+            return vllm_ver >= min_ver
+        except Exception:
+            return True
+
+
+# Made with Bob
--- a/plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
+++ b/plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
@@ -36,6 +36,3 @@ class VLLMVersionChecker:
             return vllm_ver >= min_ver
         except Exception:
             return True
-
-
-# Made with Bob
--- a/plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
+++ b/plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
@@ -11,28 +11,62 @@ class VLLMVersionChecker:
 
     THREADPOOL_MIN_VERSION = "0.20.0"
 
-    @staticmethod
-    def parse_version(image_value: list | str) -> str | None:
-        """Extract vLLM version from image property value."""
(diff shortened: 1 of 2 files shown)


plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
```diff
@@ -0,0 +1,38 @@
...
+    """Utility class for checking vLLM version and threadpool support."""
+
+    THREADPOOL_MIN_VERSION = "0.20.0"
+
+    @staticmethod
+    def parse_version(image_value: list | str) -> str | None:
+        """Extract vLLM version from image property value."""
+        if isinstance(image_value, list) and len(image_value) > 1:
+            return image_value[1]
+        return None
```
AlessandroPomponio: on a string this always returns None - is this expected?

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @classmethod
+    def supports_threadpool(
+        cls, image_value: list | str, threadpool_requested: int
+    ) -> bool:
+        """Check if threadpool is supported and should be enabled."""
```
AlessandroPomponio: This is a bit counterintuitive: threadpool support is one thing and "should be enabled" is another

I would expect a function called `supports_threadpool` to return True if the vLLM version supports it, not whether I requested it or not

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
```diff
@@ -0,0 +1,38 @@
...
+        vllm_version_str = cls.parse_version(image_value)
+        if vllm_version_str is None:
+            return True
+
+        try:
+            vllm_ver = version.parse(vllm_version_str)
+            min_ver = version.parse(cls.THREADPOOL_MIN_VERSION)
+            return vllm_ver >= min_ver
+        except Exception:
+            return True
```
AlessandroPomponio: Returning True on exception/None seems incorrect

AlessandroPomponio (changes requested):

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
mgazz: If we cannot parse the version we assume that threadpool is enabled. Especially considering that all new vllm versions will support threadpool.

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
```diff
@@ -0,0 +1,72 @@
...
+        - "0.20.1" -> "0.20.1" (already a version)
+
+        Args:
+            image: Container image string or version string
+
+        Returns:
+            Extracted version string, or None if version cannot be extracted
+        """
+        if not image or not isinstance(image, str):
+            return None
```
christian-pinto, line 31: shouldn't this raise an exception instead? I mean, you should never pass None (as per the args) or something that is not a string, in such case there is a problem somewhere else.

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
```diff
@@ -0,0 +1,72 @@
...
+        if ":" in image:
+            tag = image.split(":")[-1]
+            # Remove leading 'v' if present
+            tag = tag.removeprefix("v")
+            return tag or None
+
+        # If no colon, assume it's already a version string
+        # Remove leading 'v' if present
+        if image.startswith("v"):
+            return image[1:]
```
christian-pinto, line 43: I think you should still check that this is a valid version string, assuming the x.y.z format.

mgazz:
@AlessandroPomponio @christian-pinto please take another look

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
```diff
@@ -0,0 +1,72 @@
...
+from packaging import version
+
+
+class VLLMVersionChecker:
+    """Utility class for checking vLLM version and threadpool support."""
+
+    THREADPOOL_MIN_VERSION = "0.20.0"
+
+    @classmethod
+    def extract_version_from_image(cls, image: str) -> str | None:
```
christian-pinto, line 15: In general this would work only in the case people have the vLLM version as image tag, but it would return wrong information for any other image. Meaning, this would not necessarily work with images that are not the official vLLM images. This is the case for the geospatial experiments where a different image is neeed and even using an official one from RedHad RHAIIS

Example:
   `registry.redhat.io/rhaii/vllm-cuda-rhel9:3.4.1-1780356914`

this one uses vLLM 0.18.0

christian-pinto (changes requested):
This class is not used anywhere, we can't comment on how it is or could be used used.

mgazz:
Good, version checking is used only in the context of threadpool. I will open another PR with both version checking and threadpool enablement.

PR #1019, opened 2026-06-09
fix(autoconf): avoid_oom_recommender considers original number_gpus

### LINKED ISSUE #960: bug(autoconf): avoid_oom_recommender does not perform as per intention
## Issue Description

The avoid_oom_recommender in autoconf is intended to check if a given configuration will result in OOM, or provide an alternative configuration that will result in the opposite . However, it's current implementation merely provides a number of GPUs that will avoid OOM for a `per_device_train_batch_size` value. This makes it nearly indistinguishable from min GPU recommender, the only difference being that the latter takes in a `batch_size` (or effective batch size) value

We need to update avoid_oom_recommender to perform as per the intended behaviour.

### Python/ado/system info

Please include the output of:

```terminaloutput
python --version
ado version
Your OS
```
Python: 3.10.20
ado: 1.8.1

VassilisVassiliadis:
Fixes #960 

Changes:
- The avoid_oom_recommender considers original number_gpus, it returns a novel recommendation only when it believes the original one would run out of GPU memory. It directly uses the predictor
- Removed a couple of breaking unit-tests that were using mock code


srikumar003 (approved):
LGTM

PR #1020, opened 2026-06-09
feat(vllm): add measurements cache for performance optimization

mgazz:
- Add CachedMeasurement dataclass to store measurement results
- Add cache_measurement and get_cached_measurement methods to EnvironmentManager
- Add CacheKeyBuilder utility to generate cache keys from environment and benchmark parameters
- Cache keys combine deployment config (model, GPUs, etc.) and benchmark params (num_prompts, request_rate, etc.)
- Reuse cached measurements when identical configurations are tested
- Include comprehensive tests for cache key building and parameter extraction


mgazz:
Not relevant anymore

PR #1028, opened 2026-06-10
feat(core): enable conditional disabling of plugin model validation in resources

### LINKED ISSUE #948: feat: disable validation of operation and actuator configuration parameters on read from db

When an operation or actuator configuration resource is read from the metastore validation is triggered on the parameters. This requires the correct pydantic model for the parameters. The problem is

1. If the operator version has changed, with a change in the parameters model,  the operation YAML cannot be read
2. Same for actuator configuration
3. This also annoying as
   a.  implicitly the parameters are know to be valid on store, so no need to check again
   b. stored operation YAML cannot be consumed again (it is static) so there is never a need to validate the parameters. If it is copied they will be validated on the new operation creation
   c. actuatorconfiguration can be consumed multiple time, but they should be checked at `create operation` time 

Parameter validation for these models is executed manually. IIRC this was actually the way it was done but was removed probably without consideration of the above issues sometime recently. 

1. Operation parameters are validated on `ado create operation` (inside relevant orchestrate functions)
2. ActuatorConfiguration parameters are validated on actuatorconfiguration created AND operation create (inside relevant orchestrate functions)
3. Operation and ActuatorConfiguration parameters are not validated on `ado get` (nothing to do) 

I don't see any way we can have automatic validation AND also allow reads of the resource YAML, when versions do not match.

michael-johnston:
This PR enables skipping plugin model validation on metastore reads for operations and actuator configurations. `ado get`, `getResource`, and `ado upgrade` no longer require plugins to be present.

Create, dry-run, and runtime use paths still validate against installed plugins; 

**Before**: Reading operation or actuator configuration resources from the metastore ran plugin registry validation, so missing or outdated plugins could block ado get and upgrades when only structural access was needed.

**After**: Metastore reads skip plugin validation via a pydantic validation context. Full validation still runs on create and on use paths (get_actuator_configurations, run_experiment) by passing ignore_plugin_validation=False.

Changes

- Add ignore_plugin_validation_context and helper in orchestrator/utilities/pydantic.py
- Guard actuator and operation pydantic validators to skip when context is set
- Inject ignore context in sqlstore read paths and ado upgrade
- Add ignore_plugin_validation parameter to getResource / getResources (default True; use paths pass False)
- Remove manual validate_actuator_parameters() / validate_operator_parameters() wiring from create handlers and test fixtures
- Add tests for context-based skip on read and strong validation on create/use

Fixes #948

### DIFF
--- a/orchestrator/core/operation/config.py
+++ b/orchestrator/core/operation/config.py
@@ -14,7 +14,7 @@ from orchestrator.core.actuatorconfiguration.config import ActuatorConfiguration
 from orchestrator.core.discoveryspace.config import (
     DiscoverySpaceConfiguration,
 )
-from orchestrator.core.metadata import ConfigurationMetadata
+from orchestrator.core.metadata import ConfigurationMetadata, PackageProvenance
 from orchestrator.core.resources import CoreResourceKinds
 from orchestrator.metastore.project import ProjectContext
 from orchestrator.modules.module import (
@@ -23,7 +23,8 @@ from orchestrator.modules.module import (
     load_module_class_or_function,
 )
 from orchestrator.schema.measurementspace import MeasurementSpaceConfiguration
-from orchestrator.utilities.pydantic import ignore_plugin_validation
+from orchestrator.utilities.pydantic import ignore_plugin_validation, Pep440VersionStr
+
 
 if typing.TYPE_CHECKING:
     import orchestrator.modules.operators.base
@@ -215,7 +216,7 @@ class OperatorMetadata(pydantic.BaseModel):
         ),
     ] = None
     version: Annotated[
-        str,
+        Pep440VersionStr,
         pydantic.Field(
             description=(
                 "PEP 440 version string for the operator (e.g. '0.1.0', "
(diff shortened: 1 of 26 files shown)


orchestrator/core/operation/config.py
```diff
@@ -44,21 +44,24 @@ class DiscoveryOperationEnum(enum.Enum):
...
     project_context: ProjectContext, actuator_configuration_identifiers: list[str]
 ) -> list[ActuatorConfiguration]:
-    """Retrieves actuator configurations from the metastore
+    """Retrieves and validates actuator configurations from the metastore for use.
 
     Fetches ActuatorConfiguration resources from the metastore using the provided
-    identifiers and validates that each actuator has at most one configuration.
+    identifiers, validates parameters against actuator plugins, and checks that
+    each actuator has at most one configuration. This is the use-path fetch;
+    ``ado get`` and ``getResource`` do not call this function.
```
AlessandroPomponio, line 53: What does "This is the use-path fetch" mean?

AlessandroPomponio (commented):
I'd probably try to keep validation strong by default and disable it only when needed. For this we can use the validation context: https://pydantic.dev/docs/validation/latest/concepts/validators/#validation-context

We already use serialization context for minimizing representations: https://github.com/IBM/ado/blob/3c46721bf34f92a7b9433bb0fefc5650bc0428ad/orchestrator/schema/domain.py#L529-L547

We inject the context https://github.com/IBM/ado/blob/3c46721bf34f92a7b9433bb0fefc5650bc0428ad/orchestrator/cli/utils/resources/formatters.py#L214-L215 and then we pass it to the validation/serialization function.

I then suggest:

- Creating a `ignore_plugin_validation_context` in orchestrator/cli/utils/pydantic/constants.py
- Updating `ado get` and `getResource` to inject this context in the `model_validate` calls

orchestrator/core/operation/config.py
michael-johnston: use-path is the ado path that actual uses the resource to do something

AlessandroPomponio (approved):
LGTM thanks

PR #1032, opened 2026-06-12
feat(core): add filtering for measurements in sample store

AlessandroPomponio:
# Add Database-Level Filtering to SQLSampleStore Measurement Queries

This PR implements database-level filtering capabilities for measurement requests and results in [`SQLSampleStore`](orchestrator/core/samplestore/sql.py), enabling efficient querying of measurement data with arbitrary filter conditions.

## Overview

Adds optional `filters` parameter to key SQLSampleStore methods, allowing callers to filter measurement data at the database level rather than in-memory. This improves performance and enables more sophisticated query capabilities for CLI commands and API consumers.

### New Filter Infrastructure

- **[`MeasurementFilterBuilder`](orchestrator/core/samplestore/orm/measurements_filtering.py)**: SQLAlchemy-based filter builder that handles:
  - Direct column access for common fields (status, requestIndex, entityIdentifier)
  - JSON field queries with dialect-specific SQL generation (MySQL `JSON_CONTAINS` vs SQLite simulation)
  - Proper type handling and value parsing

### Enhanced SQLSampleStore Methods

Updated three core methods to accept optional filters:

1. **[`measurement_requests_for_operation()`](orchestrator/core/samplestore/sql.py:1519)**: Filter requests by status, requestIndex, metadata fields, etc.
2. **[`measurement_results_for_operation()`](orchestrator/core/samplestore/sql.py:1619)**: Filter results by entityIdentifier, uid, or nested data fields
3. **[`complete_measurement_request_with_results_timeseries()`](orchestrator/core/samplestore/sql.py:1840)**: Apply filters to timeseries queries

### Filter Format

Filters use the existing `prepare_query_filters_for_db()` format for consistency with metastore filtering:
```python
filters = [{"$.status": json.dumps("Success")}, {"$.requestIndex": json.dumps(0)}]
```

## Implementation Details

- **Column Mappings**: Direct mappings for common fields avoid JSON parsing overhead
- **Dialect Support**: Works with both MySQL and SQLite backends
- **AND Logic**: Multiple filters are combined with AND
- **Backward Compatible**: `filters` parameter is optional (defaults to `None`)
- **SQLAlchemy Migration**: Converted raw SQL queries to SQLAlchemy for better composability

## Testing

Comprehensive integration tests in [`tests/samplestore/test_measurements_filtering.py`](tests/samplestore/test_measurements_filtering.py:1) covering:
- Single and multiple filter conditions
- Direct column and JSON field filtering
- Empty result handling
- Dollar-prefix notation support
- Invalid path error handling

## Performance

- Direct column filters use indexed columns where available
- JSON queries only used when necessary
- No significant performance regression for unfiltered queries


orchestrator/core/samplestore/orm/measurements_filtering.py
```diff
@@ -0,0 +1,230 @@
...
+# Direct column mappings for measurement requests
+# Maps field names to actual database column names
+MEASUREMENT_REQUEST_COLUMN_MAPPINGS = {
+    "requestid": "request_id",  # Schema field: requestid (lowercase)
+    "requestIndex": "request_index",  # Schema field: requestIndex (camelCase)
+    "status": "status",
+    "operation_id": "operation_id",  # Schema field: operation_id (snake_case)
+    "timestamp": "timestamp",
+    "type": "type",
+    "experimentReference": "experiment_reference",  # Stored as stringified text
```
christian-pinto: for completeness you should mention the schema field in here

orchestrator/core/samplestore/orm/measurements_filtering.py
```diff
@@ -0,0 +1,230 @@
...
+
+        Raises:
+            ValueError: If path is not a known column or JSON field
+        """
+        # Strip leading "$." if present for compatibility
+        path = path.removeprefix("$.")
+
+        # Check if path maps to a direct column (top-level field)
+        if path in MEASUREMENT_REQUEST_COLUMN_MAPPINGS:
+            column_name = MEASUREMENT_REQUEST_COLUMN_MAPPINGS[path]
```
christian-pinto:
```suggestion
        column_name = MEASUREMENT_REQUEST_COLUMN_MAPPINGS.get(path)
        if column_name:
```

orchestrator/core/samplestore/orm/measurements_filtering.py
```diff
@@ -0,0 +1,230 @@
...
+
+        Raises:
+            ValueError: If path is not a known column or data field
+        """
+        # Strip leading "$." if present for compatibility
+        path = path.removeprefix("$.")
+
+        # Check if path maps to a direct column (top-level field)
+        if path in MEASUREMENT_RESULT_COLUMN_MAPPINGS:
+            column_name = MEASUREMENT_RESULT_COLUMN_MAPPINGS[path]
```
christian-pinto:
```suggestion
        column_name = MEASUREMENT_RESULT_COLUMN_MAPPINGS.get(path):
        if column_name:
```

christian-pinto: just not to access the same data structure twice unnecessarily.

orchestrator/core/samplestore/sql.py
```diff
@@ -407,6 +394,53 @@ def __init__(
...
+                all_tables_exist = all(
+                    table_name in metadata.tables for table_name in table_names
+                )
+
+                if not all_tables_exist:
+                    # Create tables and use the returned metadata which already has table definitions
+                    metadata = self._create_source_table()
+            except InvalidRequestError:
+                # Tables don't exist yet, create them
+                metadata = self._create_source_table()
```
christian-pinto: Is this looking at the same tables? If yes wouldn't the catch statement serve the same goal as the `if not all_tables_exist` statement?

christian-pinto (changes requested):

orchestrator/core/samplestore/orm/measurements_filtering.py
AlessandroPomponio: The schema field is the key and the db column is the value. The comment in this line was mainly there to clarify that despite being an object, the experiment reference is saved in its string representation. I have removed the comments in 7e7ac5c828445bae1087a978f112950fc7710168

orchestrator/core/samplestore/sql.py
AlessandroPomponio: When you use `metadata.reflect` and specify the `only` parameter, the method will raise an `InvalidRequestError` if at least one of the tables does not exist. We could set all_tables_exist to False before the try/except and just pass on InvalidRequestError to simplify the code a bit

AlessandroPomponio: Well, actually, given the semantics of `reflect` we can assume that an error will always be raised if at least one doesn't exist

AlessandroPomponio: e8872c614f0608292af4452bed10656ff8372f2c

christian-pinto (approved):

PR #1039, opened 2026-06-15
feat(vllm_performance): add threadpool support for vLLM IOProcessors

mgazz:
This PR adds threadpool rendering support for vLLM geospatial models (v0.20.0+) with automatic version validation, and added support for grouped sampling when entity properties contain list values.

**Changes:**

* New parameters: use_threadpool and renderer_num_workers for controlling parallel rendering
* Version validation: Automatically checks vLLM version compatibility before enabling threadpool
* Error handling: New UnsupportedThreadpoolConfigurationError prevents incompatible configurations
* Image format: Supports [image_url, version] format for explicit version specification


orchestrator/core/discoveryspace/group_samplers.py
```diff
@@ -57,7 +57,19 @@ def _build_point_group_values(
     :return: A frozen set of (key,value) pairs
     """
 
-    return frozenset({(k, v) for k, v in point.items() if k in group})
+    def make_hashable(
+        value: list | tuple | str | float | bool,
```
AlessandroPomponio, line 61:
```suggestion
        value: list | tuple | str | int | float | bool,
```

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/yaml_support/build_components.py
```diff
@@ -93,6 +93,8 @@ def deployment_yaml(
         skip_tokenizer_init: bool = False,
         io_processor_plugin: str | None = None,
         otlp_traces_endpoint: pydantic.AnyUrl | None = None,
+        threadpool: int = 1,
```
AlessandroPomponio: I'd suggest using a bool for this - 0 and 1 work as bools anyway
I suggest calling it `use_threadpool` as well

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/create_environment.py
```diff
@@ -41,6 +41,8 @@ def create_test_environment(
     skip_tokenizer_init: bool = False,
     io_processor_plugin: str | None = None,
     otlp_traces_endpoint: pydantic.AnyUrl | None = None,
+    threadpool: int = 1,
```
AlessandroPomponio: I'd suggest using a bool for this - 0 and 1 work as bools anyway
I suggest calling it `use_threadpool` as well

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/manage_components.py
```diff
@@ -262,6 +262,8 @@ def create_deployment(
         skip_tokenizer_init: bool = False,
         io_processor_plugin: str | None = None,
         otlp_traces_endpoint: pydantic.AnyUrl | None = None,
+        threadpool: int = 1,
```
AlessandroPomponio: I'd suggest using a bool for this - 0 and 1 work as bools anyway
I suggest calling it `use_threadpool` as well

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
```diff
@@ -190,12 +196,36 @@ def _create_environment(
...
+                        )
+                        is_threadpool_allowed = VLLMVersionChecker.supports_threadpool(
+                            vllm_version_str
+                        )
+                        image_name = image_value
+                    elif type(image_value) is list:
+                        is_threadpool_allowed = VLLMVersionChecker.supports_threadpool(
+                            image_value[1]
+                        )
+                        image_name = image_value[0] if len(image_value) > 0 else ""
```
AlessandroPomponio: This conditional assignment doesn't help - in L211 you access `image_value[1]` already

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
```diff
@@ -0,0 +1,72 @@
...
+        return image
+
+    @classmethod
+    def supports_threadpool(cls, vllm_version_str: str) -> bool:
+        """
+        Check if threadpool is supported. If version cannot be parsed we return True
+        to avoid halting test campaigns when we don't have version info.
+
+        New versions of vLLM will have threadpool support enabled by default therefore
+        is more likely that the version is supported than not supported.
```
AlessandroPomponio:
```suggestion
        Starting from 0.20.0, vLLM versions have threadpool support enabled by default.
        We optimistically return True unless we clearly have an unsupported version.
```

plugins/actuators/vllm_performance/tests/test_experiment_executor.py
AlessandroPomponio, line 1: These tests don't test anything meaningful, I'd remove them

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing_geospatial.yaml
```diff
@@ -207,10 +207,23 @@ performance_testing-geospatial-full:
       propertyDomain:
         variableType: "OPEN_CATEGORICAL_VARIABLE_TYPE"
         values: [None, "terratorch_segmentation"]
+    - identifier: 'use_threadpool'
+      metadata:
+        description: "Enable threadpool for vLLM renderer (0=disabled, 1=enabled). When enabled, uses multiple workers for rendering."
+      propertyDomain:
```
christian-pinto: Would it make more sense to use a discrete type here so that 0 is False and 1 is True?

I comment only here but it is intended for all instances of `use_threadpool`

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
```diff
@@ -83,6 +87,8 @@ def _build_entity_env(values: dict[str, str]) -> str:
         "dtype": values.get("dtype"),
         "cpu_offload": values.get("cpu_offload"),
         "max_num_seq": values.get("max_num_seq"),
+        "use_threadpool": values.get("use_threadpool"),
```
christian-pinto: wouldn't `values.get("use_threadpool")` return a string in this instance? In all the functions `threadpool` is an integer.

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/yaml_support/build_components.py
christian-pinto: I'd call it `use_threadpool`

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/create_environment.py
christian-pinto: I'd call it use_threadpool

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/create_environment.py
```diff
@@ -119,6 +123,8 @@ def create_test_environment(
         skip_tokenizer_init=skip_tokenizer_init,
         io_processor_plugin=io_processor_plugin,
         otlp_traces_endpoint=otlp_traces_endpoint,
+        threadpool=threadpool,
```
christian-pinto: I'd call it use_threadpool

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/manage_components.py
christian-pinto: I'd call it use_threadpool

plugins/actuators/vllm_performance/yamls/vllm_deployment_space.yaml
```diff
@@ -6,9 +6,25 @@ entitySpace:
       values:
         - ibm-granite/granite-3.3-8b-instruct
   - identifier: image
+    metadata:
+      description: Docker image with vLLM. Specify as [image, version] for threadpool validation
     propertyDomain:
       values:
-        - quay.io/dataprep1/data-prep-kit/vllm_image:0.1
+        - - quay.io/dataprep1/data-prep-kit/vllm_image:0.1
+          - "0.20.1"
```
AlessandroPomponio: This image is very old - we should replace it with something newer, maybe from Red Hat's OpenShift AI images: https://quay.io/repository/modh/vllm?tab=tags


```suggestion
        - - quay.io/dataprep1/data-prep-kit/vllm_image:0.1
          - "0.6.3"
```

AlessandroPomponio (changes requested):

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
```diff
@@ -0,0 +1,72 @@
...
+from packaging import version
+
+
+class VLLMVersionChecker:
+    """Utility class for checking vLLM version and threadpool support."""
+
+    THREADPOOL_MIN_VERSION = "0.20.0"
+
+    @classmethod
+    def extract_version_from_image(cls, image: str) -> str:
```
christian-pinto:
```suggestion
    def extract_version_from_image(cls, image: str) -> str | None:
```

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
```diff
@@ -0,0 +1,72 @@
...
+            # Remove leading 'v' if present
+            tag = tag.removeprefix("v")
+            return tag or None
+
+        # If no colon, assume it's already a version string
+        # Remove leading 'v' if present
+        if image.startswith("v"):
+            return image[1:]
+
+        return image
```
christian-pinto: What if the image name is a string and it does not follow any of the supported formats?

Here I see that if nothing matches, you just return the image name.

I would do two things:
- Raise an exception when then the image version cannot be resolved
- Return None when the image version cannot be resolved

The caller will then have to either catch and handle the exception or check the return value.

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
```diff
@@ -0,0 +1,72 @@
...
+        Returns:
+            True if threadpool is supported, False otherwise
+        """
+        try:
+            vllm_ver = version.parse(vllm_version_str)
+            min_ver = version.parse(cls.THREADPOOL_MIN_VERSION)
+            return vllm_ver >= min_ver
+
+        except Exception:
+            return True
```
christian-pinto, line 80: Is it safe to return True here? In case of exception something went wrong so we are not able to actually assess whether this instance does support threadpool or not.

christian-pinto (changes requested):

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
mgazz: The idea is that we assume threadpool is supported if we cannot parse the version based on semantic versioning. If threadpool is not supported the test will fail. We do not want to impose a format on the image tag as this approach would be too restrictive.

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
mgazz: fair point

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
mgazz: Fair point.

plugins/actuators/vllm_performance/tests/test_experiment_executor.py
mgazz: removed in 78b8ed2

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/yaml_support/build_components.py
mgazz: Fixed in [a7ee192](https://github.com/IBM/ado/pull/1039/commits/a7ee1929ccf627abb79f2ea01fa6a477704b630e)

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/create_environment.py
mgazz: Fixed in [a7ee192](https://github.com/IBM/ado/pull/1039/commits/a7ee1929ccf627abb79f2ea01fa6a477704b630e)

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/manage_components.py
mgazz: Fixed in [a7ee192](https://github.com/IBM/ado/pull/1039/commits/a7ee1929ccf627abb79f2ea01fa6a477704b630e)

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing_geospatial.yaml
mgazz: Fixed in [a7ee192](https://github.com/IBM/ado/pull/1039/commits/a7ee1929ccf627abb79f2ea01fa6a477704b630e)

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
mgazz: Fixed in [a7ee192](https://github.com/IBM/ado/pull/1039/commits/a7ee1929ccf627abb79f2ea01fa6a477704b630e)

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/yaml_support/build_components.py
mgazz: Fixed in [a7ee192](https://github.com/IBM/ado/pull/1039/commits/a7ee1929ccf627abb79f2ea01fa6a477704b630e)

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/create_environment.py
mgazz: Fixed in [a7ee192](https://github.com/IBM/ado/pull/1039/commits/a7ee1929ccf627abb79f2ea01fa6a477704b630e)

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/manage_components.py
mgazz: Fixed in [a7ee192](https://github.com/IBM/ado/pull/1039/commits/a7ee1929ccf627abb79f2ea01fa6a477704b630e)

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
mgazz: fixed in 09ed8f41f87859c77d58a4d4288f25b4058a475b

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
mgazz: reworked logic and added tests in [ab47b4a](https://github.com/IBM/ado/pull/1039/commits/ab47b4afa90a12882069d11b86ef4a3acdd07135)

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
AlessandroPomponio: In L229 in plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py this has become 

```python
bool(values.get("use_threadpool", True))
```

You might want to change all of the instances left @mgazz

plugins/actuators/vllm_performance/tests/test_version_utils.py
```diff
@@ -65,3 +67,40 @@ def test_supports_threadpool_with_old_image_extraction(self) -> None:
...
+        """Test that image without tag (no colon) removes leading 'v' if present."""
+        image = "vllm/vllm-openai"
+        assert (
+            VLLMVersionChecker.extract_version_from_image(image) == "vllm/vllm-openai"
+        )
+
+    def test_extract_version_from_image_without_tag(self) -> None:
+        """Test that image without tag (no colon) removes leading 'v' if present."""
+        image = "custom-image"
+        assert VLLMVersionChecker.extract_version_from_image(image) is None
```
AlessandroPomponio: These tests don't do what their description says

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
```diff
@@ -218,6 +260,8 @@ def _create_environment(
                         enforce_eager=values.get("enforce_eager", 0) == 1,
                         io_processor_plugin=values.get("io_processor_plugin"),
                         otlp_traces_endpoint=otlp_traces_endpoint,
+                        use_threadpool=threadpool_requested,
+                        renderer_num_workers=int(values.get("renderer_num_workers")),
```
AlessandroPomponio: Will raise an error if render_num_workers is not provided due to the cast - I suggest adding a default value

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
```diff
@@ -190,12 +197,47 @@ def _create_environment(
                     )
                 )
                 try:
+                    image_value = values.get("image", "")
```
AlessandroPomponio, line 224: The "" default will cause a VLLMVersionExtractionError which could be confusing

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
```diff
@@ -190,12 +197,47 @@ def _create_environment(
...
+                            )
+                        else:
+                            # Assume threadpool is allowed if cannot infer vllm version
+                            is_threadpool_allowed = True
+                        image_name = image_value
+                    elif type(image_value) is list:
+                        is_threadpool_allowed = VLLMVersionChecker.supports_threadpool(
+                            image_value[1]
+                        )
+                        image_name = image_value[0]
```
AlessandroPomponio: Here you check that the value is a list (which I think you should do with `isinstance`) but you don't check that it has more that 1 value. Given your structure, I'd also check that it has exactly two values and raise something if that isn't the case

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
```diff
@@ -0,0 +1,77 @@
...
+
+        Returns:
+            True if threadpool is supported, False otherwise
+        """
+        try:
+            vllm_ver = version.parse(vllm_version_str)
+            min_ver = version.parse(cls.THREADPOOL_MIN_VERSION)
+            return vllm_ver >= min_ver
+
+        except Exception:
```
AlessandroPomponio:
```suggestion
        except version.InvalidVersion:
```

orchestrator/core/discoveryspace/group_samplers.py
AlessandroPomponio: This hasn't been resolved. Also, not sure if you need to handle other unhashable data structures like dicts here

AlessandroPomponio (changes requested):

plugins/actuators/vllm_performance/tests/test_version_utils.py
mgazz: sorry missed to update the description after a copy. Will fix now

orchestrator/core/discoveryspace/group_samplers.py
mgazz: This goes against a ruff rule: https://docs.astral.sh/ruff/rules/redundant-numeric-union/

```
PYI041 [*] Use `float` instead of `int | float`
  --> orchestrator/core/discoveryspace/group_samplers.py:61:16
   |
60 |     def make_hashable(
61 |         value: list | tuple | str | int | float | bool,
   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
62 |     ) -> tuple | str | float | bool:
63 |         """
   |
help: Remove redundant type
```

plugins/actuators/vllm_performance/tests/test_version_utils.py
mgazz: Addressed with   e647dbc

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
mgazz: That's right. It does not make sense for the image to be empty. At this point I would raise an exception `EmptyImageValueError`. How does that sound?

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
mgazz: good catch 👍

mgazz: I need to review the logic. we want 
- use_threadpool set to true only if we set renderer_num_workers. 
- If renderer_num_workers is not set, use_threadpool should be disabled. 

@christian-pinto do you agree on this approach?

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
```diff
@@ -53,6 +58,33 @@
...
+    :param values: experiment values dictionary
+    :return: True if threadpool-related functionality should be enabled
+    :raises ValueError: if renderer_num_workers is negative
+    """
+    if "renderer_num_workers" not in values:
+        return False
+
+    renderer_workers = values.get("renderer_num_workers")
+    if renderer_workers is None:
+        return False
```
AlessandroPomponio: This can be simplified using a default in the `get`
```suggestion
    renderer_workers = values.get("renderer_num_workers", None)
    if renderer_workers is None:
        return False
```

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
```diff
@@ -53,6 +58,33 @@
...
+
+    renderer_workers_int = int(renderer_workers)
+    if renderer_workers_int < 0:
+        raise ValueError(
+            f"renderer_num_workers must be non-negative, got {renderer_workers}"
+        )
+
+    return renderer_workers_int > 0
```
AlessandroPomponio: This is correct but I personally would have it be

```suggestion
    return renderer_workers_int != 0
```

This is because, at a glance, I saw this after `if renderer_workers_int < 0:` and I thought: well, of course it is, not realising it was checking basically whether this wasn't 0

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
```diff
@@ -190,12 +225,55 @@ def _create_environment(
                     )
                 )
                 try:
+                    if not (image_value := values.get("image", "")):
```
AlessandroPomponio: I'd split the assignment to the check, in this case the walrus operator isn't super readable

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
```diff
@@ -53,6 +58,33 @@
 logger = logging.getLogger(__name__)
 
 
+def _determine_threadpool_usage(values: dict[str, Any]) -> bool:
```
AlessandroPomponio: I'd call this `_is_threadpool_requested`, it's more to-the-point

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
```diff
@@ -218,6 +296,12 @@ def _create_environment(
                         enforce_eager=values.get("enforce_eager", 0) == 1,
                         io_processor_plugin=values.get("io_processor_plugin"),
                         otlp_traces_endpoint=otlp_traces_endpoint,
+                        renderer_num_workers=(
+                            int(values.get("renderer_num_workers"))
+                            if values.get("renderer_num_workers") is not None
+                            and int(values.get("renderer_num_workers")) > 0
```
AlessandroPomponio: From the point of view of this assignment, 0 and None are equal, so this is a bit more straightforward:

```suggestion
                            if int(values.get("renderer_num_workers"), 0) > 0
```

plugins/actuators/vllm_performance/tests/test_threadpool_logic.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def test_with_positive_workers(self) -> None:
+        """renderer_num_workers=1 -> True (threadpool enabled)."""
+        values = {"renderer_num_workers": 1}
+        assert _determine_threadpool_usage(values) is True
+
+    def test_with_large_positive_workers(self) -> None:
+        """renderer_num_workers=64 -> True (threadpool enabled)."""
+        values = {"renderer_num_workers": 64}
+        assert _determine_threadpool_usage(values) is True
```
AlessandroPomponio: Maybe a bit redundant

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
```diff
@@ -0,0 +1,79 @@
...
+                return version.parse(tag).base_version
+            except version.InvalidVersion:
+                return tag
+
+        try:
+            return version.parse(image).base_version
+        except version.InvalidVersion:
+            if "/" in image:
+                return image
+            return None
```
AlessandroPomponio: The docstring says

>         Returns:
>             Extracted version string
> 
>         Raises:
>             VLLMVersionExtractionError: If version cannot be extracted from image string

But at L50 you return what you expect to be a string like `vllm/vllm-openai` (see `test_extract_version_from_vllm_image_without_tag`) and at L52 you return `None`.

Both are cases where "version cannot be extracted from image string", so I'd expect to see a `VLLMVersionExtractionError` being raised, according to the docstring

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing_geospatial.yaml
```diff
@@ -933,10 +992,23 @@ performance_testing-geospatial-guidellm-deployment-custom-dataset:
       propertyDomain:
         variableType: "OPEN_CATEGORICAL_VARIABLE_TYPE"
         values: [None, "terratorch_segmentation"]
+    - identifier: 'use_threadpool'
+      metadata:
+        description: "Enable threadpool for vLLM renderer. When enabled, uses multiple workers for rendering."
+      propertyDomain:
+        variableType: 'CATEGORICAL_VARIABLE_TYPE'
+        values: [false, true]
```
AlessandroPomponio: not used anymore

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing_geospatial.yaml
```diff
@@ -982,6 +1054,12 @@ performance_testing-geospatial-guidellm-deployment-custom-dataset:
     - property:
         identifier: 'io_processor_plugin'
       value: "terratorch_segmentation"
+    - property:
+        identifier: 'use_threadpool'
+      value: false
```
AlessandroPomponio: not used anymore

website/docs/examples/vllm-performance-geospatial.md
```diff
@@ -262,6 +264,36 @@ entitySpace:
...
+## Threadpool Rendering for Performance
+
+For vLLM versions 0.20.0 and later, you can enable threadpool rendering to
+improve performance when processing satellite imagery:
+
+```yaml
+entitySpace:
+  - identifier: use_threadpool
+    propertyDomain:
+      values: [0, 1]  # 0=disabled, 1=enabled
```
AlessandroPomponio: It's `[false, true]` now, no? (not that it would change anything)

AlessandroPomponio (changes requested):

website/docs/examples/vllm-performance-geospatial.md
mgazz: good catch

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
mgazz: good catch

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
mgazz: Fixed in [df3cd2a](https://github.com/IBM/ado/pull/1039/commits/df3cd2afbbff6ba2c7a19689d2a5c06a044eb63c)

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
mgazz: Fixed in [62479b0](https://github.com/IBM/ado/pull/1039/commits/62479b0e7ad9793c4562101678e845c29812bbcd)

plugins/actuators/vllm_performance/tests/test_threadpool_logic.py
mgazz: Fixed in  7c0810c

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/version_utils.py
mgazz: Fixed in [15040a8](https://github.com/IBM/ado/pull/1039/commits/15040a85cf52b5fd619be6cf3e6ce537299e2496)

website/docs/examples/vllm-performance-geospatial.md
mgazz: Fixed in [fc648ee](https://github.com/IBM/ado/pull/1039/commits/fc648ee72863bef46a10aea3e6516dad5826f3fa)

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing_geospatial.yaml
mgazz: Fixed in [fc648ee](https://github.com/IBM/ado/pull/1039/commits/fc648ee72863bef46a10aea3e6516dad5826f3fa)

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing_geospatial.yaml
mgazz: Fixed in [fc648ee](https://github.com/IBM/ado/pull/1039/commits/fc648ee72863bef46a10aea3e6516dad5826f3fa)

AlessandroPomponio (approved):
LGTM thanks

christian-pinto (approved):

PR #1043, opened 2026-06-16
feat(core): add experiment versioning support

### LINKED ISSUE #1003: feat: enhanced experiment and operator versioning support
Introduce a first-class **Algorithm Version** (`Major.Minor`) that operator and experiment developers declare independently of the Python package version, governed by these rules (related to #936 ):

| Change type | Action |
|---|---|
| Logging, refactoring, memory/performance optimisation, comments | No version bump |
| Adding new outputs or inputs while keeping existing behaviour identical | Minor bump (e.g. `v1.0 → v1.1`) |
| Any change to the mathematical definition or computation | Major bump (e.g. `v1.1 → v2.0`) |

From a developer perspective:

- Every operator declares an Algorithm Version as part of its metadata, independently of its package version.
- Every experiment (including custom experiments registered via decorator) declares an Algorithm Version.
- The Algorithm Version is the developer's assertion of *semantic sameness*: same major version means "results computed by this algorithm are interchangeable regardless of what package version produced them."

From a user perspective:

- When a discoveryspace is created, the algorithm versions of all experiments in its measurement space are frozen at creation time. Old spaces remain valid historical artefacts even after an algorithm is upgraded.
- When an algorithm major version is bumped, new spaces created will automatically use the new version and existing memoised results from the old version are not reused — re-measurement happens automatically without any user action.
- When only a minor version bump occurs (extension), existing memoised results remain valid and are reused.
- Users do not need to specify algorithm versions in space YAML; they are resolved from the experiment catalog at space creation time.

NOTE: With the versioning scheme you can only have ONE installed version of an experiment/operator. This is already the case for operators. For experiments to maintain multiple executable versions you would need to adopt a different method

---

**Describe alternatives you've considered.**

- **Embedding the algorithm version in the experiment base name** (e.g. `solve_mip_v2`): one of the current solutions. This makes versioning explicit in the identifier but ties the algorithm generation to the human-readable name, requiring users to change YAML when upgrading. For decorated experiments it also requires changing the function name. Rejected in favour of a separate version field that is encoded automatically into memoisation keys.

- **Supporting multiple algorithm versions coexisting in the catalog simultaneously** (e.g. allowing a space YAML to explicitly request `solve_mip` at `v1` even after `v2` is deployed): this would allow controlled migration and side-by-side comparison. Deferred because it requires significant changes to catalog internals and is incompatible with the decorator-based experiment registration pattern. The "frozen space at creation time" behaviour covers the most common need without this complexity.

---

**Additional context.**

The major algorithm version is the only component that affects memoisation. The minor version is informational — it records that an experiment was extended — but does not break compatibility with existing results. Both components should be stored on created resources as a frozen-at-creation-time snapshot for reproducibility auditing.

Operators without a declared Algorithm Version should produce a deprecation warning at registration time to drive adoption.

michael-johnston:
This PR introduces experiment versioning. Each experiment can now optionally specify its own version (which must be strict semver if specified). An experiment with no explicit version, has version=None. This is simply handled as if the version was the empty string `""`.  

This PR also
-  introduces a **Versioning Convention** and implements features to support this convention. 
- adds additional checks on experiment compatibility beyond version

## How to Set/Specify Version

For Experiments there is now a top level "version" field. 

```
identifier: myexp
version: 1.0.0. # <- New Field. Default None
```

Similarly experiment references have a new `experimentVersion` field. In `space.yaml` you now can have
```
experiments:
    experimentVersion: 1.0.0 # (Can be None if experiments version is None)
```

Below sections explain the versioning conventions and how e.g. space.yaml is processed at creation time. 

### Fully Qualified Identifiers

Prior to this PR there were two types of experiment identifier. The base identifier (the experiment name) and the parameterized identifier (base identifier + parameterization). Version allows us to expand this set with fully qualified identifiers (FQI): 

- **base identifier**: "**_Existing"._** The name of an experiment
- **FQI**: "**_New_**". The base identifier + the full semver. String form "${BASENAME}@${FULL_SEMVER}"  
- **Parameterized FQI**: "**_New_**" The FQI + parameterization

In output of `describe space` or `describe experiment` you will see the relevant FQI now. 

### Major Version Identifiers

The above are just the existing identifiers + semver. In addition we define:

- **major version identifier**: **_New_**". The base identifier + the major version. String form "${BASENAME}@v${MAJORVERSION}" ((note the `v`)
- **parameterized major version identifier:** "**_Modified_**": The major version identifier + parameterization

The major version identifier allows matching experiments or experiment references on major version only 

## Stringified References

The string representation of an experiment reference is always a **FQI** i.e. includes full version

## Experiment Versioning Convention

The PR is based on the following convention

| Change type | Rule | Example |
| --- | --- | --- |
| Bug fix, refactoring, logging | No bump (PATCH is fine) | `1.0.0 → 1.0.1` |
| New output / input, same core behaviour | Minor bump | `1.0.0 → 1.1.0` |
| Mathematical / algorithmic change | **Major bump** | `1.0.0 → 2.0.0` |

In particular a change minor version is expected NOT to change the method of measuring a target property. For example if the target property is called "mean_latency" then experiments versions 1.1 and 1.2 should produce the same result (for deterministic experiment) or same distribution of results given same inputs. 

This can be summarised as  **"same major version identifier means same measurement"**

## Experiment Matching

The behaviour of ado with respect to experiment versioning can be summarised as **"FQI at Creation Time, Major Version Identifier at Measurement Time".**

### Space Creation

When creating a space, the experiments in the measurement space are matched to the registry using FQI. This means, for example if you do not specify a version for an experiment reference, but there is no corresponding non-versioned experiment in the registry, it is an error. 

Measurement Space Specification | Catalog Contains | Result
--- | --- | ---
solve_mip | solve_mip | Match (implicit "" version)
solve_mip | solve_mip@1.0.0 | Error - version mismatch
solve_mip@v1 | solve_mip@1.0.0 | Error in MeasurementSpace Spec as FQI required 
solve_mip@1.0.1 | solve_mip@1.0.0 | Error - FQI mismatch

This is to avoid situations where a version is put in the space.yaml, thinking it is installed, but it's not and an older/newer version is transparently used with no warning. Since major version identifiers are used at measurement time, this is the only opportunity to warn of such a mismatch, notifying the user that possible the wrong experiment version is currently installed. 

### Operation Creation

At runtime i.e. when an operation is created on a space, major version identifiers are used. 

* **Memoization**: If the requested experiment major version identifier is the same as a stored results, the stored result is used i.e.  "same major version identifier means same measurement"
* **Actuator**: When an actuator receives a request to run an experiment it must match the experiment requested based on major version identifier e.g. 

    - request to run "myexperiment@1.1.1"
    - actuator catalog as "myexperiment@1.3.1" 
    - actuator uses "myexperiment@1.3.1". (Same major version identifier, same measurement) 

## Observed Property Naming

Since  **"same major version identifier means same measurement"**, observed properties use the major version of the experiment e.g. "my_experiment@v1-mean-latency"

An experiment with no version will just have the previous observed property identifiers. 

## Multi-Versioned Experiment Rules

- Multiple major versions of an experiment with the same base name can exist in a catalog - they do different measurements
- Multiple minor versions of *the same major version** cannot exist in a catalog - they do the same measurement 

Example:

- Allowed:  `myexperiment@1.3.4` and `myexperiment@2.0.1` together
- Not allowed:  `myexperiment@1.3.4` and  `myexperiment@1.4.0`

This is because there is little reason to keep multiple minor versions of the same major version since by convention the latest minor version does the same measurements as early + extra. 

## Additional Experiment Compatibility Checks

In addition to major version compatibility, when an explore operation starts additional compatibility checks are performed between the experiment interface the discovery space expects and the interface provided by the version in the registry

- All required and parameterized optional properties the space expects are present
- The domains of this properties are same or larger (not smaller)
- The default values of non-parameterised optional properties have not changed
- The target (observed) properties the space expects are present (there can be additional observed properties)

## Scope of Changes

The changes in `orchestrator` are relatively small, with the major additional code being to "deserialise" the string form of experiment reference robustly (required for current sql store implementation). Most new code is tests. 

## Out-Of-Scope

Memoization does not check that data from a compatible major version of an experiment, measures all the expected observed properties of a the current version of the experiment. 

This is a pre-existing issue, not introduced by this PR. It will be addressed separately. 

## Summary 

-  FQI at creation time, major version identifier at measurement time
    - Creation time establishes the experiment interface the discovery space requires
-  same major version identifier means same measurement
-  the experiment interface expected by the space must be compatible with that provided by the experiment in the registry


orchestrator/cli/resources/discovery_space/create.py
```diff
@@ -194,6 +195,12 @@ def create_discovery_space(parameters: AdoCreateCommandParameters) -> str | None
                 stderr=True,
             )
             raise typer.Exit(1) from error
+        except AlgorithmVersionMismatchError as error:
```
AlessandroPomponio: If it's an experiment version mismatch, why is the class called `AlgorithmVersionMismatchError`?

orchestrator/modules/actuators/catalog.py
```diff
@@ -102,36 +118,159 @@ def identifier(self) -> str:
...
+    def experiment_semantic_identifiers(self) -> list[str]:
+        """Return the semantic identifiers of the experiments in the catalog
 
-    def experimentForReference(self, reference: ExperimentReference) -> Experiment:
-        """Returns the experiment matching reference or None if there is no match
-
-        Note: Here an experiment matches a reference if they have same actuator id and experiment id.
-        In particular if the ExperimentReference is parameterized it does not affect the matching
+        Returns:
+            Dict keyed by each experiment's semantic identifier.
```
AlessandroPomponio: Not a dict

orchestrator/modules/actuators/catalog.py
```diff
@@ -102,36 +118,159 @@ def identifier(self) -> str:
...
+        this method returns either a single match or ``None``.
+
+        Args:
+            reference: The experiment reference to look up.
+
+        Returns:
+            The matching Experiment, or None if no match is found.
+        """
+        for experiment in self.experiments:
+            if _catalog_reference_match(experiment.reference, reference):
```
AlessandroPomponio: Looks like `_catalog_reference_match` is only used here. It's also a very simple comparison that we can have inline as:

```python
experiment.reference.actuatorIdentifier == reference.actuatorIdentifier and experiment.reference.semantic_experiment_identifier == reference.semantic_experiment_identifier
```

orchestrator/modules/actuators/catalog.py
```diff
@@ -4,14 +4,30 @@
...
+    """Return True when lookup_reference matches catalog reference.
+
+    Compares actuator and semantic experiment identifier.
+    """
+    if catalog_reference.actuatorIdentifier != lookup_reference.actuatorIdentifier:
+        return False
+    return (
+        catalog_reference.semantic_experiment_identifier
+        == lookup_reference.semantic_experiment_identifier
+    )
```
AlessandroPomponio: I'd remove this - see comment on L147

orchestrator/modules/actuators/catalog.py
```diff
@@ -102,36 +118,159 @@ def identifier(self) -> str:
...
+        3. Raises :class:`~orchestrator.modules.actuators.base.DeprecatedExperimentError`
+           if the resolved experiment is deprecated.
+        4. Wraps the result in a :class:`~orchestrator.schema.experiment.ParameterizedExperiment`
+           if the reference carries parameterization.
+
+        Args:
+            reference: The experiment reference to resolve.
+            mode: ``"semantic"`` (default) — Match on MAJOR version
+                ``"fully_qualified"`` — additionally requires the exact version
+                (MAJOR.MINOR.PATCH) to match.
```
AlessandroPomponio: I would suggest renaming this to `match_on` and then have as options `major_version` and `exact_version` which more clearly convey what's happening

orchestrator/modules/actuators/catalog.py
```diff
@@ -102,36 +118,159 @@ def identifier(self) -> str:
...
+                f"An experiment with semantic identifier {experiment.semantic_identifier!r} "
+                f"is already registered in catalog {self._identifier!r}. "
             )
-            self.log.debug(
-                f"New experiment {experiment.model_dump()}, existing experiment "
-                f"{self._experiments.get(experiment.identifier).model_dump()}"
+
+        self._experiments[experiment.semantic_identifier] = experiment
+
+    def resolve_reference(
```
AlessandroPomponio: I think the name isn't really conveying what this method does. On its own, I would've called this method `experiment_by_reference`. We already have `experimentForReference` but I would say it could be merged down into this experiment if we introduce an appropriate `mode`/`match_on` (e.g., `lax` mode or `base_identifier_only` for the match_on)

orchestrator/modules/actuators/catalog.py
```diff
@@ -102,36 +118,159 @@ def identifier(self) -> str:
...
+            )
+
+        if reference.parameterization:
+            return ParameterizedExperiment(
+                parameterization=reference.parameterization, **experiment.model_dump()
+            )
+        return experiment
+
+
+class AlgorithmVersionMismatchError(Exception):
```
AlessandroPomponio: (reiterating here that I would call this `ExperimentVersionMismatchError`)

orchestrator/modules/actuators/custom_experiments.py
```diff
@@ -768,21 +774,17 @@ def submit(
...
+        try:
+            targetExperiment = self._catalog.resolve_reference(experimentReference)
+        except Exception:
             if self._catalog.experiments:
                 raise ValueError(
                     f"Requested experiments {experimentReference} is not in the CustomExperiments actuator catalog. "
-                    f"Known experiments are {list(self._catalog.experimentsMap.keys())}"
-                )
+                    f"Known experiments are {self._catalog.experiment_semantic_identifiers}"
+                ) from None
```
AlessandroPomponio: Why from None? It should be from the exception in L779

orchestrator/modules/actuators/custom_experiments.py
```diff
@@ -768,21 +774,17 @@ def submit(
...
             raise ValueError(
                 f"Requested experiments {experimentReference} is not in the CustomExperiments actuator catalog (which is empty). "
-            )
-
-        targetExperiment = self._catalog.experimentForReference(experimentReference)
-        if targetExperiment.deprecated:
-            raise DeprecatedExperimentError(
-                f"{targetExperiment.actuatorIdentifier}.{targetExperiment.identifier} is deprecated."
-            )
+            ) from None
```
AlessandroPomponio: Same here

orchestrator/modules/actuators/registry.py
```diff
@@ -444,10 +447,94 @@ def experimentForReference(
                 f"The {reference.actuatorIdentifier}  actuator was found but it did not contain "
                 f"the {reference.experimentIdentifier} experiment."
             )
-            raise UnknownExperimentError(reference)
+            message = (
+                f"The {reference.actuatorIdentifier} actuator was found but it did not "
+                f"contain the {reference} experiment."
+            )
+            if reference.experimentVersion is None and actuator_catalog:
+                candidates = actuator_catalog.experiments_matching_identifier(reference)
```
AlessandroPomponio, line 544: It's a bit weird that we have so many methods comparing references and identifiers all called similarly but doing slightly different things:

- `experiments_matching_identifier` matches on `actuatorIdentifier` and `experimentIdentifier`
- `experimentForReference` matches on `actuatorIdentifier` and `semantic_experiment_identifier`
- `resolve_reference` matches on `actuatorIdentifier`, `semantic_experiment_identifier` and `fully_qualified_identifier`

It's extremely difficult to follow

orchestrator/modules/actuators/registry.py
```diff
@@ -444,10 +447,94 @@ def experimentForReference(
...
+                    message = f"{message}Available versions: {available_versions}."
+            raise UnknownExperimentError(message)
 
         return experiment
 
+    def resolve_reference(
+        self,
+        reference: ExperimentReference,
+        additional_catalogs: list[ExperimentCatalog] | None = None,
+        mode: typing.Literal["semantic", "fully_qualified"] = "semantic",
```
AlessandroPomponio: Similar comment regarding the mode

orchestrator/modules/actuators/registry.py
```diff
@@ -444,10 +447,94 @@ def experimentForReference(
...
+                if candidates:
+                    available_versions = ", ".join(
+                        sorted({e.version for e in candidates if e.version})
+                    )
+                    message = f"{message}Available versions: {available_versions}."
+            raise UnknownExperimentError(message)
 
         return experiment
 
+    def resolve_reference(
```
AlessandroPomponio: This is more of a lookup than a resolution

orchestrator/schema/experiment.py
```diff
@@ -113,6 +114,73 @@ class Experiment(pydantic.BaseModel):
...
+            default=None,
+            description=(
+                "Algorithm version for this experiment following strict SemVer "
+                "(MAJOR.MINOR.PATCH). MAJOR identifies the memoisation boundary: "
+                "results from different major versions are never reused. MINOR covers "
+                "backward-compatible extensions (new outputs/inputs). PATCH covers "
+                "bug fixes and refactoring that do not change observable outputs."
+            ),
+        ),
+    ]
```
AlessandroPomponio:
```suggestion
    version: Annotated[
        StrictSemVerStr | None,
        pydantic.Field(
            description=(
                "Algorithm version for this experiment following strict SemVer "
                "(MAJOR.MINOR.PATCH). MAJOR identifies the memoisation boundary: "
                "results from different major versions are never reused. MINOR covers "
                "backward-compatible extensions (new outputs/inputs). PATCH covers "
                "bug fixes and refactoring that do not change observable outputs."
            ),
        ),
    ] = None
```

orchestrator/schema/experiment.py
```diff
@@ -113,6 +114,73 @@ class Experiment(pydantic.BaseModel):
...
+
+    @property
+    def semantic_identifier(self) -> str:
```
AlessandroPomponio: Given the name, I'd assume this was returning the whole semver. I would call this `major_version_identifer`

orchestrator/schema/experiment.py
```diff
@@ -833,27 +906,36 @@ class ParameterizedExperiment(Experiment):
...
-        Different parameterized versions of an experiment have a different parameterized identifier.
-        Their experimentIdentifier field, which identifies the base experiment, will be the same.
+        """Return the semantic parameterized identifier.
+
+        Uses the base experiment's :attr:`~Experiment.semantic_identifier`
+        as the prefix so that the memoisation key encodes the major algorithm
+        version.  For example: ``'solve_mip@v1-time_limit_s.3600'``.
+
+        Returns:
+            Semantic parameterized identifier string.
```
AlessandroPomponio: I'd say this is a major version parameterized identifier

orchestrator/schema/experiment.py
```diff
@@ -973,11 +1055,17 @@ def validate_parameterization(self) -> "ParameterizedExperiment":
 
     @property
     def reference(self) -> ExperimentReference:
+        """Return an ExperimentReference for this parameterized experiment.
 
+        Returns:
+            An ExperimentReference carrying the parameterization and
+            algorithm version.
```
AlessandroPomponio, line 1106: Algorithm version is just confusing

orchestrator/schema/reference.py
```diff
@@ -21,6 +33,95 @@ def reference_string_from_fields(
     return f"{actuator_identifier}.{experiment_identifier}"
 
 
+def _coerce_parameter_value(value_str: str) -> typing.Any:  # noqa: ANN401
```
AlessandroPomponio, line 36: Pydantic coerces values by default https://pydantic.dev/docs/validation/latest/concepts/strict_mode/

orchestrator/schema/reference.py
```diff
@@ -21,6 +33,95 @@ def reference_string_from_fields(
...
+    except ValueError:
+        pass
+    try:
+        return float(value_str)
+    except ValueError:
+        pass
+    return value_str
+
+
+def _parameterization_from_suffix(
```
AlessandroPomponio, line 133: Didn't we say this shouldn't be possible? Especially since after a certain length we are hashing the experiment identifiers?

orchestrator/schema/reference.py
```diff
@@ -40,20 +141,90 @@ class ExperimentReference(pydantic.BaseModel):
...
+        pydantic.Field(
+            default=None,
+            description=(
+                "Algorithm version of the referenced experiment (strict SemVer "
+                "MAJOR.MINOR.PATCH). When set, memoisation keys and equality checks "
+                "encode only the MAJOR component so that minor/patch changes do not "
+                "invalidate cached results."
+            ),
+        ),
+    ]
```
AlessandroPomponio:
```suggestion
    experimentVersion: Annotated[
        StrictSemVerStr | None,
        pydantic.Field(
            description=(
                "Algorithm version of the referenced experiment (strict SemVer "
                "MAJOR.MINOR.PATCH). When set, memoisation keys and equality checks "
                "encode only the MAJOR component so that minor/patch changes do not "
                "invalidate cached results."
            ),
        ),
    ] = None
```

orchestrator/schema/reference.py
```diff
@@ -40,20 +141,90 @@ class ExperimentReference(pydantic.BaseModel):
...
+        """Reject version suffixes embedded in experimentIdentifier."""
+        if "@" in value:
+            raise ValueError(
+                "experimentIdentifier must not contain '@'. "
+                "Use experimentVersion for algorithm versioning."
+            )
+        return value
+
+    @property
+    def semantic_experiment_identifier(self) -> str:
```
AlessandroPomponio: Again, I disagree this is the semantic identifier

AlessandroPomponio (changes requested):
I think the naming here is extremely confusing.

orchestrator/cli/resources/discovery_space/create.py
michael-johnston: Wanted the same exception for Operators as well and calling (Experiment+Operation)= Algorithms. 

But we can have separate ones

orchestrator/modules/actuators/catalog.py
michael-johnston: ok

orchestrator/schema/reference.py
michael-johnston: c.f. the discussion on slack. If we want to support the way currently str(reference) is used in the db we need it. All the string parsing code added is to fix the storage problem. 

Also I believe we hash the entity identifiers not experiment identifiers.

orchestrator/schema/reference.py
michael-johnston: Note I'm using term semantic (experiment) identifier not semantic version

semantic identifier means "an identifier which has a meaning", "semantic version" a version which has meaning beyond numbers. 

The semantic experiment identifier has a particular meaning - two experiments with same semantic experiment identifier do the same measurement.  

Whatever we call it is a semantic identifier is an identifier with a particular semantics - just as a semantic code versioning is a numbering scheme that has semantics.

orchestrator/modules/actuators/catalog.py
michael-johnston: This method resolves the reference - meaning it returns parameterized experiment instances if necessary.

orchestrator/modules/actuators/registry.py
michael-johnston: Its a resolution due to conversion to ParameterizedExperiment when necessary

orchestrator/modules/actuators/registry.py
michael-johnston: We need to support three matching types (base, semantic (major version), and fully qualified) - each has identifiers. For both Experiment and ParameterizedExperiment. 

These then are going to have different comparison methods that are going to be called something similar but do different matching. We will have different ways of searching. We can do it via different methods or parameters on one method. 

resolve_reference is different - its converting the reference into the Experiment/ParameterizedExperiment. 

Open to suggestions

orchestrator/schema/reference.py
michael-johnston: This does not work as the type of the field is a union including "str". Pydantic 2 will not coerce a type that is already an accepted type of the union (in pydantic 1 it would traverse in order of field types).

michael-johnston:
@VassilisVassiliadis @mgazz check the PR description and see if the changes make sense to you.

mgazz:
~Can I specify the version using the `Major Version Identifiers`?~

```
experiments:
    experimentVersion: v1
```

~Will the FQI use latest semver version matching the major version?~

I understand that experimentVersion does not support Major Version Identifiers (e.g., specifying only v1). Moreover, ado would resolve and execute all experiment versions matching the specified major version. I recommend making this behavior explicit in the documentation, with a dedicated section covering edge cases and expected resolution rules.

For instance the example reported above

> request to run "myexperiment@1.1.1"
actuator catalog as "myexperiment@1.3.1"
actuator uses "myexperiment@1.3.1". (Same major version identifier, same measurement)

This is reasonable given that 1.3.1 is a superset of 1.1.1 so the experiment would provide what was specified in 1.1.1 and more. 

A key question is how to handle the inverse case:

* request to run "myexperiment@1.3.1"
* actuator catalog as "myexperiment@1.1.1"

What is the expected behaviour? In this scenario, the catalog version is older than the requested one, meaning the experiment may run with a subset of the expected inputs/outputs. I would allow only forward compatible resolution on major version at runtime.

michael-johnston:
> What is the expected behaviour? In this scenario, the catalog version is older than the requested one, meaning the experiment may run with a subset of the expected inputs/outputs.

This would happen indeed with current implementation

>  I would allow only forward compatible resolution on major version at runtime.

We can do this

orchestrator/cli/resources/experiment/describe.py
```diff
@@ -34,14 +41,61 @@ def describe_experiment(parameters: AdoDescribeCommandParameters) -> None:
...
+            f"{HINT}Did you mean one of {error.actuators_with_experiments}?"
+            if len(error.actuators_with_experiments) > 1
+            else f"{HINT}Did you mean {error.actuators_with_experiments.pop()}?"
+        )
+        console_print(
+            f"{ERROR}Requested actuator {parameters.actuator_id} does not match "
+            f"experiment {parameters.resource_id}\n{hint_text}",
+            stderr=True,
+        )
+        raise typer.Exit(1) from error
```
AlessandroPomponio: This exception can only be raised in `_ado_get_actuator_from_experiment_id` I think and it's already handled there

orchestrator/cli/resources/experiment/describe.py
```diff
@@ -34,14 +41,61 @@ def describe_experiment(parameters: AdoDescribeCommandParameters) -> None:
...
+    except ValueError as error:
+        console_print(f"{ERROR}{error}", stderr=True)
+        raise typer.Exit(1) from error
+
+    if reference.experimentVersion is not None:
+        experiment = registry.experimentForReference(
+            reference, match_on="fully_qualified_version"
         )
```
AlessandroPomponio, line 69: this method actually raises exceptions that aren't handled in describe.py
It's also missing resolve = True so it could return a None

orchestrator/cli/resources/experiment/describe.py
```diff
@@ -34,14 +41,61 @@ def describe_experiment(parameters: AdoDescribeCommandParameters) -> None:
...
+        if len(matches) == 0:
+            raise UnknownExperimentError(
+                f"The {reference.actuatorIdentifier} actuator was found but it did not "
+                f"contain the {reference.experimentIdentifier} experiment."
+            )
+        if len(matches) > 1:
+            available_versions = ", ".join(
+                sorted({e.version for e in matches if e.version is not None})
+            )
+            raise UnknownExperimentError(
```
AlessandroPomponio: I'd have a specific Error for this as it's a different semantic - maybe AmbiguousExperimentIdentifierError

orchestrator/modules/actuators/catalog.py
```diff
@@ -102,36 +108,151 @@ def identifier(self) -> str:
...
-        In particular if the ExperimentReference is parameterized it does not affect the matching
+        Returns:
+            A list of major version identifiers.
+        """
+        return [e.major_version_identifier for e in self.experiments]
+
+    def experimentForReference(
+        self,
+        reference: ExperimentReference,
+        *,
```
AlessandroPomponio, line 127: Why do we need this?

orchestrator/modules/actuators/catalog.py
```diff
@@ -102,36 +108,151 @@ def identifier(self) -> str:
...
+        match_on: Literal["major_version", "fully_qualified_version"] = "major_version",
+        resolve: bool = False,
+    ) -> Experiment | ParameterizedExperiment | None:
+        """Return the experiment matching reference.
+
+        Matching compares on actuator and major version experiment identifier.
+        When ``match_on='fully_qualified_version'``, the exact version must also
+        match. When ``resolve=True``, deprecated experiments raise and references
```
AlessandroPomponio: deprecated experiments raise an error

orchestrator/modules/actuators/registry.py
```diff
@@ -377,76 +378,129 @@ def experimentForReference(
...
+                    f"Checking registry for the catalog of actuator {reference.actuatorIdentifier}"
                 )
-            else:
-                log.warning(
-                    f"No catalog registered for actuator {reference.actuatorIdentifier}"
+                actuator_catalog = self.catalogForActuatorIdentifier(
+                    actuatorid=reference.actuatorIdentifier
                 )
+                catalogs_to_try.append(actuator_catalog)
+            except KeyError:
```
AlessandroPomponio: can key error even come from here?

orchestrator/modules/actuators/registry.py
AlessandroPomponio, line 440: This method is really tough to follow

website/docs/actuators/creating-actuator-classes.md
```diff
@@ -167,6 +167,90 @@ From the `submit` callers point of view this means:
...
+
+> [!NOTE]
+> Experiments without a declared version emit a `DeprecationWarning` at
+> registration time.
+
+<!-- markdownlint-enable no-blanks-blockquote -->
+
+### Resolving experiment references in `submit`
+
+In your `submit` method, use `catalog.resolve_reference(reference)` to obtain
```
AlessandroPomponio: This method was renamed to `catalog.experimentForReference(reference, resolve=True)`

orchestrator/modules/actuators/custom_experiments.py
```diff
@@ -768,21 +774,19 @@ def submit(
             f"Received a request to measure {experimentReference} on {[e.identifier for e in entities]}"
         )
 
-        if self._catalog.experimentForReference(experimentReference) is None:
+        try:
+            targetExperiment = self._catalog.experimentForReference(
+                experimentReference, resolve=True
+            )
+        except Exception as error:
```
AlessandroPomponio: According to Bob:

> Broad except Exception swallows ExperimentVersionMismatchError / DeprecatedExperimentError — Specific, user-actionable error messages are replaced with a generic "not in catalog" ValueError.

orchestrator/schema/experiment.py
```diff
@@ -113,6 +114,72 @@ class Experiment(pydantic.BaseModel):
...
+        Returns:
+            True if both experiments have the same actuator and major version
+            identifier.
+        """
+        if not isinstance(other, Experiment):
+            return False
+        return (
+            self.actuatorIdentifier == other.actuatorIdentifier
+            and self.major_version_identifier == other.major_version_identifier
+        )
```
AlessandroPomponio, line 185: According to Bob:

> Experiment.__eq__ is asymmetric with ParameterizedExperiment.__eq__ — base_exp == param_exp can be True while param_exp == base_exp is False, violating the equality contract and causing unexpected behaviour in sets/dicts.

orchestrator/schema/reference.py
```diff
@@ -21,6 +33,95 @@ def reference_string_from_fields(
...
+        )
+        if (
+            parameterization_suffix
+            and "." in _PARAMETERIZATION_SEGMENT_SPLIT.split(parameterization_suffix)[0]
+        ):
+            return (
+                base_identifier,
+                None,
+                _parameterization_from_suffix(parameterization_suffix),
+            )
```
AlessandroPomponio: According to Bob:

> Hyphenated identifiers without valid parameterization are silently treated as plain identifiers — No warning is emitted; malformed input only surfaces later as a confusing UnknownExperimentError.

orchestrator/modules/actuators/registry.py
```diff
@@ -377,76 +378,129 @@ def experimentForReference(
...
+            message = (
+                f"The {reference.actuatorIdentifier} actuator was found but a match to "
+                f"{reference} was not found using mode {match_on}."
+            )
+            if actuator_catalog is not None:
+                candidates = actuator_catalog.experiments_matching_identifier(reference)
+                if candidates:
+                    available_versions = ", ".join(
+                        sorted({e.version for e in candidates if e.version})
+                    )
```
AlessandroPomponio, line 548: According to Bob: 
> Version hints in error messages use lexicographic sorted() — ["2.0.0", "10.0.0"] sorts as ["10.0.0", "2.0.0"] instead of numerically. Same issue at line 499-501.

orchestrator/modules/actuators/registry.py
```diff
@@ -377,76 +378,129 @@ def experimentForReference(
...
+                    message = f"{message} Available versions in catalog: {available_versions}."
+            raise UnknownExperimentError(message)
+
+        if not self.actuatorForIdentifier(reference.actuatorIdentifier):
+            raise UnknownActuatorError(reference.actuatorIdentifier)
```
AlessandroPomponio: According to Bob:

> actuatorForIdentifier raises rather than returns falsy — if not self.actuatorForIdentifier(...) will never be True; the guard is dead code and the method will raise UnknownActuatorError unconditionally, masking the intended UnknownExperimentError.

orchestrator/schema/reference.py
```diff
@@ -1,16 +1,28 @@
...
+    ConstitutivePropertyDescriptor,
+)
 from orchestrator.schema.property_value import (
     ConstitutivePropertyValue,
 )
+from orchestrator.utilities.pydantic import StrictSemVerStr, semver_major
+
+_FQ_VERSION_WITH_PARAMS_PATTERN = re.compile(
+    r"^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-(.*))?$"
+)
```
AlessandroPomponio, line 22: According to Bob:

> _FQ_VERSION_WITH_PARAMS_PATTERN duplicates the core SemVer pattern from pydantic.py — Two independent sources of truth must be kept in sync.

orchestrator/utilities/pydantic.py
```diff
@@ -119,3 +119,44 @@ def validate_pep440_version(value: str) -> str:
...
+def semver_major(version: str) -> int:
+    """Extract the MAJOR component from a strict SemVer string.
+
+    Args:
+        version: A strict SemVer string (e.g. ``"1.2.3"``).
+
+    Returns:
+        The integer major version component.
+    """
+    return int(version.split(".")[0])
```
AlessandroPomponio: According to Bob:

> semver_major silently trusts callers to pass validated strings — No input guard; unvalidated strings produce context-free ValueErrors.

AlessandroPomponio (dismissed):

orchestrator/cli/resources/experiment/describe.py
michael-johnston: resolve=True will raise DeprecationError which would mean you can't describe the deprecated experiment. 

Suggestions?

orchestrator/cli/resources/experiment/describe.py
michael-johnston: ok

orchestrator/modules/actuators/catalog.py
michael-johnston: It's a new python syntax I learned - it means you need to specify match_on and resolve as keyword args e.g. you can't do experimentForReference(reference "major_version", False)

orchestrator/modules/actuators/catalog.py
michael-johnston: Do you want to change "raise" -> "raise an error" ??

orchestrator/modules/actuators/custom_experiments.py
michael-johnston: Can scope it (you can't have DeprecantedExperimentError from custom experiments btw).

orchestrator/modules/actuators/registry.py
michael-johnston: You'll have to be more specific

orchestrator/modules/actuators/registry.py
michael-johnston: I don't think so - will fix

orchestrator/schema/experiment.py
michael-johnston: Good catch

orchestrator/schema/reference.py
michael-johnston: Is Bob right?

orchestrator/schema/reference.py
michael-johnston: What's the suggestion?

VassilisVassiliadis:
I get what this is proposing, why this is the approach the PR is going for, and I do think that this is an improvement over what we currently have.

Still, I think that the last 2 rules are inconsistent.

1. Memoization uses Major version to decide a cache hit/miss
2. Catalog uses Major version to decide which experiment to launch
3. ExperimentSpace must contain the entire semver of the experiment

Unfortunately, I cannot think of a reasonable way to remove the source of inconsistency at the code level. One way would for the DiscoverySpace to support defining the version as either just the major version OR a full semver but that could make validation complex.

So, I think we'll just go with the current design and deal with what I perceive as "inconsistency" with appropriate documentation. At the end of the day, we'll keep using ado so if we find that this is confusing we can just improve it down the line.

orchestrator/cli/resources/experiment/describe.py
michael-johnston: You mean ActuatorDoesNotHaveExperimentError?

AlessandroPomponio: I believe so

orchestrator/cli/resources/experiment/describe.py
AlessandroPomponio: Ah, I see. I wonder should the two things be coupled? Resolving and handling deprecated experiments?

orchestrator/modules/actuators/catalog.py
AlessandroPomponio: Yes, I think that's what I meant, sorry

orchestrator/schema/reference.py
AlessandroPomponio: To an extent yes, in the sense that the "core SemVer pattern" is the same but of course this one additionally supports a hyphenated suffix. I guess it's fine

orchestrator/schema/reference.py
AlessandroPomponio: I'm guessing something should be done in case the if condition in L 112 is false
The point is that if L108 is true and L112 is false, then nothing happens to experiment_part and down the line this ends up raising an UnknownExperimentError

orchestrator/modules/actuators/registry.py
AlessandroPomponio: it's very nested and has many try/except/else blocks that are then followed by other ifs statements, etc. It gets overwhelming and I can't really keep track of what's going on

orchestrator/cli/resources/experiment/describe.py
michael-johnston: Up to know they both happen at same time always.

orchestrator/modules/actuators/registry.py
michael-johnston: It's a bit convoluted to sort them numerically - not sure it is worth it

michael-johnston:
@AlessandroPomponio @mgazz @VassilisVassiliadis 

Suggestion for `describe space` output

Screenshot 2026-06-23 at 18 25 44

Differences
- Experiments table lists the base identifier, required major version, parameterization separately (idea is to clearly state this space requires a give major version of the experiment)
- Experiment details section - Uses the major version identifier (instead of FQI)
- Add text that the details on inputs/outputs are the  "Expected interface" and what version this is based from  i.e. this is the version that the space was created with.

michael-johnston:
@mgazz 
>  I would allow only forward compatible resolution on major version at runtime.

- Does patch version matter e.g. 1.2.1 is in registry and space is 1.2.2?
- If the space experiment is 1.3 and registry is 1.2 but the experiment interface requires is identical i.e. all input parameters can be passed, all output metrics are present (maybe there was some optimization), what would you expect?

I'm trying to choose between
- Any minor version of same major version (current sate)
- only forward compatible resolution on minor+patch version at runtime (minor same or greater, if minor same patch same or greater)
- only forward compatible resolution on major version at runtime - not including patch (minor same or greater)
- OR interface compatibility - any minor version is allows if it provides same interface as expected. 

I think the last makes most sense.

VassilisVassiliadis:
What is the full version of the experiment in the screenshot you pasted?

michael-johnston:
> What is the full version of the experiment in the screenshot you pasted?

The version of the experiment used to create the space, and hence define the required interface is on line 

"Expected Interface (from v1.0.0)"

VassilisVassiliadis:
Thanks, wanted to make sure that it is indeed 1.0.0

VassilisVassiliadis:
I prefer "Any minor version of same major version PROVIDED that the interfaces match".

mgazz:
> I prefer "Any minor version of same major version PROVIDED that the interfaces match".

Same

danielelotito:
@michael-johnston 
I'd go with the 4th one https://github.com/IBM/ado/pull/1043#issuecomment-4783066247 .
But I'd like a command (Idk what's the recommended one atm) that tells me which spaces in the sample store have a measurement space compatible with the new experiment (i.e. those created with same major version)

mgazz:
@danielelotito 

> But I'd like a command (Idk what's the recommended one atm) that tells me which spaces in the sample store have a measurement space compatible with the new experiment (i.e. those created with same major version)

Following option 4 logic, if the goal is to leverage memoization, wouldn't the command be more useful if it matches a minor version and same interface?

danielelotito:
@mgazz Yes, ideally the command could accept some flags to obtain the behavior you described.

However, @michael-johnston I am also considering an alternative option. 

"Any minor version of same major version PROVIDED that the interfaces of the most older experiments match the input properties and include the target property"


The difference is that Input and targets are treated differently. We require exact input matching, while older version might have other target properties.

I reasoned along this line. I start with exp version v0.1.0

A minor version change could change interface (e.g. add a new target property). -> v0.2.0
Then we realize that the measured property is computationally expensive to obtain and almost useless. 
We try to make the code more performant v0.2.1,
We are not satisfied, so do a minor version change again in  v0.3.0 the expensive property is removed.

I'd like v0.2 results retrieved by v0.3 experiment

mgazz:
> Any minor version of same major version PROVIDED that the interfaces of the most older experiments match the input properties and include the target property

That's fair.

michael-johnston:
> Any minor version of same major version PROVIDED that the interfaces of the most older experiments match the input properties and include the target property

I was actually thinking this for both inputs and outputs i.e. satisfying the interface given in the space (contract) means I can call you as I expect and will get back what I expected

Explicitly

Input

- The experiment must take all required inputs and parameterized inputs specified in space (name and domain)
- All optional parameters (not overridden in spacE) must be present, their default values must be the same
- The default value of any new optional parameter must lead to the same behaviour as before it was added **(This is the one that is different from what @danielelotito requested)**

This means the experiment can't 
- have new required inputs 
- have reduced the domain of any inputs 
- have removed any parameter (required, user parameterized or that was assumed to take a default value)
- have changed the default value of any optional parameter not parameterized 

It can:
- have added new optional parameters. Their default values are assumed to give the same behaviour as before they were added e.g. I added a parameter "warm_start" to solve_mip  that defaults to None - I expect a space cretated before warm_start was added (a hence doesn't use) can use the new experiment version

Output
- The experiment must return all expected outputs
- It can add new outputs

michael-johnston:
> A minor version change could change interface (e.g. add a new target property). -> v0.2.0
Then we realize that the measured property is computationally expensive to obtain and almost useless.
We try to make the code more performant v0.2.1,
We are not satisfied, so do a minor version change again in v0.3.0 the expensive property is removed.

> I'd like v0.2 results retrieved by v0.3 experiment

Yes this would work. If you create a space with v0.3, execute with v0.3, then v0.2 results would be memoized. 

If you create a space with v0.3, and try to execute with v0.2, it will work.  (v0.2 returns superset of v0.3 reqs)

If you have an existing space created with v0.2 and try to execute with v0.3, it will fail (v0.3 can't measure the expensive useless property required by the existing space)

michael-johnston:
Keep in mind there are two places where we have to make a decision

- When starting the operation: Does the experiment version in the catalog satisfy what the space requires (need to consider inputs and outputs)

- When memoizing: Does the stored results for an experiment (with same major version), include all the outputs required by the space

danielelotito:
I sensed that the default values match would have required a deeper discussion.
I'd like that a change in defaults value does not prevent memorization.
What are you trying to prevent with the interface check?
Why you think it's better than option 1(current behaviour)?

michael-johnston:
> What are you trying to prevent with the interface check?

I want to avoid the case where someone describes a space with  v1.2 of an experiment that is completely valid for v1.1, yet can't be executed by v1.1

I want to avoid the case where default value of a parameter is changed between v1.1. and v1.2 which implicitly changes what the experiment does, but then a space created with v1.1 is measured using v1.2 

I want to avoid the case where from v1.1 to v1.2 a metric is removed, but then v1.2 is used to execute on the space

In the same situation, I want to avoid the case where from v1.1 is used to execute on a space, with v1.2 results in the store, but v1.2 results are memoized

> Why you think it's better than option 1(current behaviour)?

The current behaviour for selecting if an experiment is compatible for executing an explore operation on a space/memoization - only relying on minor version >= - would allow all above.

michael-johnston:
basically the proposal is standard polymorphism/duck-typing - if it looks like a duck and quacks like a duck, and has same major version - its a duck.

danielelotito:
I'm still thinking about this, especially about requiring defaults and domain to match.


> I want to avoid the case where default value of a parameter is changed between v1.1. and v1.2 which implicitly changes what the experiment does, but then a space created with v1.1 is measured using v1.2

Are spaces and measurements always materialized with their defaults explicit?
If so (which I think would be better) I do not see why this is a case we want to avoid.

I imagine you want that two space yaml generate the same experiment behavior in "compatible" versions of the experiment, this happen if you change the default.
However, I see the yaml as just a way to load an object in the system, and it is okay for two yaml to create objects that differs for the value of the changed default.

The "looks like a duck and quacks like a duck" could be applied at the level of the created objects.
This implies a less strict interface match than the one I have previously proposed.

The thing I want to avoid is a user losing (an easy access to) memoization just because he realized that he wanted a bigger domain.

michael-johnston:
> I'm still thinking about this, especially about requiring defaults and domain to match.

The domain doesn't have to match it just has to be same or bigger

> The thing I want to avoid is a user losing (an easy access to) memoization just because he realized that he wanted a bigger domain.

c.f. above. A bigger domain is fine. A smaller domain means, potentially, a value in the entity space is no longer valid for the experiment.  

> Are spaces and measurements always materialized with their defaults explicit?

The defaults values of the experiment are there (because we store the full experiment YAML with the space) but are not carried with the reference (which is what's passed to the actuator). This is because it's assumed the defaults will be the same. 


We could change this but I would do it separately. If we did then we wouldn't care about defaults changing.

orchestrator/api/dependencies/validation.py
```diff
@@ -1,17 +1,56 @@
...
+        HTTPException: Mapped from registry or catalog lookup errors.
+    """
+    if isinstance(error, UnknownExperimentError):
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"Actuator {actuator_id} does not have experiment {experiment_id}",
+        ) from error
+    if isinstance(error, UnexpectedCatalogRetrievalError):
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
```
AlessandroPomponio: It should be a [500 Internal Server Error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/500) given the semantics in https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#server_error_responses

orchestrator/cli/resources/experiment/describe.py
```diff
@@ -14,8 +14,18 @@
...
+    UnknownExperimentError,
+)
+from orchestrator.schema.reference import (
+    ExperimentReference,
+    _parse_experiment_part_from_string,
+)
+
+
+class AmbiguousExperimentIdentifierError(Exception):
+    """There are multiple matches for the given identifier in the catalog"""
```
AlessandroPomponio: I think this belongs to orchestrator/modules/actuators/registry.py (where we already have e.g., UnknownExperimentError)

orchestrator/modules/actuators/catalog.py
```diff
@@ -102,36 +108,152 @@ def identifier(self) -> str:
...
+
+
+class ExperimentVersionMismatchError(Exception):
+    """Raised when the version of a resolved experiment does not match the reference.
+
+    This error is only raised when :meth:`ExperimentCatalog.experimentForReference` is
+    called with ``resolve=True``, ``match_on='fully_qualified_version'``, and the
+    exact version in the catalog differs from the version recorded on the
+    :class:`~orchestrator.schema.reference.ExperimentReference`.
+    """
```
AlessandroPomponio: Maybe actually we should have a file with experiment-related errors, because we seem to have them scattered around

orchestrator/modules/actuators/registry.py
```diff
@@ -48,6 +54,76 @@ class UnexpectedCatalogRetrievalError(Exception):
...
+        expected_experiment: The measurement-space experiment being validated.
+        issue: Structured interface mismatch returned by compatibility checking.
+
+    Returns:
+        A user-facing issue string for logging or error reporting.
+    """
+    template = _MEASUREMENT_SPACE_INTERFACE_ISSUE_TEMPLATES[issue.kind]
+    detail = template.format(**issue.model_dump(exclude={"kind"}, exclude_none=True))
+    return (
+        f"ExperimentInterfaceMismatchError: "
```
AlessandroPomponio, line 110: We don't have any class called `ExperimentInterfaceMismatchError` btw

orchestrator/schema/experiment.py
```diff
@@ -998,6 +1091,266 @@ def valueForOptionalProperty(
...
+
+
+def _parameterized_optional_identifiers(experiment: "Experiment") -> set[str]:
+    """Return optional property identifiers with custom parameterization."""
+    if isinstance(experiment, ParameterizedExperiment):
+        return {value.property.identifier for value in experiment.parameterization}
+    return set()
+
+
+def check_experiment_interface_compatible(
```
AlessandroPomponio: Should these maybe be instance methods of Experiment?

I would at that point call `check_experiment_interface_compatible` `has_compatible_interface_with`

orchestrator/schema/experiment.py
```diff
@@ -998,6 +1091,266 @@ def valueForOptionalProperty(
...
+        ExperimentInterfaceIssue(
+            kind=ExperimentInterfaceIssueKind.MISSING_REQUIRED_CONSTITUTIVE_IN_PROVIDED,
+            propertyIdentifier=property_identifier,
+        )
+        for property_identifier in expected_required_constitutive_ids
+        if _constitutive_property_by_identifier(
+            provided_experiment, property_identifier
+        )
+        is None
+    )
```
AlessandroPomponio: All these things look very odd, although I'm not sure what to suggest

orchestrator/schema/measurementspace.py
```diff
@@ -185,13 +192,30 @@ def __rich__(self) -> "RenderableType":
...
+                (
+                    None
+                    if isinstance(e, Experiment)
+                    else identifier_for_parameterized_experiment(
+                        "", e.parameterization
+                    )[1:]
+                ),
+            ]
+            for e in self.experiments
+        ]
```
AlessandroPomponio, line 217: This is tough to understand

michael-johnston: It just populating 3 fields per experiment for the table. 
- base identifier
- Major version if there is one
- Parameterization if any (as a string )

orchestrator/modules/actuators/registry.py
michael-johnston: It's just to match the format of the other messages checkMeasurementSpaceSupported returns.

orchestrator/schema/experiment.py
michael-johnston: Me neither

orchestrator/schema/experiment.py
michael-johnston: Maybe check_experiment_interface_compatible could be an instance method - although it doesn't have to be.

orchestrator/api/dependencies/validation.py
```diff
@@ -49,19 +90,22 @@ def validated_experiment_id(
 
     Raises:
         HTTPException: If the experiment is not associated with the actuator
-            (404 Not Found).
+            (404 Not Found), the catalog is unavailable (503 Service Unavailable),
```
AlessandroPomponio:
```suggestion
            (404 Not Found), the catalog is unavailable (500 Internal Server Error),
```

orchestrator/schema/experiment.py
```diff
@@ -687,6 +690,31 @@ def requiredObservedProperties(self) -> list[ObservedProperty]:
...
+    ) -> ConstitutiveProperty | None:
+        """Return a declared constitutive property by identifier, if present."""
+
+        v = [
+            p
+            for p in self.requiredConstitutiveProperties + list(self.optionalProperties)
+            if p.identifier == property_identifier
+        ]
+
+        return None if len(v) == 0 else v[0]
```
AlessandroPomponio:
```suggestion
        return next(
            (
                p
                for p in (*self.requiredConstitutiveProperties, *self.optionalProperties)
                if p.identifier == property_identifier
            ),
            None,
        )
```

orchestrator/schema/experiment.py
```diff
@@ -1161,128 +1185,106 @@ class ExperimentInterfaceIssue(pydantic.BaseModel):
...
-) -> ConstitutiveProperty | None:
-    """Return a constitutive property by identifier, if present."""
-    for prop in experiment.requiredConstitutiveProperties:
-        if prop.identifier == property_identifier:
-            return prop
-    for prop in experiment.optionalProperties:
-        if prop.identifier == property_identifier:
-            return prop
-    return None
+def _issues_for_ids(
```
AlessandroPomponio: I'd call this `_property_issues_of_kind` or something along the lines. Basically, I'd stress that this just returns issues of a specific kind

website/docs/actuators/creating-actuator-classes.md
AlessandroPomponio: This is still an issue

orchestrator/modules/actuators/registry.py
```diff
@@ -523,20 +632,44 @@ def updateCatalogs(
...
 
         issues = []
         for experiment in measurement_space.experiments:
+            ref = experiment.reference
             try:
-                self.experimentForReference(experiment.reference)
-            except UnknownExperimentError as error:  # noqa: PERF203
+                catalog = self.catalogForActuatorIdentifier(ref.actuatorIdentifier)
+                catalog.experimentForReference(ref, resolve=True)
+                provided_experiment = catalog.experimentForReference(ref, resolve=False)
```
AlessandroPomponio: I guess we can assign to the first result directly


```suggestion
                provided_experiment = catalog.experimentForReference(ref, resolve=True)
```

orchestrator/cli/resources/experiment/describe.py
```diff
@@ -34,14 +43,49 @@ def describe_experiment(parameters: AdoDescribeCommandParameters) -> None:
...
+        if len(matches) == 0:
+            raise UnknownExperimentError(
+                f"The {reference.actuatorIdentifier} actuator was found but it did not "
+                f"contain the {reference.experimentIdentifier} experiment."
+            )
+        if len(matches) > 1:
+            available_versions = ", ".join(
+                sorted({e.version for e in matches if e.version is not None})
+            )
+            raise AmbiguousExperimentIdentifierError(
```
AlessandroPomponio: We need to add handling for these two errors in describe.py

orchestrator/modules/actuators/custom_experiments.py
```diff
@@ -297,32 +297,37 @@ def custom_experiment(
...
+        metadata: Metadata for the experiment.
+        use_ray: If True the CustomExperiments actuator will launch the experiment as a ray remote task.
         ray_options: A dictionary containing ray remote task options.
-            The keys and allowed values are defined by RayRemoteOptions
+            The keys and allowed values are defined by RayRemoteOptions.
+        version: Algorithm version string in strict SemVer format (``"MAJOR.MINOR.PATCH"``).
+            Use this to declare the scientific identity of the experiment independently of the
+            Python package version.  The MAJOR component is encoded into memoisation keys so
+            that results from different major versions are never reused.
+            If omitted a :class:`DeprecationWarning` is emitted at registration time.
```
AlessandroPomponio: According to Bob: Both the docstring and documentation pages promise a DeprecationWarning when version=None, but no warnings.warn() call is present in the implementation.

orchestrator/cli/resources/experiment/describe.py
michael-johnston: this is describe.py? 

What does handle mean

to do like 
```
 console_print(f"{ERROR}{error}", stderr=True)
        raise typer.Exit(1) from error
```

orchestrator/modules/actuators/custom_experiments.py
michael-johnston: Removed

orchestrator/schema/experiment.py
michael-johnston: Change to _create_issues_of_kind_for_ids

orchestrator/cli/resources/experiment/describe.py
AlessandroPomponio: That could work - we could have a better message or something along the lines. The point of this suggestion is mainly to prevent the exception from going straight through with the traceback and all

website/docs/actuators/creating-actuator-classes.md
```diff
@@ -1,4 +1,5 @@
 <!-- markdownlint-disable first-line-h1 -->
+<!-- markdownlint-disable first-line-h1 -->
```
AlessandroPomponio, line 2: This is a duplicate of the previous line

orchestrator/modules/actuators/custom_experiments.py
```diff
@@ -314,10 +314,7 @@ def custom_experiment(
         ray_options: A dictionary containing ray remote task options.
             The keys and allowed values are defined by RayRemoteOptions.
         version: Algorithm version string in strict SemVer format (``"MAJOR.MINOR.PATCH"``).
-            Use this to declare the scientific identity of the experiment independently of the
-            Python package version.  The MAJOR component is encoded into memoisation keys so
-            that results from different major versions are never reused.
-            If omitted a :class:`DeprecationWarning` is emitted at registration time.
+            Can be None.
```
AlessandroPomponio, line 317: Was this change intended?

AlessandroPomponio (approved):
LGTM - two very minor things in case they weren't intentional

orchestrator/modules/actuators/custom_experiments.py
michael-johnston: Yes - not emitting a warning and the information about encoding of memoization key etc. is not for this place.

website/docs/actuators/creating-actuator-classes.md
michael-johnston: I'll remove it in future PR

PR #1045, opened 2026-06-17
refactor(sfttrainer_download_hf_weights): do not use ray

VassilisVassiliadis:
Since download_weights() no longer configures packages, there's no need to use ray at all.

This also addresses an issue where sfttrainer_download_hf_weights does not work when using ray.


danielelotito (approved):
Looks convincing, simple changes

PR #1069, opened 2026-06-23
build(hooks): add new pre-commit hooks

AlessandroPomponio:
### What changed

**New pre-commit hooks** (`pre-commit-hooks` v6.0.0) were added to `.pre-commit-config.yaml`:
- `check-yaml`, `check-json` — validate YAML/JSON syntax
- `check-merge-conflict` — catch unresolved merge markers
- `check-added-large-files` — prevent accidental large-file commits
- `debug-statements` — flag leftover debug calls
- `end-of-file-fixer`, `trailing-whitespace` — enforce consistent file formatting

**Autofix sweep** — running these hooks on the existing codebase produced a large formatting-only diff (trailing whitespace, missing final newlines) across ~87 files. No logic was changed.

**Docs / YAML fixes** that accompanied the sweep:
- The KubeRay service-account YAML was split into three separate single-resource files (as required by the `check-yaml` hook which expects one CRD per file).
- The `--unsafe` flag was added to `check-yaml` to allow the `!!python/name:` tags used in `mkdocs.yml`.


AlessandroPomponio:
Looks like the changes to the autogluon models, despite not changing the actual content, are enough to break autoconf cc @srikumar003 @danielelotito 

I can try to restore the files from main and hopefully add an exclusion for those subdirs but we should be aware of how fragile they are

christian-pinto (approved):

PR #1078, opened 2026-06-24
fix(vllm_performance): prevent TypeError when parsing renderer_num_workers

mgazz:
## Problem

A `TypeError` was occurring during remote execution in the vllm_performance actuator when processing the `renderer_num_workers` parameter:

```
TypeError: int() can't convert non-string with explicit base
```

**Error Location:** The error occurred at line 298 in `experiment_executor.py`:

```python
# BEFORE - caused TypeError
renderer_num_workers=(
    int(values.get("renderer_num_workers"))
    if int(values.get("renderer_num_workers"), 0) > 0  # ← typo: ", 0" in wrong position
    else None
),
```

**Root Cause:** A typo placed `, 0` as the second argument to `int()` instead of as the default argument to `values.get()`. When `int()` receives a second argument, it interprets it as a numeric base (for string conversion), so passing a non-string as the first argument causes the `TypeError`. The fix is simply moving `, 0` to the correct position:

```python
# WRONG: int(values.get("renderer_num_workers"), 0)  — passes 0 as base to int()
# RIGHT: int(values.get("renderer_num_workers", 0))   — passes 0 as default to .get()
```

## Solution

Define `renderer_num_workers` before the `create_test_environment` call with the correct default:

```python
renderer_num_workers = int(values.get("renderer_num_workers", 0))
create_test_environment(
    ...
    renderer_num_workers=renderer_num_workers,
    ...
)
```

This is safe because `_is_threadpool_requested()` (called just before) already validates that `renderer_num_workers`, if specified, is a non-negative number.

### DIFF
--- a/plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
+++ b/plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
@@ -265,6 +265,7 @@ def _create_environment(
                             f"Threadpool requested but not supported by image {image_name}"
                         )
 
+                    renderer_num_workers = int(values.get("renderer_num_workers", 0))
                     create_test_environment(
                         k8s_name=env.k8s_name,
                         model=model,
@@ -293,11 +294,7 @@ def _create_environment(
                         enforce_eager=values.get("enforce_eager", 0) == 1,
                         io_processor_plugin=values.get("io_processor_plugin"),
                         otlp_traces_endpoint=otlp_traces_endpoint,
-                        renderer_num_workers=(
-                            int(values.get("renderer_num_workers"))
-                            if int(values.get("renderer_num_workers"), 0) > 0
-                            else None
-                        ),
+                        renderer_num_workers=renderer_num_workers,
                         check_interval=check_interval,
                         timeout=timeout,
                     )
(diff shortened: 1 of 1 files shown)


AlessandroPomponio (changes requested):
The root cause of the issue is incorrect - the TypeError is raised because there's a typo in where the `, 0` is placed in L298.

It should have been 

```python
int(values.get("renderer_num_workers", 0))
```

Instead of 
```python
int(values.get("renderer_num_workers"), 0)
```

In L262 you're already calling `_is_threadpool_requested` which is already checking that if `renderer_num_workers` is specified, it's a number (by casting it to an int) and that the number is >= 0.

The only case that could cause errors in L298 is if `renderer_num_workers` hasn't been specified (i.e., it's `None`), which is exactly what's causing the issue here.

**TLDR;** for readability purposes I would define in L267 (before the call to `create_test_environment`) the variable:

```python
renderer_num_workers = int(values.get("renderer_num_workers", 0))
```

And use it in the assignment, instead of doing the other changes

michael-johnston:
@mgazz can you update the problem/solution in the PR description?

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
```diff
@@ -293,11 +294,7 @@ def _create_environment(
                         enforce_eager=values.get("enforce_eager", 0) == 1,
                         io_processor_plugin=values.get("io_processor_plugin"),
                         otlp_traces_endpoint=otlp_traces_endpoint,
-                        renderer_num_workers=(
-                            int(values.get("renderer_num_workers"))
-                            if int(values.get("renderer_num_workers"), 0) > 0
-                            else None
-                        ),
+                        renderer_num_workers=renderer_num_workers,
```
AlessandroPomponio: I think this is now incorrect - the if should have been retained but simplified to be 
`renderer_num_workers if renderer_num_workers > 0 else None`

mgazz:
@michael-johnston @AlessandroPomponio updated as requested

AlessandroPomponio (approved):
LGTM thanks

PR #1095, opened 2026-06-26
feat(core): add ado provenance

michael-johnston:
Also support adding provenance to script operators


orchestrator/core/metadata.py
```diff
@@ -139,29 +142,79 @@ def from_module_conf(cls, module_conf: object) -> Self | None:
...
+
+        Args:
+            **fields: Provenance field values for this model or subclass.
+
+        Returns:
+            A provenance instance with ``ado`` populated when available.
+        """
+        if fields.get("ado") is None:
+            fields["ado"] = cls.resolve_ado_provenance()
+        return cls(**fields)
```
AlessandroPomponio: I find it a bit weird for this to be its own class method. In conjunction with the fact that `ado`'s provenance can be `None`, I believe this would lead to instances created without provenance.

I would:

- Declare `ado` to be of type `Defaultable[PackageProvenance]` and have the default be `PackageProvenance.from_distribution_name("ado-core")`
- Remove the `resolve_ado_provenance` method (it's a wrapper for the call above)
- Remove the `at_creation` method as it won't be needed anymore
- Remove `with_ado_core_provenance` as long as the `ADOResources` also have `ado` as `Defaultable[PackageProvenance]`

AlessandroPomponio (dismissed):
I think we should have a simpler approach

orchestrator/core/metadata.py
michael-johnston: The problem is the provenance field of ADOResource defaults to ProvenanceInfo, which before this is empty dict. This allows resource without provenance to be loaded from db - they have empty provenance. 

If we do above, so ado field is populated automatically in ProvenanceInfo then all resource loaded from db will get the current ado version recorded in their provenance section which isn't correct.

michael-johnston: There reason for the class method is to handle this - it distinguishes between actual first time creation and loading an serialisation

michael-johnston: The only other option I can see is to use a validation context, that turns off populating auto-populating ado field. Then use this when loading resources.

AlessandroPomponio: I see what you mean - but this way we will always have this legacy support for pre-provenance resources and suboptimal semantics because of it. I'd avoid this, especially given that we're going towards version 2.0.0

We could:

- Set a version that's clearly unknown (e.g., 0.0.0.post1)
- Create a function that guesstimates based on created date and assigns it the ado version that was stable at that time

michael-johnston: We can't do either of this though as fundamentally we can't now why the ado version is None i.e. is it because we are in latest ado creating the model , or are we loading a pre-provenance resource. 

Hence we can't know given its none what to do without some external signal e.g. validation context, we use from_creation method, we add a from_stored method or something.

AlessandroPomponio: Ah, I see, you're right... At that point I would suggest the validation context + custom validator that we can remove down the line in - say - v3
Just to simplify semantics. Plus, given that it would be a context that we have when loading from db, we might have other instances where this can be useful

michael-johnston:
@AlessandroPomponio Update to use a validation context. 

We need to two cases 
- none provenance: field  (oldest resource)
- none provenance.ado field (resource with plugin provenance but not ado provenance)

AlessandroPomponio:
> * none provenance.ado field (resource with plugin provenance but not ado provenance)

This only affects people that used ado from main in the last 3 weeks, right?

Btw I see there are merge conflicts

michael-johnston:
> This only affects people that used ado from main in the last 3 weeks, right?

yes

AlessandroPomponio (approved):

PR #1096, opened 2026-06-26
refactor(cli)!: remove legacy migrator system in ado upgrade

### LINKED ISSUE #1094: refactor(cli)!: remove legacy migrator system in ado upgrade

AlessandroPomponio:
## Remove legacy migrator system from `ado upgrade`

This is a breaking change that removes the legacy migrator subsystem that was previously used to handle schema migrations for old/deprecated resource formats.

The current state and a few additional migrators will be temporarily kept in the `ap_20260626_pre_removal_migrators` branch

### What changed

**Deleted (~3,800 lines removed across 30 files):**
- All migrator implementations under `orchestrator/core/legacy/` — including migrators for `samplestore`, `discoveryspace`, `operation`, `actuatorconfiguration`, and `resource` resource types
- The `LegacyMigratorRegistry` and all supporting infrastructure
- CLI utilities for listing/applying legacy migrators (`orchestrator/cli/utils/legacy/`)
- All associated tests (`tests/core/test_legacy_migrators.py`, `test_legacy_registry.py`, `test_migrator_dependencies.py`, `test_upgrade_transaction_safety.py`, and related fixtures)

**Simplified (`ado upgrade` command):**
- Removed `--apply-legacy-migrator` and `--list-legacy-migrators` CLI flags
- The upgrade handler is now a straightforward load-and-resave loop with no legacy branching logic

**Docs updated:**
- `website/docs/getting-started/ado.md` — removed all references to legacy migrator flags and their usage examples

### Why

The legacy migrator system was built to handle one-time schema migrations for resources created with old field names/shapes. Those migrations are no longer needed, so the entire subsystem is being removed to reduce complexity.


michael-johnston:
A question re: the cli removal. Might we want to keep the command in case we in future need new legacy migrators?

AlessandroPomponio:
I think in the future we shouldn't be removing validators as quickly as we have been doing, and we should try to avoid having situations where legacy migrators are needed.

I also don't think this was the best approach for handling these migrators, I should've had a look at wrap validators instead.

In general, if we ever need to bring these - back, we can always fish things from this commit, so I don't think it's a problem.

And, to be clear, this PR only removes the legacy validator system, not ado upgrade, which will still be part of the CLI.

michael-johnston (approved):

PR #1106, opened 2026-06-30
feat(trim): implement fall-back strategies for entities that do not measure the targetOutput

### LINKED ISSUE #1064: feat(TRIM): add support for entities that are missing the target variable measurement

The current version of TRIM requires that your experiment can obtain a target variable measurement for every entity in the space and that these measurements are of the same type (numbers or strings, vectors are not supported).

you want to happen.

It is difficult to design a one-size-fits-all solution. So we'll support 3 mutually exclusive modes of operation when using TRIM to apply the experiment on an entity which does not measure the TRIM target variable:

- (default) Raise an exception (current behaviour)
- Use a default value, with up to `missingTargetMeasurements.budget` times and raise an exception when there's too many measurements that are missing the target variable
- Skip an Entity, up to `missingTargetMeasurements.budget` times. Raise an exception if skipping too many entities OR if running out of entities

The default behavior is to raise an exception on first missing target variable measurement. The default of  `missingTargetMeasurements.budget` is None which means that the budget is infinite. The `missingTargetMeasurements.budget` parameter of TRIM must either be None, or greater than 0.

To pick a mode set the parameter missingTargetMeasurements.mode which can be one of:

- `RaiseError`
- `InjectDefaultValue` - which uses the value of `missingTargetMeasurements.defaultValue`
- `Skip`

if `missingTargetMeasurements` is not set, then it defaults to `missingTargetMeasurements.mode = RaiseError`

We'll also refactor the NoPriorsSampleSelector to keep track of whether the operator managed to measure the target variable or not.

- Update the actuators to use a default value instead: This is a hack

**Additional context**. Add any other context or screenshots about the feature
request here.

VassilisVassiliadis:
Fixes #1064 

Changes:
- implement fall-back strategies for entities that do not measure the targetOutput, details in #1064 
- both samplers use ADO api to check whether an Entity has measured the targetOutput or not
- use fully qualified targetOutput name and auto-resolve it when using shorter version, error when discoverySpace does not contain it

To support the "Skip" fall-back strategy I had to refactor a bit the TRIM Sampler. A positive side-effect of the refactoring is that it is relatively easy now to support batchSizes greater than 1.


VassilisVassiliadis:
@danielelotito this is now ready for review

website/docs/operators/trim.md
```diff
@@ -189,6 +192,65 @@ parameters:
...
+parameters:
+  missing_target_variables:
+    mode: InjectDefaultValue
+    defaultValue: -1.0
+```
+
+```yaml
+# Silently skip unmeasurable entities; fail after 5 misses
+parameters:
+  missing_target_variables:
```
VassilisVassiliadis: Daniele, sometimes TRIM refers to the "targetOutput" as "target variable". Do you want to keep just 1 of the terms ?

danielelotito:
Really good job @VassilisVassiliadis !
Can you have a look at "self._injected_rows_df" in "_handle_missing_target_row" ?  maybe it is not used.
```bash
❯ grep -r '_injected_rows_df' plugins/operators/trim
plugins/operators/trim/src/trim/trim_sampler.py:197:            self._injected_rows_df = pd.concat(
plugins/operators/trim/src/trim/trim_sampler.py:198:                [self._injected_rows_df, one_additional_row], ignore_index=True
plugins/operators/trim/src/trim/trim_sampler.py:1058:        self._injected_rows_df: pd.DataFrame = pd.DataFrame()
```

AlessandroPomponio:
Please wait before merging this PR - I might need you to do some changes

website/docs/operators/trim.md
```diff
@@ -189,6 +192,65 @@ parameters:
...
+**Purpose:** Controls how both the no-priors and the iterative sampler handle
+entities that do not produce a measurement for `targetOutput` (e.g. due to an
+`InvalidMeasurement`). Three modes are available:
+
+<!-- markdownlint-disable MD013 MD060 -->
+
+| Mode | Behaviour |
+|---|---|
+| `RaiseError` | Raise `InsufficientDataError` immediately **(default)** |
+| `InjectDefaultValue` | Inject `defaultValue` as a synthetic row; entity counts towards the sampling quota |
```
danielelotito: To have a unified naming convention, we can rename quota -> budget here and in similar places

website/docs/operators/trim.md
```diff
@@ -189,6 +192,65 @@ parameters:
...
+| `budget` | `int \| None` | `None` | Maximum number of entities allowed to miss the target before raising `InsufficientDataError`. Must be `> 0` when set. |
+
+<!-- markdownlint-enable MD013 MD060 -->
+
+**Tuning Guidance:**
+
+- Use `InjectDefaultValue` when missing measurements are occasional and you want
+  the model to treat those entities as having a known sentinel value (e.g.
+  `-1.0` for a non-negative target, or `0.0` for a strictly positive one). An
+  inappropriate `defaultValue` can silently degrade model quality.
```
danielelotito:
```suggestion
- Use `InjectDefaultValue` when missing measurements are rare and you want
  the model to treat those entities as having a known value in use cases
  such as [censored regression](https://en.wikipedia.org/wiki/Censored_regression_model).
  We recommend extreme caution in using this mode, as it typically degrades
   model quality.
```

website/docs/operators/trim.md
danielelotito: Target variable is common in statistical learning lingo, and targetOutput makes clear that we want an ado output property. I used both terms for this reason, feel free to clarify.
However I adopted lowercaseUppercase in yaml declaration.

We may want to keep the same style/convention here, I leave a dedicated comment as well

danielelotito: I adopted lowercaseUppercase in yaml declaration.
We may want to keep the same style/convention here, as it is used in ado (e.g. examples/ml-multi-cloud)

website/docs/operators/trim.md
```diff
@@ -569,10 +631,16 @@ TRIM maintains a rolling holdout set:
...
-target output property, if this is not the case TRIM raises
-`InsufficientDataError`. To inspect what happened you can show the entities in
-the space with the following command
+By default, TRIM assumes that all measurements produce the observed target
+output property. If an entity does not produce a value for the target variable,
+TRIM raises `InsufficientDataError`. If this is expected behaviour in your
+experiment (e.g. some entities don't measure the target variable), configure
+[`missing_target_variables`](#missing_target_variables): set `mode` to
+`InjectDefaultValue` (and supply a `defaultValue`) to substitute a sentinel, or
+to `Skip` to exclude those entities from model training entirely.
```
danielelotito:
```suggestion
[`missing_target_variables`](#missing_target_variables): set `mode` to
to `Skip` to exclude those entities from model training entirely,
advanced users can also use `InjectDefaultValue` (and supply a `defaultValue`) 
to impute missing values of those entities with the supplied `defaultValue`.
```

website/docs/operators/trim.md
```diff
@@ -189,6 +192,65 @@ parameters:
...
+  inappropriate `defaultValue` can silently degrade model quality.
+- Use `Skip` when entities that cannot produce a target measurement should be
+  excluded entirely from model training. TRIM will draw extra candidates from
+  the pool to compensate.
+- Set `budget` to cap total misses and surface unexpected failures early.
+
+**Examples:**
+
+```yaml
+# Inject a sentinel value for entities that cannot be measured
```
danielelotito, line 238: The imputation of missing target measurements in TRIM should not be treated as annotation of a
Sentinel value, CS (https://en.wikipedia.org/wiki/Sentinel_value),
because it degrades the performance of the regressor.

I understand that in scientific computing experiment may fail systematically to provide a given target,
the best practice is imo to have a proper handling at the experiment level. 
When this is not possible or for unforeseen failures the `skip` feature is useful.
The injection mode needs to be reserved to very niche use-cases in which an expert statistical learning person wants to have more control and he is aware of the consequences.

website/docs/operators/trim.md
VassilisVassiliadis: Indeed, we're meant to use camelCase for pydantic fiends. Updating the code now.

VassilisVassiliadis:
> Really good job @VassilisVassiliadis ! Can you have a look at "self._injected_rows_df" in "_handle_missing_target_row" ? maybe it is not used.
> 
> ```shell
> ❯ grep -r '_injected_rows_df' plugins/operators/trim
> plugins/operators/trim/src/trim/trim_sampler.py:197:            self._injected_rows_df = pd.concat(
> plugins/operators/trim/src/trim/trim_sampler.py:198:                [self._injected_rows_df, one_additional_row], ignore_index=True
> plugins/operators/trim/src/trim/trim_sampler.py:1058:        self._injected_rows_df: pd.DataFrame = pd.DataFrame()
> ```

Good catch, that code is now dead. I was originally using it to track the injected rows but then I consolidated that code with the code that deals with inserting measured rows into "current_source_df".

danielelotito:
3 minor comments:
- _handle_missing_target_row inside _handle_new_measured_entity is called on an entity that _did_entity_measure_target_output already confirmed as a hit .  so it seems  it always returns via the entity_hit_in_source , it is okay but maybe you want to polish that edge as well
- Unit tests seems failing in the CI
- I appreciate the refactoring in trim_sampler.py, the code imo can now still benefit from separating some of the _methods in free functions (top-level functions taking explicit args), this matches the current style and makes the code readable.

plugins/operators/trim/src/trim/trim_sampler.py
```diff
@@ -56,6 +59,191 @@
...
+        current_source_df: The current source DataFrame (may be mutated).
+        missing_count: Running count of entities with missing target measurements.
+
+    Returns:
+        ``(one_additional_row, skip, current_source_df, missing_count)``
+        ``one_additional_row`` is the DataFrame row representing the entity
+        (measured or synthetic); ``skip=True`` means the entity produced no
+        usable measurement and should be dropped.
+    """
+    hit, series = entity_measured_target(entity, discoverySpace, params.targetOutput)
```
VassilisVassiliadis, line 185: Same issue here too

plugins/operators/trim/src/trim/samplers/no_priors_sampler.py
```diff
@@ -30,163 +34,222 @@ def samplerCompatibleWithDiscoverySpaceRemote(
...
+
+        for entity in list_of_entities:
+            if quota_count >= quota:
+                break
+
+            # Check whether the entity we yielded in the previous iteration produced
+            # a target measurement. By this point async_wrapper has already called
+            # discoverySpace.sample_store.refresh(), so the snapshot is current.
+            if last_entity is not None:
+                hit, _ = entity_measured_target(
```
VassilisVassiliadis, line 81: @michael-johnston this is the bit that's not super clear to me.

I'm trying to detect whether an Entity was measured or not in both phases of TRIM.

Challenges:
- entity identifier does not survive roundtrip to SampleStore - for example, an entity with an integer parameter "foo" is called `foo.1` in 1st RandomWalk and `foo.1.0` in 2nd RandomWalk
- there's no way to return information from sampler of 1st RandomWalk to sampler of 2nd RandomWalk - I'm propagating the OperationId of 1st randomwalk to 2nd RandomWalk to figure out which entities were attempted to be sampled
- due to the async consumption of samplers:
    - we cannot yield more than 1 entity per sample-request
    - we have to first yield an entity, then check in the followup call to generator whether the last entity was yielded or not
    - need to release the CPU control in between yields of the sampler using an asyncio.sleep() --> The assumption being that the next time the async generator gets the CPU control the last entity will have been measured.

VassilisVassiliadis: So basically, in this PR after the async iterator releases the CPU control, it refreshes the SampleStore. Then in the next generation, it checks whether the last yielded entity was measured or not. To do this, it compares by-value the measured entities in the space against the last yielded entity.

Then the samplers decide whether the sampling should stop or not. For the 1st RandomWalk (nopriors) the sampler stops when it has measured at least X points. For the 2nd RandomWalk (iterative model building) the sampler stops when TRIM decides it's time to stop.

VassilisVassiliadis: I think a better way to do this would be to propagate the RandomWalk operation to the Sampler it's using and then have the sampler use ado's API to decide whether the last yielded entity resulted to a Measurement that contained the TargetOutput observed-property or not.

VassilisVassiliadis: However, the challenge is that I don't think I can find the operation id before starting the operation.

michael-johnston: I'm trying to digest this to get to the actual problem. In understanding maybe incorrectly that you don;t think the following strategy works

> Then in the next generation, it checks whether the last yielded entity was measured or not. To do this, it compares by-value the measured entities in the space against the last yielded entity.

because in 

https://github.com/IBM/ado/blob/f3942155002ae055c82fbac270ca5eeb9a05c83c/plugins/operators/trim/src/trim/samplers/missing_target_utils.py#L63-L67

you think there is some issue with identifiers?

VassilisVassiliadis: Yes, I think that the identifier of an entity that contains at least one integer value varies depending on when you look at it it.

When you evaluate the identity of an Entity that you get out of loading a DiscoverySpace and prior to storing the Entity to the SampleStore it is `foo.1`. When you evaluate the identifier of the same entity but after you de-serialize the Entity object from the sample store it you find that the identifier is `foo.1.0`. 

I think that somewhere along this round-trip integers get turned into floats.

michael-johnston: So I first note the following which may not be obvious
- Entities compare on **constitutive property values** not identifier
- Identifier is NOT guaranteed to be the same for two entities that have identical constitutive property values. Different labels (identifiers) can be used for the same set of properties. The reason can be seen in case of something like ml-multi-cloud example where the identifier is a particular implicit string formatting of the property values.   

Given above if you have an Entity you can find if it is in a set of entities with 

```python
if last_entity in space.measured_entities:
    # it was measured with a ValidMeasurementResult 
```

Alternatively you need the stored_entity and you don't want to get measured_entities
```python
if (ents := space.storedEntitiesWithConstitutivePropertyValues(entity.constitutive_property_values))
```

In the case of TRIM sampler and how is set up, if last_entity is not in space.measurement_entities , its expected that the measurement produced InvalidMeasurementResult. 

If then you need to know particular values of measurements, because for example, some ValidMeasurements don't produce the metrics,  you could do  (after extracting the stored entity matching last_entity)

```python
### Get all the measurement results recorded for reference
measurement_results = stored_entity.measurement_results_for_experiment_reference(reference)
# filter based on property that indicates everything was measured e.g. is_valid
....
```
 
 or
 
 ```python
 values = stored_entity.valuesForObservedPropertyIdentifier(observed_property_identifier)
 # filter values
 ```
 
 ### Additional Info
 
I get the sense the real need is not exactly to identify the success of last_entity, but rather to ensure a dataframe is constructed with valid entities for a given property for all sampled entities

Some suggested strategies for two different cases: 
 
 The problem is experiments which send InvalidMeasurementResult for an entity, so nothing is recorded. 
         - This is ==  last_entity not in measured_entities
         - Strategy: On above skip or insert "placeholder" measurement. Can also be done for sampled entities.
            For example if placeholder, anytime you want to create the CSV you can do 
            ```python
            unmeasured_entities = sampled_entities - measuremed_entities 
            #Add rows for unmeasured entities
            ```

 The problem is experiments which record ValidMeasurementResults but do not have a value, or have an undesirable value, for target property 
       - If there is a signal property e.g. is_valid, can simply split the measured_entities_table on is_valid and then for is_valid=False create a new dataframe with placeholder values
       - If there is no signal property, can filter the table for entities with undesirable values for property

VassilisVassiliadis: The sampler doesn't know the experiment reference, does it?

VassilisVassiliadis: I think I can extend the trick I've got for the NoPrior sampler to provide argument to the sampler that are not part of the TRIM parameters. 

Then I can propagate the experiment reference from the TRIM operator to the Samplers. After this change, I should be able to refactor the code towards using the above ado APIs.

michael-johnston: > The sampler doesn't know the experiment reference, does it?

It does - samplers are passed the space to sample from.

michael-johnston: Two things

Just to confirm - you understand you should NOT use Entity.identifier to compare entities? Use Entity __eq__

Also If you think something is necessary and you can't see how to do it, open an issue or ask question here. Maybe no tricks are required c.f. exepriment reference above

VassilisVassiliadis: > It does - samplers are passed the space to sample from.

I was under the impression that a discovery space could contain multiple experimentReferences in it. Even so, I know how to identify which experimentReference the operator is using.

> Just to confirm - you understand you should NOT use Entity.identifier to compare entities? Use Entity eq

I understand.

> Also If you think something is necessary and you can't see how to do it, open an issue or ask question here. Maybe no tricks are required c.f. exepriment reference above

Sounds good, thank you

michael-johnston: > I was under the impression that a discovery space could contain multiple experimentReferences in it. 

It can

All experiments will be applied on each sample

You have the observed property id so if there are multiple experiments its straightforward to find the one the correct one.

Even with target property id, you can find it OR find that there are multiple experiments that measure it, which TRIM should id on start and exit with error.

VassilisVassiliadis: Then I think the current approach is OK.

I've updated TRIM to ensure that `targetOutput` points to the observed property of exactly 1 experiment.

I'm using the `entity_measured_target()` method to fetch the stored entity via `space.entity_for_point()` which internally calls `space.storedEntitiesWithConstitutivePropertyValues(entity.constitutive_property_values)`. The latter, handles edge cases like matching more than 1 entity, or the entity being non-measured in which case it basically returns a newly crafted Entity object.

After I obtain the stored entity I check whether it has a non-nan value for the `targetOutput` so I think we're all set. I've also rebased the code to account for the recent changes to how we're importing the orchestrator code.

I'll run a couple more tests and then I'll ask Daniele to review when he's back.

VassilisVassiliadis:
This PR packs too many things instead of focusing on just on the feature of fall-back strategies for entities that do not measure the targetOutput. Off the top of my head:

- it refactors the way TRIM reads the measurements that RandomWalk is sampling
- it fixes a bug causing TRIM to skip finalizing a model
- it fixes a bug causing TRIM to break right before dumping the model to the disk
- it fixes a bug where TRIM does not verify that the targetOutput is an observed property till it attempts to build a model for the 1st time
- refactors the 2 samplers to have a bit more similar approach because at the time I couldn't exactly work out how they behaved
- fall-back strategies for entities that do not measure the targetOutput

I'll close this issue without merging it and I will create individual Pull Requests for the above issues. After the individual bugs are fixed, I'm going to implement the fallback strategy but without refactoring the way TRIM reads the measurements that RandomWalk is sampling. This new PR will also include tests for the 3 different strategies.

PR #1122, opened 2026-07-01
refactor(core): rename search operations to explore

### LINKED ISSUE #850: refactor(metastore): change SEARCH/EXPLORE operation types

The documentation extensively refers to explore operators as type of operator that does sampling+measurement. 
 However `ado describe operators` lists the type of these operators as "Search". This is also the type that is stored with the YAML of an operation run with one of these operators (operatorType field). This is a bit confusion. 

That where search is shown know it says explore. (ado get operators, and YAML of operations) 

Change explore to search: this is easier technically speaking. However the term search is a bit narrow as not all these operators are "searching" for something

**Additional context**. 

The reason the problem exists is that originally operators that performed sampling+measurment were separated into to classes, characterise (samples to gain information about a space) and search (sampled to search for something). When writing docs however it became evident that this was an overly fine categorisation, as technically both types of operator look identical.  Hence the term "explore" was used to capture both. 

The reason its was not changed before is that the change is either breaking or if try not to break, difficult, as it means changing the enum of discovery operation types which is fundamental. Unfortunately there is no alias mechanism with python enums. 

- If we rename Search to  Explore: This is breaking: all YAML with "search" becomes invalid. Need new explore operator versions. 
- If we add Explore and keep Search: we have to add logic wherever any filtering or comparison on operation types is done to group them as the same.

AlessandroPomponio:
## Rename `operationType: search` → `operationType: explore`

This PR renames the `search` operation type to `explore` across the entire codebase to align the serialised value with what is displayed in `ado get operators`.

### What changed

- **Core enum** (`DiscoveryOperationEnum`): `SEARCH` is renamed to `EXPLORE`. A `_missing_` hook transparently maps the legacy `"search"` string to `EXPLORE`, so all existing stored operations and YAML files continue to work without modification.
- **All YAML files** (examples, plugins, tests, skills): updated `operationType: search` → `operationType: explore`.
- **Documentation**: all prose and code examples updated to use `explore`.
- **Migration guide** (`1x-to-2x.md`): a new "Non-breaking changes" section explains the rename, confirms backwards compatibility, and documents the optional `ado upgrade operations` command for updating stored records to the canonical new value.

### Backwards compatibility

No action is required for existing workflows. The old `"search"` value is silently accepted. New operations and YAML files should use `operationType: explore`.

### DIFF
--- a/.cursor/skills/examining-ado-operations/SKILL.md
+++ b/.cursor/skills/examining-ado-operations/SKILL.md
@@ -36,8 +36,8 @@ on, and whether measurements and results look healthy.
 Operations are applied to discoveryspaces. There are different types of
 operation. The General Workflow can be applied to all types of operation.
 
-In addition, the Explore/Search Workflow operations can be applied to
-Explore/Search operations.
+In addition, the Explore Operation Workflow can be applied to
+Explore operations.
 
 - Read [operations](../../../website/docs/resources/operation.md) documentation
   for details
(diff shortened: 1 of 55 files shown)


michael-johnston (dismissed):

AlessandroPomponio:
Updated to fix references to SEARCH that have been added in other commits merged to main

christian-pinto (approved):

PR #1123, opened 2026-07-01
refactor(tests): always override ado app dir in tests 

michael-johnston:
A number of CLI tests (19) invoke ado without --override-ado-app-dir. This leads to various race conditions with reading/writing files in the default ado app dir.

This change creates a fixtures that enforces app dir isolation by overriding AdoConfiguration.load, and uses it in ado/describe ado/get modules, so all tests loading will use a tmp dir.


tests/fixtures/ado_cli_isolation.py
```diff
@@ -0,0 +1,46 @@
...
+            _override_config_dir = tmp_path
+        return original_load(
+            cls, *args, _override_config_dir=_override_config_dir, **kwargs
+        )
+
+    monkeypatch.setattr(AdoConfiguration, "load", load_with_isolated_app_dir)
+
+
+@pytest.fixture
+def isolated_ado_app_dir(
```
AlessandroPomponio: I'd put autouse=True here

AlessandroPomponio (commented):
Seems great, thank you! Only one thing: can't we use our usual pattern of importing the fixture in the top level conftest.py so that we can reuse it without creating multiple copies (e.g., in describe/conftest.py and get/conftests.py)?

tests/fixtures/ado_cli_isolation.py
michael-johnston: I think this specific fixture was actually supposed to be removed. Reason why given below in answer to above question

michael-johnston:
> : can't we use our usual pattern of importing the fixture in the top level conftest.py so that we can reuse it without creating multiple copies (e.g., in describe/conftest.py and get/conftests.py)?

The idea here was to scope monkey patching of the load to specific modules that wanted it i.e. so it wouldn't happen transparently everywhere, breaking tests that wanted to test the actual real load path.

AlessandroPomponio:
Actually, we can always override the app dir - every test should create the environment it needs to run

mgazz (approved):

PR #1124, opened 2026-07-01
feat(core): improve versioning support for operators

michael-johnston:
Operators already supported independent versioning, however it was not explicitly stored in operation YAML and was usually just set to operator package version

This PR
- persists the version in operation YAML via new OperatorMetadata.version field
- updates the version field tyoe to be strictsemver to match Experiments
- updates all in-tree operators to have strict semver versions, separate from ado-package version or operator package version
- Allows setting the version in operation.yaml to pin it or leaving it None to take whatever latest version of operator is (current behaviour and most widely assumed)


orchestrator/core/operation/config.py
```diff
@@ -215,11 +215,11 @@ class OperatorMetadata(pydantic.BaseModel):
         ),
     ] = None
     version: Annotated[
-        Pep440VersionStr,
+        StrictSemVerStr,
         pydantic.Field(
             description=(
-                "PEP 440 version string for the operator (e.g. '0.1.0', "
-                "'1.2.3.dev4+abc.dirty').  Validated on construction."
+                "Algorithm version for this operator (strict SemVer "
```
AlessandroPomponio:
```suggestion
                "Versioning information for this operator (strict SemVer "
```

orchestrator/core/operation/config.py
```diff
@@ -290,6 +299,16 @@ class OperatorReference(pydantic.BaseModel):
         DiscoveryOperationEnum, pydantic.Field(description="The type of the operation")
     ]
     operatorName: Annotated[str, pydantic.Field(description="The name of the operator")]
+    operatorVersion: Annotated[
+        StrictSemVerStr | None,
+        pydantic.Field(
+            description=(
+                "Algorithm version of the referenced operator (strict SemVer "
```
AlessandroPomponio:
```suggestion
                "Versioning information of the referenced operator (strict SemVer "
```

orchestrator/core/operation/config.py
```diff
@@ -290,6 +299,16 @@ class OperatorReference(pydantic.BaseModel):
...
+                "MAJOR.MINOR.PATCH). When omitted at create time, resolved from "
```
AlessandroPomponio:
```suggestion
                "MAJOR.MINOR.PATCH). When omitted at creation time, resolved from "
```

orchestrator/modules/operators/collections.py
```diff
@@ -378,6 +380,61 @@ def wrapper(
...
+    metadata = collection.operators.get(ref.operatorName)
+    if metadata is None:
+        raise ValueError(
+            f"Operator {ref.operatorName} had no functions of type {ref.operationType}"
+        )
+    return metadata
+
+
+def resolve_operator_reference(ref: OperatorReference) -> OperatorReference:
+    """Resolve and pin the algorithm version for an operator reference.
```
AlessandroPomponio:
```suggestion
    """Resolve and pin the versioning information for an operator reference.
```

orchestrator/modules/operators/collections.py
```diff
@@ -378,6 +380,61 @@ def wrapper(
...
+        A copy of *ref* with ``operatorVersion`` set to the resolved version.
+
+    Raises:
+        ValueError: If the operator is not registered.
+        OperatorVersionMismatchError: If ``operatorVersion`` is set and does not
+            match the registry.
+    """
+    metadata = operator_metadata_for_reference(ref)
+    if ref.operatorVersion is None:
+        return ref.model_copy(update={"operatorVersion": metadata.version})
```
AlessandroPomponio: The docstring should reference `ref.operatorVersion` instead of just `operatorVersion`, I thought it was a parameter

In general, I also find the docstring to be deceiving: _"When ``operatorVersion`` is omitted, the version from the operator registry is pinned on the returned reference."_ to me it seemed like this would set the operatorVersion in the reference passed (especially since it's called `ref`). Saying "Returns an OperatorReference instance with operatorVersion set" is way less prone to misinterpretations in my opinion.

As a last point, I would avoid using `update` in model_copy, especially since we then reference the field name with a string, making it harder for us down the line to keep track of it. I'd split that line in two: model_copy call and assignment

tests/operators/test_operator_versioning.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def test_operator_metadata_identifier_uses_at_separator() -> None:
+    """OperatorMetadata.operatorIdentifier returns '{name}@{version}'."""
+    meta = OperatorMetadata(
+        name="my_op",
+        version="2.0.0",
+        configuration_model=_ExampleConfig,
+        example_configuration=_ExampleConfig(),
+        type=DiscoveryOperationEnum.SEARCH,
```
AlessandroPomponio: We need to update the branch from main and change these to be EXPLORE

tests/operators/test_operator_versioning.py
```diff
@@ -0,0 +1,150 @@
...
+    )
+    assert meta.operatorIdentifier == "my_op@2.0.0"
+
+
+def test_operator_metadata_reference_includes_version() -> None:
+    """OperatorMetadata.reference carries the algorithm version."""
```
AlessandroPomponio:
```suggestion
    """OperatorMetadata.reference carries the versioning information."""
```

tests/operators/test_operator_versioning.py
```diff
@@ -0,0 +1,150 @@
...
+        OperatorMetadata(
+            name="my_op",
+            version="1.0.2.dev17+5e50632",
+            configuration_model=_ExampleConfig,
+            example_configuration=_ExampleConfig(),
+            type=DiscoveryOperationEnum.SEARCH,
+        )
+
+
+def test_operator_reference_identifier_with_pinned_version() -> None:
```
AlessandroPomponio: - `test_operator_metadata_identifier_uses_at_separator`
- `test_operator_metadata_reference_includes_version`
- `test_operator_reference_identifier_with_pinned_version`

Basically test the same things

tests/operators/test_operator_versioning.py
```diff
@@ -0,0 +1,150 @@
...
+
+def test_resolve_operator_reference_exact_match() -> None:
+    """Explicit operatorVersion matching registry resolves successfully."""
+    ref = OperatorReference(
+        operatorName="random_walk",
+        operationType=DiscoveryOperationEnum.SEARCH,
+        operatorVersion="2.0.0",
+    )
+    resolved = resolve_operator_reference(ref)
+    assert resolved.operatorVersion == "2.0.0"
```
AlessandroPomponio: These test the same thing

tests/operators/test_operator_versioning.py
AlessandroPomponio, line 1: All the tests that don't pin versioning information should somehow fetch it dynamically or they will rot and require manual fixes every time we change the version.

website/docs/resources/operation.md
```diff
@@ -107,8 +107,14 @@ for example:
 module:
   operatorName: rifferla # The name of the operator
   operationType: modify #The type of the operation/operator
+  # operatorVersion: "2.0.0" # Optional: pin a specific algorithm version
 ```
 
+When `operatorVersion` is omitted, at operation create time the
+currently installed version is resolved and used.
+Set `operatorVersion` explicitly to pin a specific operator version; a mismatch
+with the registry raises an error at create time.
```
AlessandroPomponio, line 116: I don't understand: why allow pinning an operator version if the user *must* pin the exact version we expect? Is it just a safety net?

website/docs/resources/operation.md
```diff
@@ -171,8 +177,8 @@ ones described in [resources](resources.md#common-features-of-resources) These
 are (with example values):
 
 ```yaml
-operationType: characterize # The type of the operation: characterize, modify etc.
-operatorIdentifier: profile-1.1 # The identifier of the operator including its version
+operationType: characterize # characterize, modify, etc.
+operatorIdentifier: profile@2.0.0 # operator id including algorithm version
```
AlessandroPomponio:
```suggestion
operatorIdentifier: profile@2.0.0 # operator id including versioning information
```

AlessandroPomponio (changes requested):
Given the changes, we should also decouple operator versioning in the respective pyproject to that of `ado-core`

website/docs/resources/operation.md
michael-johnston: TL;DR yes. 

Basically up until now there was no version, so we had implicitly behaviour of dynamically resolving to the installed version. 

This is most useful pattern as you normally create the operation YAML and then reuse and don't really care about a specific version installed. 

So to preserve this we have the behaviour of version being None == look it up. 

Now the issue is the version field exists - so we need a behaviour if someone bothers to set it. 
The simplest is that if they set a version we look for exact match in the collection. If it's not there it fails to be created. 

So it's an extra feature, may only be useful in edge-cases, but we have to have something since the version field is part of the model.

michael-johnston:
> Given the changes, we should also decouple operator versioning in the respective pyproject to that of ado-core

I do this separately though as that's the operator package version - and with the decoupling isn't related anymore to operator version :-)

tests/operators/test_operator_versioning.py
michael-johnston: Combined

tests/operators/test_operator_versioning.py
michael-johnston: The don't 

- The first tests that if NO version is given, it is fetched from the catalog
- The second tests that if a version is given, and its correct, there is no exception.

tests/operators/test_operator_versioning.py
michael-johnston: Dynamic versioning used now

AlessandroPomponio (approved):

christian-pinto (approved):

PR #1125, opened 2026-07-01
refactor(cli): unify experiment id parsing across commands

michael-johnston:
## Unify experiment ID parsing across CLI commands

The parsing rules for experiment identifiers were slightly different between `get experiment`, `describe experiment`, and `template --from-experiment`. This PR unifies them all behind a single lookup path.

### What changed

**New unified parsing rule**

Any of the following forms are now accepted everywhere:

```
{actuator_id}.{base_experiment_id}@{version}
           └── all parts optional ──┘
```

- **`{actuator_id}.`** — optional. If omitted, ado searches all actuators for a matching base experiment ID. An error is raised if more than one actuator matches.
- **`@{version}`** — optional. Accepts either a full semver (`@1.2.3`) or a major-version tag (`@v1`). If omitted, ado resolves by base identifier. An error is raised if the identifier is ambiguous (multiple catalog versions exist).

**Core changes**

- `orchestrator/modules/actuators/catalog.py` — `experimentForReference` gains two new `match_on` modes: `"base"` (match on base identifier only) and `"any"` (try fully-qualified → major → base in order). A new `_ambiguous_experiment_identifier_error` helper builds consistent error messages.
- `orchestrator/modules/actuators/registry.py` — New `experiment_for_experiment_identifier(identifier)` method encapsulates the full resolve-or-raise logic (actuator prefix detection, cross-actuator lookup, reference construction). Errors from the CLI layer (`NoActuatorWithExperimentError`, `TooManyActuatorsWithExperimentError`) are moved here from `cli/exceptions/actuators.py`.
- `orchestrator/schema/reference.py` — `_parse_experiment_part_from_string` and `ExperimentReference.referenceFromString` gain an `allow_parameterization=False` mode for CLI use, which also accepts `@v1`-style major-version suffixes.
- `orchestrator/cli/utils/resources/experiments.py` — All per-command lookup logic replaced by a single `_ado_lookup_cli_experiment` helper that delegates to `registry.experiment_for_experiment_identifier`.
- `orchestrator/cli/exceptions/actuators.py` — **Deleted**; its error classes moved to `orchestrator/modules/actuators/errors.py`.
- `orchestrator/cli/resources/experiment/describe.py` and `get.py` — Simplified to use the new unified helper.

> **Note:** Parameterized experiment IDs are not handled; behaviour is unchanged from before.

**Tests**

New test modules cover: reference string parsing with `allow_parameterization=False`, `@v1` major-version suffix parsing, `match_on="base"` and `match_on="any"` catalog lookup modes, cross-actuator disambiguation, regression for silently dropped version information, and end-to-end CLI `get`/`describe` scenarios.


orchestrator/cli/utils/resources/experiments.py
```diff
@@ -4,12 +4,33 @@
...
+    """Split ``resource_id`` on the first ``.`` unless ``@`` precedes it."""
+    dot_index = resource_id.find(".")
+    if dot_index == -1:
+        return None, resource_id
+
+    at_index = resource_id.find("@")
+    if at_index != -1 and at_index < dot_index:
+        return None, resource_id
+
+    return resource_id[:dot_index], resource_id[dot_index + 1 :]
```
AlessandroPomponio: This seems like a very specific function for experiment names and it's used only in `experiment_reference_from_cli_resource_id` -  not sure if we need it in its own function

orchestrator/cli/utils/resources/experiments.py
```diff
@@ -53,81 +74,220 @@ def get_actuators_from_experiment_id(
...
+
+
+def resolved_experiment_reference_from_cli_resource_id(
+    resource_id: str,
+    *,
+    registry: ActuatorRegistry | None = None,
+) -> ExperimentReference:
+    """Parse a CLI experiment resource identifier and return a resolved reference."""
+    experiment = experiment_from_cli_resource_id(resource_id, registry=registry)
+    return experiment.reference
```
AlessandroPomponio: I think these methods are fairly general and they shouldn't be here.

- `lookup_experiment_for_reference` is a wrapper around `experimentForReference`. Maybe it's worth changing the signature of `experimentForReference` to have two boolean options instead of `match_on` being a `Literal`:
  - `allow_major_version_matches` (equivalent of major_version_only - default True)
  - `allow_identifier_matches` (provides this path - default False)
- `experiment_reference_from_cli_resource_id` should probably be an `ActuatorRegistry` method and be called `experiment_reference_from_experiment_id`
- Not sure about the value that `experiment_from_cli_resource_id` and `resolved_experiment_reference_from_cli_resource_id` provide

orchestrator/cli/utils/resources/experiments.py
```diff
@@ -53,81 +74,220 @@ def get_actuators_from_experiment_id(
...
+            f"{base_experiment_identifier}.",
             stderr=True,
         )
-        raise typer.Exit(1) from e
+        raise typer.Exit(1) from error
+    except (UnknownExperimentError, AmbiguousExperimentIdentifierError) as error:
+        console_print(f"{ERROR}{error}", stderr=True)
+        raise typer.Exit(1) from error
+    except ValueError as error:
+        console_print(f"{ERROR}{error}", stderr=True)
```
AlessandroPomponio: These two look very similar

orchestrator/schema/reference.py
```diff
@@ -154,24 +155,46 @@ def _parameterization_from_suffix(
...
     Returns:
         A tuple of base experiment identifier, optional strict SemVer version,
         and optional parameterization values.
     """
     if "@" in experiment_part:
         base_identifier, version_and_params = experiment_part.split("@", maxsplit=1)
+        if not allow_parameterization:
+            version_match = re.match(
+                r"^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$", version_and_params
+            )
```
AlessandroPomponio: this is `_STRICT_SEMVER_PATTERN` from orchestrator/utilities/pydantic.py

orchestrator/cli/utils/resources/experiments.py
michael-johnston: It just for readability - if its placed in experiment_reference_from_cli_resource_id its a bit harder to read what is happening

orchestrator/cli/utils/resources/experiments.py
michael-johnston: Which two?

AlessandroPomponio: _ado_experiment_from_cli_resource_id and _ado_resolved_experiment_reference_from_cli_resource_id

orchestrator/cli/utils/resources/experiments.py
michael-johnston: Nearly everything is moved to registry now

orchestrator/cli/utils/resources/experiments.py
michael-johnston: Combined

christian-pinto (approved):

AlessandroPomponio:
@michael-johnston there are issues with the parsing of experiments containing version ids with dots 

For ml-multi-cloud I have fixed it in https://github.com/IBM/ado/pull/1199 but for example:

```
ado describe experiment finetune_full_benchmark-v1.0.0
⠏ Initializing Actuator RegistryERROR:  The SFTTrainer actuator was found but a match to SFTTrainer.finetune_full_benchmark-v1.0.0 was not found using mode any.
```

Bob's analysis from earlier said:

---

Root cause
The issue is a naming collision between the legacy in-identifier version suffix (-v1.0) and the @-based SemVer versioning system.

Trace through experiment_for_experiment_identifier("ml-multicloud-cost-v1.0")
Step 1 — [_split_resource_id_prefix](vscode-webview://0er3kgbjrlleohcg37c12ku6ruh0bmnsg7ro98jvcpkugg9q31to/ado/modules/actuators/registry.py:101)

The string ml-multicloud-cost-v1.0 contains a . (in v1.0), so the function splits on the first .:

prefix = ml-multicloud-cost-v1 ← treated as an actuator name
rest = 0
Since ml-multicloud-cost-v1 is not a registered actuator in self.actuatorIdentifierMap, the code falls through to the "no prefix" path.

Step 2 — [actuators_containing_experiment_with_base_identifier](vscode-webview://0er3kgbjrlleohcg37c12ku6ruh0bmnsg7ro98jvcpkugg9q31to/ado/modules/actuators/registry.py:583)

This calls _parse_experiment_part_from_string("ml-multicloud-cost-v1.0", allow_parameterization=False) which returns base_identifier = "ml-multicloud-cost-v1.0" (no @, no parameterization). It then searches all catalogs for an experiment whose experiment.identifier == "ml-multicloud-cost-v1.0".

The custom_experiments catalog stores the experiment under its major_version_identifier as the dictionary key ([addExperiment](vscode-webview://0er3kgbjrlleohcg37c12ku6ruh0bmnsg7ro98jvcpkugg9q31to/ado/modules/actuators/catalog.py:311)), and the experiment does have identifier ml-multicloud-cost-v1.0 — so one actuator is found: custom_experiments.

Step 3 — Build the reference

A fully-qualified string custom_experiments.ml-multicloud-cost-v1.0 is built and parsed via [ExperimentReference.referenceFromString](vscode-webview://0er3kgbjrlleohcg37c12ku6ruh0bmnsg7ro98jvcpkugg9q31to/ado/schema/reference.py:304). This splits on the first .:

actuator_identifier = custom_experiments
experiment_part = ml-multicloud-cost-v1.0
Then _parse_experiment_part_from_string("ml-multicloud-cost-v1.0") is called with allow_parameterization=True. The string contains no @, so it hits [_split_base_and_parameterization_suffix](vscode-webview://0er3kgbjrlleohcg37c12ku6ruh0bmnsg7ro98jvcpkugg9q31to/ado/schema/reference.py:69). The part v1.0 matches _is_parameterization_segment (it looks like the propertyIdentifier.value pattern where v1 is the property name and 0 is the value). So the result is:

base identifier = ml-multicloud-cost- (everything before the v1.0 segment)… actually let's be precise: splitting ml-multicloud-cost-v1.0 on _PARAMETERIZATION_SEGMENT_SPLIT (which splits on - only before prop.value pairs) — the segment v1.0 is treated as a propertyIdentifier.value pair (property=v1, value=0).
So the parsed reference ends up with:

experimentIdentifier = "ml-multicloud-cost" (base before the v1.0 param)
parameterization = [{property: "v1", value: 0}]
Step 4 — Lookup in catalog

[experimentForReference](vscode-webview://0er3kgbjrlleohcg37c12ku6ruh0bmnsg7ro98jvcpkugg9q31to/ado/modules/actuators/catalog.py:165) with match_on="any" tries all modes. The catalog entry has experiment.identifier = "ml-multicloud-cost-v1.0" but the reference has experimentIdentifier = "ml-multicloud-cost". No match is found → UnknownExperimentError.

Why ado get experiments shows it
ado get experiments simply enumerates all registered experiments via catalog.experiments — it iterates the raw dictionary values without any reference parsing. So the entry is found fine there.

Issue #376, opened 2026-01-12
feat(cli): add ado get <resource> --related-to <resource-id>


AlessandroPomponio:
## Issue

People might want to get e.g., operations related to a space. At the moment they could retrieve these identifiers as part of `ado show related` but not retrieve the contents

AlessandroPomponio:
The issue here is that we would need to know what kind of resource the provided id relates to. Could be something like ado get resource --related-to kind=id

PR #1127, opened 2026-07-01
feat(core): add document resource type

### LINKED ISSUE #925: feat: document resource type for storing reports and plans

With agents generating reports or plans on projects, space or operations, MD files accumulate locally. There is no way for others to access the reports, so they must spend the tokens and time again. Also the organisation is ad-hoc. 

The ability to store the reports or planning or todo documents with the resource YAML potentially as an ado resource like  "document" - similar to datacontainer that can then link to resource as a related resource. 

- Storing in metadata of the resource. Simple, but then the resource YAML could become bloated, and consume bandwidth (every fetch of YAML has to fetch all reports)

**Additional context**. 

For projects there is no resource so there is no current way to associate something with it via an existing ado mechanism.

michael-johnston:
This PR adds a new resource type, document, for storing markdown or html docs that describe spaces, operations or combinations thereof. 

Create with `ado create document -f DOCUMENT.YAML`.  Read with `ado get document $RESOURCEID`. A document can have multiple related resources. It's not required that it has any - for example a document containing a report about the entire project at a point in time should not have to list every single space and operation, and there is no "project" resource. 

Example document yaml
```yaml
metadata:
  name: Example operation report
  description: Summary report for an operation
contentType: "markdown"
content: |
  # Operation report

  This is an example document resource.
relatedResources:
  - id: operation-test-12345678
     role: parent
  ```
  
  The ` relatedResources` is a list of dictionaries. The dicts have to keys 
  - id (the id of the related resource)
  - role - either parent or child. 

`ado describe document` will

- for markdown, render the content field using rich (when redirected to a file, rich handles it)
- for html, write the content field html (when redirected to a file, the same is done)

#925

### CODE REVIEW

**Comment by AlessandroPomponio** [comment]
From just the description, I wonder whether this design will lead to more of the same problems that we already have for Datacontainers. As it stands, Datacontainers store their content in the resources table and can be very large (we have examples of datacontainers from sft-trainer which exceed 18MB), ballooning the whole table up and slowing down retrieval.

In general, as well, storing byte-like objects in a stringified format in a JSON/text column seems like an antipattern. I would think about only storing references to BLOBs and having a dedicated table for those. Later on in the future we could also take advantage of it to slim down datacontainers

**Comment by michael-johnston** [comment]
@AlessandroPomponio We could simplify this and remove the attachments functionality for initial implementation. Add it back later if needed/when we decided how to store the attachments (i.e. link instead of raw data, where the raw data goes).

**Comment by AlessandroPomponio** [comment]
Sounds good!

File: .cursor/skills/examining-ado-operations/SKILL.md [hunk 1]
```diff
@@ -125,11 +125,20 @@ failure. In this case:
...
-- Check if there is an existing report for this operation in
-  `reports/<ado_context_name>/`
-- If yes, check if that report indicated the operation was finished
-  - If yes, ask the user if they want to replace it with a new report
-  - If no, continue with creating new report
+- Query the metastore for an existing document linked to this operation:
+
+  ```bash
+  uv run ado get document -q 'config.relatedResources=OPERATION_ID'
+  ```
```
AlessandroPomponio, line 132: Maybe before adding documents we could implement https://github.com/IBM/ado/issues/376 and remove this field, since we already have a related resources table that has this exact purpose anyway.

This, however, would mean that we wouldn't be able to just retrieve the resource if we want to display this information, we would also have to lookup the related resources table

ado/core/document/config.py
```diff
@@ -0,0 +1,34 @@
...
+        str,
+        pydantic.Field(description="Markdown body of the document"),
+    ]
+    relatedResources: Annotated[
+        list[str],
+        pydantic.Field(
+            default_factory=list,
+            description="Identifiers of related ado resources, if any",
+        ),
+    ]
```
AlessandroPomponio, line 55: See other comment

ado/core/document/resource.py
```diff
@@ -0,0 +1,29 @@
...
+from ado.core.resources import ADOResource, CoreResourceKinds
+from ado.utilities.pydantic import Defaultable
+
+
+class DocumentResource(ADOResource):
+    """A resource that stores markdown documents and optional attachments."""
+
+    @staticmethod
+    def _identifier_from_data(data: dict[str, Any]) -> str:
+        return f"document-{str(uuid.uuid4())[:8]}"
```
AlessandroPomponio: This doesn't require data at all - you can inline it in a lambda in the field directly

ado/core/document/resource.py
```diff
@@ -0,0 +1,29 @@
...
+
+    version: Annotated[str, pydantic.Field()] = "v1"
+    kind: Annotated[CoreResourceKinds, pydantic.Field()] = CoreResourceKinds.DOCUMENT
```
AlessandroPomponio: Unrelated to this PR but I wonder if in theory this type of definition would allow me to create a `DocumentResource` with a different `CoreResourceKinds`, since the Annotated bit still has CoreResourceKinds

tests/core/document/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+@pytest.fixture
+def document_configuration_file() -> pathlib.Path:
+    """Return path to a valid document configuration fixture."""
+    return pathlib.Path("tests/fixtures/document.yaml")
+
+
+@pytest.fixture
+def document_with_attachment_configuration_file() -> pathlib.Path:
+    """Return path to a document configuration fixture with attachments."""
+    return pathlib.Path("tests/fixtures/document_with_attachment.yaml")
```
AlessandroPomponio: iirc we don't support attachments anymore

tests/core/document/test_document_config.py
```diff
@@ -0,0 +1,47 @@
...
+import yaml
+
+from ado.core.document.config import DocumentConfiguration
+from ado.utilities.output import pydantic_model_as_yaml
+
+
+def test_document_configuration_defaults() -> None:
+    """DocumentConfiguration defaults relatedResources and attachments."""
+    config = DocumentConfiguration(content="Example report")
+    assert config.relatedResources == []
```
AlessandroPomponio: We don't support attachments anymore and this seems like a fairly pointless test given pydantic's model

tests/core/document/test_document_config.py
```diff
@@ -0,0 +1,47 @@
...
+    """DocumentConfiguration round-trips through YAML."""
+    config = DocumentConfiguration(
+        content="# Report\n\nBody text",
+        relatedResources=["operation-abc-12345678"],
+        attachments={"chart.png": "cGV4"},
+        metadata={"name": "Test report"},
+    )
+    yaml_text = pydantic_model_as_yaml(config)
+    restored = DocumentConfiguration.model_validate(yaml.safe_load(yaml_text))
+    assert restored == config
```
AlessandroPomponio, line 38: No attachments

tests/core/document/test_document_resource.py
```diff
@@ -0,0 +1,52 @@
...
+    config = DocumentConfiguration(
+        content="# Report\n\nExample body",
+        relatedResources=["operation-test-12345678"],
+        attachments={"chart.png": "cGV4"},
+    )
+    resource = DocumentResource(config=config)
+
+    dumped = resource.model_dump()
+    restored = DocumentResource.model_validate(dumped)
+    assert restored == resource
```
AlessandroPomponio, line 20: no attachments and i think there is already another store-load test

tests/core/document/test_document_resource.py
```diff
@@ -0,0 +1,52 @@
...
+
+    assert re.fullmatch(r"document-[0-9a-f]{8}", resource.identifier)
+
+
+def test_document_resource_kind_is_pinned() -> None:
+    """DocumentResource kind is pinned to document."""
+    config = DocumentConfiguration(content="Example report")
+    resource = DocumentResource(config=config)
+
+    assert resource.kind == CoreResourceKinds.DOCUMENT
```
AlessandroPomponio: Pointless tests

tests/core/document/test_document_resource.py
```diff
@@ -0,0 +1,52 @@
...
+
+def test_document_resource_requires_content() -> None:
+    """DocumentResource requires config content."""
+    with pytest.raises(pydantic.ValidationError, match="content"):
+        DocumentResource.model_validate(
+            {
+                "kind": CoreResourceKinds.DOCUMENT,
+                "config": {},
+            }
+        )
```
AlessandroPomponio: Duplicate test

tests/fixtures/document_with_attachment.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: Report with attachment
+  description: Example document with an embedded image
+content: |
+  # Report with chart
+
+  ![Performance chart](perf_chart.png)
+relatedResources:
+  - space-test-12345678
+attachments:
+  perf_chart.png: cGV4
```
AlessandroPomponio: no attachments

AlessandroPomponio (changes requested):
The PR description says this is also meant for HTML reports. If we keep it to markdown only, though, we could also have a `describe` command that uses `rich`'s Markdown renderer to render and print the document to the terminal.

This PR is also missing updating the ado cli reference doc

.cursor/skills/examining-ado-operations/SKILL.md
michael-johnston: The thing if its not there how on create do we know the resources the report is related to? :-) 

This field is the same as `spaces` in operation yaml or `samplestore` in space yaml

michael-johnston:
> The PR description says this is also meant for HTML reports. If we keep it to markdown only, though, we could also have a describe command that uses rich's Markdown renderer to render and print the document to the terminal.

Yes I did think of this and it was originally markdown. However then Daniele or Vassilis was showing me that bob was creating HTML reports with graphical elements and I thought we should store them. 

I mean we could have it also id html and describe open a brokers.

It would be good to have the describe for markdown in any case as reading the reports as human is not easy as is.

michael-johnston:
If implemented describe for markdown and html, also discriminating behaviour between tty and no tty

ado/cli/resources/document/describe.py
```diff
@@ -0,0 +1,108 @@
...
+    """Return True when stdout is an interactive terminal."""
+    return stdout_console.is_terminal
+
+
+def describe_document(parameters: AdoDescribeCommandParameters) -> None:
+    """Print a human-friendly description of a document resource.
+
+    When stdout is a terminal, markdown is rendered with rich and HTML is opened
+    in the default browser. When stdout is redirected (pipe or file), the raw
+    ``content`` body is written with no styling.
```
AlessandroPomponio: I don't think we should support this behaviour to keep a clear cut definition of describe. Markdown reports should be always rendered, while HTML reports should exit with a clear error when the output is not a terminal.

ado/cli/resources/document/describe.py
```diff
@@ -0,0 +1,108 @@
...
+        sys.stdout.write(config.content)
+        if not config.content.endswith("\n"):
+            sys.stdout.write("\n")
+        return
+
+    header_parts: list[Text | str] = [
+        Text.assemble(
+            ("Identifier: ", "bold"), (document_resource.identifier, "bold green")
+        ),
+        Text.assemble(("Content type: ", "bold"), (config.contentType,)),
```
AlessandroPomponio: I don't think there's any useful information here

ado/cli/resources/document/describe.py
```diff
@@ -0,0 +1,108 @@
...
+    console_print,
+    stdout_console,
+)
+from ado.core.resources import CoreResourceKinds
+from ado.metastore.base import ResourceDoesNotExistError
+
+
+def _stdout_is_terminal() -> bool:
+    """Return True when stdout is an interactive terminal."""
+    return stdout_console.is_terminal
```
AlessandroPomponio: i don't think we need this function

tests/core/document/test_document_config.py
```diff
@@ -11,31 +11,45 @@
...
 def test_document_configuration_requires_content() -> None:
     """DocumentConfiguration requires content."""
     with pytest.raises(pydantic.ValidationError):
         DocumentConfiguration.model_validate({})
 
 
+def test_document_configuration_default_content_type() -> None:
+    """DocumentConfiguration defaults contentType to markdown."""
+    config = DocumentConfiguration(content="Example report")
+    assert config.contentType == "markdown"
```
AlessandroPomponio, line 23: it's the model's default

AlessandroPomponio (dismissed):

ado/cli/resources/document/describe.py
michael-johnston: How can I save a human readable md file then with the content?

ado/cli/resources/document/describe.py
michael-johnston: You are suggesting not to output the identifier or content type in describe?

tests/core/document/test_document_config.py
michael-johnston: yes - this tests that it still is, in case something changes it unexpectedly.

ado/cli/resources/document/describe.py
AlessandroPomponio: The identifier could be okay but the content type definitely isn't useful

ado/cli/resources/document/describe.py
AlessandroPomponio: I would say `ado get doc -o content`. We could also internally map `content` to `config` for other resource types

michael-johnston: I think the definition of describe should be to output human readable version of the content. I think it's fine for this output to adapt based on the capabilities of what stdout it connected to.

ado/cli/resources/document/describe.py
michael-johnston: Ok I will take it then that the identifier *is* okay :-)

.cursor/skills/examining-ado-operations/SKILL.md
AlessandroPomponio: that's true - we should add some kind of mechanism to ensure this field doesn't become stale, though

michael-johnston: How does it become stale

ado/cli/resources/document/describe.py
AlessandroPomponio: The describe help says:
>  Print a human-friendly description of a resource or an experiment.

I think rendered markdown fits perfectly this definition, and I can see how despite not being "printed", saving an html file and opening it in the browser can map to it as well.

If we instead detect the redirect and either:
- Print raw markdown
- Print raw html

We're not printing any kind of human-friendly description. We're just allowing you to `get` the resource content by means of describe.

.cursor/skills/examining-ado-operations/SKILL.md
AlessandroPomponio: If a new operation/space/whatever becomes related to the report and it's either:
- Not added to the `relatedResources` field
- Not appropriately added to the `resources_relationship` table

ado/cli/resources/document/describe.py
michael-johnston: Not really if you send the markdown to file (as currently implemented) you get an md file with proper formatting - humans readable 

Screenshot 2026-07-28 at 09 13 11


if you get the yaml you get a compressed string - not really human readable.


Screenshot 2026-07-28 at 09 12 09


if you send the rich formatted output to a file then you get additional formatting symbols - not really human readable.

ado/core/document/resource.py
AlessandroPomponio: This is indeed an issue - please tighten it to an Annotated[Literal[CoreResourceKinds.DOCUMENT], pydantic.Field()]

ref: https://github.com/IBM/ado/issues/1294

.cursor/skills/examining-ado-operations/SKILL.md
michael-johnston: This is already a possibility, and has to be handled by a convention. There is no way to validate that whatever is in the field (string of text) is actually related to what is written or not. 

So for reports that are on an operation/space - we assume who ever writes the report will put the correct resource ids.  We can't edit this afterwards as of this PR - but it is tied to single resource ids anyway, so can't become stale in the sense you say. 

For report that are on a project, adding every single resource in the project to the related resources seems overkill, although nothing stopping anyone. For project reports the date of the report and the date of the resources is enough to tell you what was present when it was written. 

For a document which is about future to-be-created spaces/operations etc. (a study)  there is nothing to add on creation (usually) and the related resources can't be edited. With the current capabilities of document we would rely on a convention, that would be to define labels.

ado/core/document/resource.py
AlessandroPomponio: Actually it's not as straightforward because then coercion doesn't kick in - i'll have to troubleshoot it as part of 1294

AlessandroPomponio: I had to make CoreResourceKinds inherit from str as well in https://github.com/IBM/ado/pull/1296 and now everything is smooth sailing again

michael-johnston: What to do here - in current branch  Annotated[Literal[CoreResourceKinds.DOCUMENT], pydantic.Field()] this won't work because it doesn't allow the string "document" as you found.

ado/cli/resources/document/describe.py
AlessandroPomponio: > if you send the markdown to file (as currently implemented) you get an md file with proper formatting 

That's fine, but that's why I suggested `-o content`, not `-o yaml`, to mimic what we already have for `-o config` for other resource types. I do not expect `-o yaml` to produce any kind of useful output for a nested md block.

Again, this would be the only ad-hoc handling for `describe` compared to any other class. The way `describe` is implemented throughout the code is the definition of a `__rich__` method that is then automatically used by rich to do pretty prints. The only exception I would suggest having to this is the temporary html saving to a file and opening in web browser for interactive terminals

michael-johnston: I find it very strange that if we use our human readable command `describe` and pipe to file we don't get easily human readable result. Seems counterintuitive for the CLI behaviour (if I just read CLI commands and docs). 

but lets step back - do you know if rich auto removes the additional terminal colouring characters on sending to non tty - I never actually tested it. 

With describe experiment if you redirect to file there is none.

.cursor/skills/examining-ado-operations/SKILL.md
michael-johnston: @AlessandroPomponio Let me know if you have a concrete alternate proposal to above, otherwise we should just go with this.

AlessandroPomponio: I had another look and it seems to me that the `relatedResources` field doesn't translate to relationships being added to the `resources_relationship` table, it's just an annotation, right? If so, then this doesn't really matter

ado/cli/resources/document/describe.py
AlessandroPomponio: > I find it very strange that if we use our human readable command `describe` and pipe to file we don't get easily human readable result. Seems counterintuitive for the CLI behaviour (if I just read CLI commands and docs).

How is rendered markdown non human readable?



> do you know if rich auto removes the additional terminal colouring characters on sending to non tty

Yes it does (which is also why we should change our coloured logging handling to rich's `logging` integration down the line - i'll open an issue about it because I forgot about this at least 3-4 times

.cursor/skills/examining-ado-operations/SKILL.md
michael-johnston: I will update so they are stored (they should be)

michael-johnston:
So this the next steps

- Ensure the relationships are added to the db
- Remove isatty branching as rich removes its terminal annotations on detecting the redirect already
- Change html behaviour just to print the html - either to terminal or file 
    - human readable description of a document resource with content in markdown format is the rich rendering of the markdown - both to terminal and file (rich handles difference)
    - human readable description of a document resource with content in html format is the "print" of the html content

michael-johnston:
@AlessandroPomponio latest version is ready

ado/cli/resources/document/create.py
```diff
@@ -48,8 +68,27 @@ def create_document(parameters: AdoCreateCommandParameters) -> str | None:
...
+        console_print(
+            f"{ERROR}Unknown related resource identifier(s): "
+            f"{', '.join(missing_related)}",
+            stderr=True,
+        )
+        raise typer.Exit(1)
+
     with Status(ADO_SPINNER_SAVING_TO_DB):
         sql.addResource(resource_to_be_created)
+        _add_document_relationships(
```
AlessandroPomponio: I would just have the for loop here - it won't be used anywhere else anyway

ado/cli/resources/document/describe.py
```diff
@@ -79,30 +49,16 @@ def describe_document(parameters: AdoDescribeCommandParameters) -> None:
...
-            encoding="utf-8",
-        ) as html_file:
-            html_file.write(config.content)
-            html_path = pathlib.Path(html_file.name)
-
-        webbrowser.open(html_path.as_uri())
-        console_print(
-            f"{INFO}Opened HTML document content in the default browser ({html_path})."
-        )
+        console_print(Group(*header_parts, config.content))
```
AlessandroPomponio: I would only write html to file - the normal use case in which a person does `ado describe` would otherwise typically create output that overflows the terminal and that is so dense to make it difficult to find where the command was first started.

I realise that this would be an unusual behaviour, though. Another option would be to print the content but add a warning like when tables are too big to use an output-file. We don't currently support this flag for describe, though.

ado/cli/resources/document/describe.py
```diff
@@ -15,61 +10,36 @@
...
 
     config = document_resource.config
 
-    if not to_terminal:
-        sys.stdout.write(config.content)
-        if not config.content.endswith("\n"):
-            sys.stdout.write("\n")
-        return
-
     header_parts: list[Text | str] = [
```
AlessandroPomponio: I think the markdown part should be handled at the level of `DocumentConfiguration` in a `__rich__` method

ado/core/document/config.py
```diff
@@ -9,6 +9,23 @@
 from ado.core.metadata import ConfigurationMetadata
 
 
+class RelatedResource(pydantic.BaseModel):
```
AlessandroPomponio, line 16: I feel like this is a very generic name and, unless we want to use it elsewhere (such in the multi input operators or something like that), it might confuse both people and agents.

A very scoped-down name could be `ResourceRelationshipToDocument`, but I wonder if this goes too far the opposite way

tests/ado/create/test_ado_create_document.py
```diff
@@ -66,8 +69,16 @@ def test_create_document(
...
-    document_file = "tests/fixtures/document.yaml"
+    """Create persists a document resource with no related resources."""
+    document_file = tmp_path / "document.yaml"
+    document_file.write_text(
+        yaml.safe_dump(
+            {
+                "metadata": {"name": "Create test report"},
+                "content": "# Report\n\nBody.\n",
+            }
+        )
```
AlessandroPomponio: This should probably be a `model_dump_json`/`serialise_pydantic_model` call from a model instance rather than this, just to avoid rot

tests/ado/create/test_ado_create_document.py
```diff
@@ -80,14 +91,153 @@ def test_create_document(
...
+        yaml.safe_dump(
+            {
+                "metadata": {"name": "Parent link report"},
+                "content": "# Report\n",
+                "relatedResources": [
+                    {"id": parent.identifier, "role": "parent"},
+                ],
+            }
+        )
+    )
```
AlessandroPomponio, line 122: This shouldn't be a dict either

tests/ado/create/test_ado_create_document.py
```diff
@@ -80,14 +91,153 @@ def test_create_document(
...
+        yaml.safe_dump(
+            {
+                "metadata": {"name": "Child link report"},
+                "content": "# Report\n",
+                "relatedResources": [
+                    {"id": child.identifier, "role": "child"},
+                ],
+            }
+        )
+    )
```
AlessandroPomponio, line 168: Should be pydantic model

tests/ado/create/test_ado_create_document.py
```diff
@@ -80,14 +91,153 @@ def test_create_document(
...
+    document_file.write_text(
+        yaml.safe_dump(
+            {
+                "metadata": {"name": "Missing related"},
+                "content": "# Report\n",
+                "relatedResources": [
+                    {"id": "i-do-not-exist", "role": "parent"},
+                ],
+            }
+        )
```
AlessandroPomponio: Pydantic model

ado/cli/resources/document/describe.py
michael-johnston: Lets start with no exceptional behaviour, and then if someone really has a problem with describing html without redirecting to file and losing history, then can open an issue.

ado/cli/resources/document/describe.py
michael-johnston: To expand, you mean the creation of the formatted markdown string is done in a __rich__ method?

ado/core/document/config.py
michael-johnston: I think we may be getting into weeds here - if there no clear signal what it should be or that there's a problem, lets leave it. The paths to where a person might encounter this and get confused seem narrow and also difficult to imagine what the outcome of the confusion would be.

ado/cli/resources/document/describe.py
AlessandroPomponio: Yes something like https://github.com/IBM/ado/blob/54af4d69ebf14ee6f2c9610201f3f719db9aa759/ado/core/samplestore/sql.py#L134-L145

ado/metastore/sqlstore.py
```diff
@@ -1474,6 +1474,36 @@ def delete_actuator_configuration(self, identifier: str) -> None:
...
+        import sqlalchemy.orm
+
+        with sqlalchemy.orm.Session(self.engine) as session:
+            try:
+                with session.begin():
+                    session.execute(
+                        sqlalchemy.text(
+                            r"DELETE FROM resource_relationships WHERE object_identifier=:identifier"
+                        ).bindparams(identifier=identifier)
+                    )
```
AlessandroPomponio, line 1487: This is missing a delete where the identifier is the `subject_identifier`

AlessandroPomponio (dismissed):

AlessandroPomponio (approved):
LGTM thanks

PR #1175, opened 2026-07-07
fix(vllm-performance): Delete deployment if timed out during creation

### LINKED ISSUE #1163: bug: timed out deployments are not deleted before retries
## Issue Description

When vllm_performance creates a deployment and this fails to become ready in time, the following attempt tries to create an already existing deployment that fails due to conflicts with the previous one. 

https://github.com/IBM/ado/blob/fd89ea09c4b795143664b9404525b7c45737c4d2/plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/create_environment.py#L18

Failure message:

```
Failed to create test environment vllm-qwen3-6-27b-fp8-fe19c86f9d984a4280f21fc8795b0b29:
Failed to create test environment (409)
Reason: Conflict
HTTP response body: {
  "kind": "Status",
  "status": "Failure",
  "message": "deployments.apps \"vllm-qwen3-6-27b-fp8-fe19c86f9d984a4280f21fc8795b0b29\" already exists",
  "reason": "AlreadyExists",
  "code": 409
}
```

The error is caught as a `K8sEnvironmentCreationError` and recorded as a InvalidMeasurementResult. 

## Expected behaviour

There are two possible options to solve this issue:
1. if it's not the first attempt, skip creation.
2. clean up and perform the deployment again from scratch

I would prefer the second option

FYI @christian-pinto

christian-pinto:
This PR fixes #1163 by deleting timed out deployment before they are created.


plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/create_environment.py
```diff
@@ -125,11 +126,23 @@ def create_test_environment(
...
+        # during retries.
+        logger.debug(
+            "Creating the {k8s_name} deployment has timed out. Deleting the deployment."
+        )
+        c_manager.delete_deployment(k8s_name=k8s_name, raise_if_not_found=False)
+        raise (e)
+
     logger.info("deployment ready")
     # service
     c_manager.create_service(k8s_name=k8s_name, template=service_template)
```
mgazz: We should catch any errors generated during the setup of the whole environment, cleanup the Kubernetes objects (deployment + service) and retry from scratch.

mgazz (commented):
Thank you for looking into this. The overall idea works for me but I think we should defend against all potential errors in the creation of the environment.

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/manage_components.py
```diff
@@ -178,16 +179,20 @@ def check_service_exists(self, k8s_name: str) -> bool:
...
-            name=k8s_name,
-        )
+        try:
+            self.kube_client_V1.delete_namespaced_service(
+                namespace=self.namespace,
+                name=k8s_name,
+            )
+        except ApiException as e:
+            if e.reason != "Not Found" or raise_if_not_found:
+                raise e
```
AlessandroPomponio: Given that you normally raise the exceptions and there is just a flag to silence not found errors, i'd rename it to `suppress_not_found_error` and change the raising logic

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/manage_components.py
```diff
@@ -225,19 +230,23 @@ def check_deployment_exist(self, k8s_name: str) -> bool:
...
+            self.kube_client.delete_namespaced_deployment(
+                namespace=self.namespace,
+                name=k8s_name,
+                body=client.V1DeleteOptions(
+                    propagation_policy="Foreground", grace_period_seconds=5
+                ),
+            )
+        except ApiException as e:
+            if e.reason != "Not Found" or raise_if_not_found:
+                raise e
```
AlessandroPomponio: Given that you normally raise the exceptions and there is just a flag to silence not found errors, i'd rename it to `suppress_not_found_error` and change the raising logic

mgazz (dismissed):
LGTM

mgazz (approved):

PR #1187, opened 2026-07-08
feat(core): add upgrade mechanism for orchestrator references in modules

### LINKED ISSUE #1186: feat(core): provide sample store resource upgrade validators for ado 1 -> 2 transition
## Summary

After the Python package was renamed from `orchestrator` to `ado`, resources that were persisted before the rename can no longer be loaded because their stored module names still reference the old package. This causes a fatal import error at load time. It would be worth adding a compatibility layer that transparently rewrites legacy module names on the way in, so existing stored resources continue to work without any data migration.

## Motivation

Users who have resources persisted under the old package name currently experience a hard failure when those resources are loaded. Since the stored data cannot reasonably be expected to be migrated manually, the fix should live in the loading path itself — silently normalising any legacy name to the current one. This is a low-risk, targeted change that restores backward compatibility for all affected resources with minimal footprint.

## Scope

The changes should be confined to:

- **Module configuration loading** — the component responsible for validating and resolving module names when resources are deserialised. A small compatibility shim here would be sufficient to cover all affected resource types.
- **Tests for the module configuration** — a focused set of tests that cover the rewrite behaviour (legacy name → current name) as well as the pass-through cases (already-current names and absent values).
- **Code style** — linting should be run over any changed files to keep them consistent with the rest of the codebase.

No changes to stored data, no migration tooling, and no modifications to the upgrade command are expected or required.

AlessandroPomponio:
### Problem
Resources created before the `orchestrator` → `ado` package rename (commit 5687809) stored `moduleName` values prefixed with `orchestrator.` (e.g. `orchestrator.core.samplestore.sql`). Loading those resources after the rename would fail.

### Solution
Add a Pydantic `field_validator` on `ModuleConf.moduleName` that transparently rewrites any `orchestrator.*` prefix to `ado.*` at parse time, so existing stored resources remain usable without manual intervention.

### Changes
- **`ado/modules/module.py`** — New `migrate_orchestrator_module_name` validator on `ModuleConf` rewrites `orchestrator.*` → `ado.*` and calls `_warn_legacy_module_name` (base: plain log warning). New `_warn_legacy_module_name` hook method for subclasses to override with richer warnings.
- **`ado/core/operation/config.py`** — `OperatorModuleConf` overrides `_warn_legacy_module_name` to emit a structured deprecation callout pointing users to the 1.x → 2.x migration guide.
- **`ado/core/samplestore/config.py`** — `SampleStoreModuleConf` does the same for samplestore resources.
- **`tests/core/test_module_conf.py`** — New test file covering: prefix rewrite + warning emission, mid-string `orchestrator.` left unchanged, `None` defaulting, and the rich warning path via the samplestore subclass.

### Deprecation window
`moduleName` with an `orchestrator.*` prefix is deprecated from **2.0.0** and will be removed in **3.0.0**.

### DIFF
--- a/ado/core/operation/config.py
+++ b/ado/core/operation/config.py
@@ -190,6 +190,24 @@ def validate_actuator_configuration_ids_against_space_ids(
 class OperatorModuleConf(ModuleConf):
     moduleType: Annotated[ModuleTypeEnum, pydantic.Field()] = ModuleTypeEnum.OPERATION
 
+    @classmethod
+    def _warn_legacy_module_name(cls, old_name: str, new_name: str) -> None:
+        from ado.core.resources import (
+            CoreResourceKinds,
+            warn_deprecated_resource_model_in_use,
+        )
+
+        warn_deprecated_resource_model_in_use(
+            affected_resource=CoreResourceKinds.OPERATION,
+            deprecated_from_ado_version="2.0.0",
+            removed_from_ado_version="3.0.0",
+            deprecated_fields="moduleName",
+            latest_format_documentation_url=(
+                "https://ibm.github.io/ado/migration/1x-to-2x/"
+                "#renamed-python-import-package-orchestrator-ado"
+            ),
+        )
+
     @property
     def operationType(self) -> DiscoveryOperationEnum:
         c: type[ado.modules.operators.base.DiscoveryOperationBase] = (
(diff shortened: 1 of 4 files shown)


christian-pinto (approved):
Works for me, thanks

PR #1217, opened 2026-07-14
feat(cli): support --use-latest for datacontainers in ado describe

### LINKED ISSUE #1209: feat(cli): support --use-latest for everything except experiment in ado describe
## Summary

The `ado describe` command currently supports `--use-latest` only for discovery spaces. Since data containers are also first-class metastore resources, it would be natural to extend this convenience flag to cover them as well — removing an inconsistency in the CLI that forces users to look up a data container identifier manually before describing it.

## Motivation

Historically, `--use-latest` was restricted to discovery spaces because the CLI resolved "latest" from a local file-backed map of resources created. Data containers cannot be created through the CLI, so they were never tracked in that map and the flag was rightfully blocked for them.

The resolution mechanism has since been replaced: the CLI now fetches the identifier of the latest resource directly from the database. This makes the old restriction obsolete — data containers are stored in the same metastore and their latest identifier can be retrieved just as reliably as for any other resource type. Lifting the restriction aligns the CLI with its current capabilities and removes unnecessary friction for users who want to quickly inspect the most recently registered data container.

## Scope

The changes needed are narrow and self-contained:

- **CLI command logic** — the guard that restricts `--use-latest` to spaces should be relaxed so it applies only to resource types where the flag genuinely has no meaning (e.g. experiments), while allowing it for any proper metastore resource such as data containers. Related help text and UI labels should be updated to reflect the broader applicability.
- **Tests** — new test cases should cover the happy path for `ado describe datacontainer --use-latest` and confirm that the flag is still correctly rejected for resource types that do not support it.
- **CLI skill documentation** — the `ado describe` section of the agent skill should be updated to show `--use-latest` working with data containers alongside the existing space example.

AlessandroPomponio:
## Summary

Previously, the `--use-latest` flag in `ado describe` was restricted to discovery spaces only. Any other resource type would produce an error.

This PR broadens support so that `--use-latest` works for data containers, and only rejects it for **experiments** (which have no "latest" concept).

### Changes

**`ado/cli/commands/describe.py`**
- Removed the `SPACE_ONLY_OPTIONS` panel label that isolated `--use-latest`
  as a space-only option.
- Updated the `--use-latest` help text to reflect the wider scope and note
  the experiment exclusion.
- Changed the validation logic: instead of blocking every resource type
  except `DISCOVERY_SPACE`, it now only blocks `EXPERIMENT`.

**`tests/ado/describe/test_ado_describe.py`**
- Added `test_describe_datacontainer_with_use_latest`: verifies that
  `ado describe datacontainer --use-latest` resolves and describes the most
  recently created data container.
- Added `test_describe_use_latest_rejected_for_experiment`: verifies that
  `ado describe experiment --use-latest` exits with code `1`.

### DIFF
--- a/ado/cli/commands/describe.py
+++ b/ado/cli/commands/describe.py
@@ -36,8 +36,6 @@ from ado.modules.actuators.errors import (
 if typing.TYPE_CHECKING:
     from ado.cli.core.config import AdoConfiguration
 
-SPACE_ONLY_OPTIONS = "Space-only options"
-
 
 def describe_resource(
     ctx: typer.Context,
@@ -62,9 +60,9 @@ def describe_resource(
         bool,
         typer.Option(
             "--use-latest",
-            help="Describe the space using the latest space identifier created. "
+            help="Describe the resource using the latest identifier created. "
+            "Not supported for experiments. "
             "Ignored if a resource identifier is also specified.",
-            rich_help_panel=SPACE_ONLY_OPTIONS,
             show_default=False,
         ),
     ] = False,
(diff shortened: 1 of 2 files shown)


mgazz (approved):
LGTM

PR #1228, opened 2026-07-17
feat(core): support fetching entities and entity identifiers for multiple operations at once

### LINKED ISSUE #1227: feat(core): support fetching entities and entity identifiers for multiple operations at once
## Summary

The samplestore's entity-fetching methods currently accept only a single operation ID at a time. It would be worth extending them to accept a collection of operation IDs, so that data across multiple operations can be retrieved in a single database call rather than one round-trip per operation.

## Motivation

At the moment, retrieving entities sampled across several operations requires callers to loop over each operation ID and issue repeated queries. This is both inefficient and unnecessarily complex. Widening the accepted input to allow a set of IDs — while keeping single-ID calls working exactly as today — would remove that loop, reduce database round-trips to one, and make the API more expressive. The change is additive and backwards-compatible, so nothing outside the feature's own call sites should be affected.

## Scope

The following areas of the codebase would likely need to be touched:

- **Samplestore layer** — the concrete SQL implementation of the two entity-fetching methods (`entities_in_operation` and `entity_identifiers_in_operation`) could be updated to accept either a single ID or a set of IDs, routing to an `IN (…)` SQL clause in the multi-ID case.
- **Abstract samplestore interface** — the corresponding abstract method signatures and docstrings should be kept in sync with the concrete implementation.
- **Discovery space** — the delegate method that wraps `entity_identifiers_in_operation`, and the `sampledEntities` logic that today loops over operation IDs, could be simplified to pass the full set in one call once the lower layer supports it.
- **Tests** — new test cases alongside the existing single-operation tests to verify that multi-operation queries return the correct union of entities, with no duplicates.

AlessandroPomponio:
### What changed

Two samplestore methods — `entities_in_operation` and `entity_identifiers_in_operation` — previously only accepted a single operation ID. They now accept either a single ID or a set of IDs, allowing callers to fetch deduplicated entities across multiple operations in **one database query** instead of N queries.

### Key changes

| Layer | Before | After |
|---|---|---|
| `SQLSampleStore` | `entities_in_operation(operation_id: str)` | `entities_in_operations(operation_ids: str \| set[str])` |
| `SQLSampleStore` | `entity_identifiers_in_operation(operation_id: str)` | `entity_identifiers_in_operations(operation_ids: str \| set[str])` |
| `ActiveSampleStore` (base) | abstract `entities_in_operation` | abstract `entities_in_operations` |
| `DiscoverySpace` | called both methods separately in a branch | delegates directly to `entities_in_operations` |

The old single-ID methods are kept as **deprecated shims** (emit `DeprecationWarning`) that forward to the new plural methods, preserving backwards compatibility.

The SQL queries were updated to use `IN :operation_ids` with SQLAlchemy's `expanding=True` bind parameter, which handles the single-or-many case natively.

### Tests

Existing tests were renamed from `*_in_operation` to `*_in_operations` and a new test `test_entities_in_multiple_operations` was added to cover the multi-operation deduplication path end-to-end.

### DIFF
--- a/ado/core/samplestore/sql.py
+++ b/ado/core/samplestore/sql.py
@@ -822,19 +822,23 @@ class SQLSampleStore(ActiveSampleStore):
 
         return list(entities_dict.values())
 
-    def entities_in_operation(self, operation_id: str) -> list[Entity]:
-        """Get entities directly from a single operation in one query.
+    def entities_in_operation(self, operation_ids: str | set[str]) -> list[Entity]:
+        """Get entities directly from one or more operations in one query.
 
-        This method is optimized for the common case of fetching entities from
-        a single operation. It performs the entire operation in a single database
-        query, avoiding the need to first fetch entity IDs and then fetch entities.
+        This method fetches entities from one or more operations in a single
+        database query, avoiding the need to first fetch entity IDs and then
+        fetch entities.
 
         Args:
-            operation_id: The operation identifier to fetch entities for
+            operation_ids: A single operation identifier or a set of operation
+                identifiers to fetch entities for.
 
         Returns:
-            List of Entity objects that were sampled in the specified operation
+            List of Entity objects that were sampled in the specified operation(s)
         """
+        if isinstance(operation_ids, str):
+            operation_ids = {operation_ids}
+
         query = sqlalchemy.text(f"""
             SELECT
                 ent.identifier,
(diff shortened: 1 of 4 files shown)


ado/core/samplestore/sql.py
```diff
@@ -2122,24 +2128,40 @@ def experiments_in_operation(self, operation_id: str) -> list[Experiment]:
             for e in cur
         ]
 
-    def entity_identifiers_in_operation(self, operation_id: str) -> set[str]:
+    def entity_identifiers_in_operation(
```
christian-pinto:
```suggestion
    def entity_identifiers_in_operations(
```

christian-pinto (changes requested):

christian-pinto (approved):

PR #1230, opened 2026-07-17
perf(core): memoize sampledEntities result with optional refresh

### LINKED ISSUE #1229: perf(core): do not recompute DiscoverySpace.sampledEntities by default
## Summary

`DiscoverySpace` should cache the result of fetching sampled entities so that repeated calls within the same instance lifetime no longer re-query the database every time. A simple opt-out mechanism (e.g. a `refresh` flag) would let callers explicitly request fresh data when they need it.

## Motivation

Fetching sampled entities currently hits the database on every invocation — once to resolve linked operations and once to deserialise entities from the sample store. Several callers already invoke this method multiple times within the same request or run, paying the full I/O cost each time with no benefit. Caching the result on the instance after the first fetch would eliminate these redundant round-trips with minimal code change, while a `refresh` escape hatch keeps the door open for callers that genuinely need up-to-date data. As a minor housekeeping improvement, an existing filter that is no longer necessary could be removed at the same time.

## Scope

The changes would be confined to two areas:

- **`DiscoverySpace` implementation** — add an instance-level cache for sampled entities, update the relevant method to populate and respect that cache, and refine its docstring to document the caching semantics and best-effort snapshot caveat.
- **Tests for `DiscoverySpace`** — extend coverage to verify cache-hit behaviour (same object returned on repeated calls) and that passing the refresh flag correctly bypasses the cache.

AlessandroPomponio:
### What changed

`DiscoverySpace.sampledEntities()` previously queried the metastore and sample-store on **every call**. In workflows that call it repeatedly (e.g. operator loops), this produced redundant round-trips to the database with no benefit.

### Changes

**`ado/core/discoveryspace/space.py`**
- Added a `_sampled_entities_cache` instance variable (initialised to `None`).
- `sampledEntities()` now returns the cached list on subsequent calls, skipping both the operation-ID lookup and the sample-store fetch.
- Added a `refresh: bool = False` parameter — pass `True` to force a re-query and update the cache (e.g. after new operations are added).
- Removed the `isEntityInSpace` filter that was already marked as a `TODO` candidate for removal.

**`tests/core/test_space.py`**
- Extended the existing `test_sampled_entities` test to assert:
  - a second call returns the **same list object** (cache hit).
  - a call with `refresh=True` returns a **new list object** (cache
    bypassed and refreshed).

### Impact

- Purely additive / backwards-compatible: existing callers with no `refresh` argument continue to work identically on the first call.
- Reduces database pressure in long-running operator loops.

### DIFF
--- a/ado/core/discoveryspace/space.py
+++ b/ado/core/discoveryspace/space.py
@@ -446,6 +446,9 @@ class DiscoverySpace:
         # Pre-populated by from_operation_id to avoid a redundant DB round-trip.
         self._verified_operation_ids: set[str] = set()
 
+        # Cache for sampledEntities(). None means the cache is cold.
+        self._sampled_entities_cache: list[Entity] | None = None
+
     def __rich__(self) -> "RenderableType":
         """Rich console representation of the DiscoverySpace."""
         import rich.box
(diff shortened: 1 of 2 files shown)


ado/core/discoveryspace/space.py
```diff
@@ -616,20 +619,38 @@ def saveSpace(self) -> None:
                 f"Unable to store space {self._identifier} as no metadata storage provided"
             )
 
-    def sampledEntities(self) -> list[Entity]:
-        """Returns the entities sampled so far in the space"""
+    def sampledEntities(self, refresh: bool = False) -> list[Entity]:
```
michael-johnston, line 622: Fo non-breaking backwards compatibility it seems refresh should be True by default?

ado/core/discoveryspace/space.py
```diff
@@ -616,20 +619,38 @@ def saveSpace(self) -> None:
...
+        """Returns the entities sampled so far in the space.
+
+        The result is cached on the instance after the first call. Subsequent
+        calls return the cached list directly, bypassing both the metastore
+        query for operation IDs and the sample-store entity fetch.
+
+        The cache is a best-effort snapshot: it is **not** automatically
+        invalidated when new operations are added to this space (including
+        writes from other processes). Pass ``refresh=True`` to force a
+        re-fetch and update the cache.
```
michael-johnston, line 632: This is very verbose and can be completely explained by the refresh parameter docs
```suggestion
```

ado/core/discoveryspace/space.py
michael-johnston: Also the point of the method is to 

Returns the entities sampled so far in the space.

So given no parameters it should be expected to do this.

ado/core/discoveryspace/space.py
```diff
@@ -616,20 +619,38 @@ def saveSpace(self) -> None:
...
+
+        Args:
+            refresh: When ``True``, ignore any cached value and re-query the
+                database. The cache is updated with the fresh result.
```
michael-johnston, line 636:
```suggestion
            use_cache: When ``True``, the most recent fetched list of sampled  entities is returned if there is one. The store is not queried. Use if you know nothing has been sampled between calls to this method.  
```

michael-johnston:
>. In workflows that call it repeatedly (e.g. operator loops), this produced redundant round-trips to the database with no benefit.

Which are the workflows alluded to that call it repeatedly, and redundantly?

michael-johnston:
> Purely additive / backwards-compatible: existing callers with no refresh argument continue to work identically on the first call.

Not breaking because the first call is not broken - after that though ... 😆

AlessandroPomponio:
Not worth doing anymore after the improvements to the way we're handling entities and show stats space

PR #1232, opened 2026-07-17
perf(core): enable partial cache reuse in entities_with_identifiers

### LINKED ISSUE #1231: perf(core): allow partial cache reuse in entities_with_identifiers

AlessandroPomponio:
## Partial cache reuse in `entities_with_identifiers`

Previously, `entities_with_identifiers` only used the in-memory entity cache as an all-or-nothing shortcut: if every requested entity was already cached it returned immediately, otherwise it re-fetched **all** requested entities from the database — even ones already in memory.

This PR changes the logic so that the cache is always consulted first and only the **uncached** entities are sent to the database. The results are then merged before returning.

### Changes
- **`ado/core/samplestore/sql.py`** — `entities_with_identifiers` now partitions the requested IDs into cached vs. uncached sets, queries the DB only for uncached IDs, and combines both sets in the return value.
- **`tests/samplestore/get/test_get.py`** — new test `test_entities_by_identifiers_partial_cache_reuse` verifies that a partially warm cache is reused correctly (only the missing entities are fetched from the database).

### Impact
Read performance improves for any caller that requests a mix of already-cached and not-yet-cached entities, since previously cached entities no longer trigger redundant database round-trips.

### DIFF
--- a/ado/core/samplestore/sql.py
+++ b/ado/core/samplestore/sql.py
@@ -744,18 +744,20 @@ class SQLSampleStore(ActiveSampleStore):
             else entity_identifiers
         )
 
-        # Check cache first - if all requested entities are cached, return them
-        if self._entities:
-            cached_entities = [
-                self._entities[entity_id]
-                for entity_id in entity_ids_set
-                if entity_id in self._entities
-            ]
-            # If we got all requested entities from cache, return them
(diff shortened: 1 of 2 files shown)


VassilisVassiliadis (approved):
LGTM

PR #1234, opened 2026-07-17
perf(core): memoize matchingEntities result with optional refresh

AlessandroPomponio:


AlessandroPomponio:
Not worth doing anymore after the improvements to the way we're handling entities and show stats space

Issue #265, opened 2025-11-28
feat(core): require users to explicitly set identifier for resources


AlessandroPomponio:
## Context

Right now, users do not set identifiers for their resources: they create configuration files and run `ado create $RESOURCE -f $FILE` and an identifier is generated for them. Over time, we saw this result in users creating many copies of the same resources (e.g., an actuator configuration), causing confusion.

We should adopt a behaviour similar to Kubernetes:

- Users should provide an identifier for their resource
- We could have an optional field called `generateIdentifier` that allows users to request for a randomly generated identifier based on a prefix

This will also fix the issue of not being able to understand what each space/sample store/etc contains.

## Suggestions/Considerations

- Add the `generateIdentifier` field, supporting strings like `my-space-` to which ado then appends a random string, before popping the field and putting the generated value into `identifier`.
- Add a validator that requires exactly one of {`identifier`, `generateIdentifier`}
- Think about what to do with the `config.metadata.name` field

AlessandroPomponio:
This would also fix the issue where many copies of the same resource end up being created while using the `--with` flag

PR #1248, opened 2026-07-21
feat: multi-input multi-resource type operators

### LINKED ISSUE #646: refactor(operators): create GenericOperatorParameter class and change all parameter classes to inherit from it

### LINKED ISSUE #647: refactor(core): use GenericOperatorParameters instead of dicts

### LINKED ISSUE #926: feat: enable wider range of operator function inputs

Currently `operators` can only operate on one discoveryspace. This is limiting as you might want to operate on 
- multiple spaces
- a specific operation
- a datacontainer

a result is that many analysis operation you need while doing research can't be encoded currently as an operator leading to multiple scripts accumulating locally, particularly with agents.

A related problem is that the operator interface require parameters as a model - this adds a small, but noticeable barrier to the agents as they prefer to write a python function with all the args. 

- operators can be any python function that takes ado resource as types + a parameters model
- when creating an operator you specify the ado resource for each named parameter

Instead of 
```
spaces: 
  - space
```

you can have 
```
spaceA: id1
dc1: id2
```

either at top level or under a new field. Although this means we have to dynamically create the model/fields

- The operator function can have parameters which are lists of ids e.g. `func(spaces:list, operations: list, dcrs: list, params: ParamClass)`
     - This would mean we would not have fields with arbitrary names in the YAML
-  As above and if the operator function had any other parameters we create the parameters model for it dynamically - as long as it's not a dict or custom model e.g. `func(dcrs: list, param1: int=2, param2: float=1.0)` we create model

```
class MyModel(BaseModel):
       param1: int = 2
       param2: float = 1.0
```

With option3 it would be very easy to take any function that's consuming e.g. a datacontainer and turn it into an operator.

michael-johnston:
This PR introduces multi-input multi-resource type operators.

It also changes the type of the operator functions `parameters` parameter to be instances of the operators parameter model instead of a dictionary (answers #647) 

Example new function definition.

```python
def operator_function(my_param_one: OperatorInputType, 
                 another_resource_param: OperatorInputType, 
                 operationInfo: FunctionOperationInfo=None, 
                 parameters:GenericOperatorParameters= GenericOperatorParameters())
  ...
```

Note: Here we use `OperatorInputType` to denote the allowed input types e.g. DiscoverySpace (see more below) and GenericOperatorParameters as a placeholder for the specific parameter model of the operator

Prior to this PR operators could only work on a single input resource, which had to be of type DiscoverySpace and named discoverySpace. Further the parameters for the operator were passed as keyword args i.e. 

```python
operator_fn(discoverySpace:DiscoverySpace, operationInfo:FunctionOperationInfo=None, **kwargs)
```

### Operator Input Types

ADO resources that can be operated on are now associated with a "rich" type that will be passed to the operator. In the PR we have two resources that can be operated on 

- DiscoverySpaceResource -> DiscoverySpace (rich type) <- As it is now
- DataContainerResource -> DataContainerResource (rich type) <- New resource type and associated rich type

Note: Its a particular property of DataContainerResource that it can be its own rich type. 

### Setting inputs in operation.yaml

The PR removes the DiscoveryOperationConfiguration `spaces` field and introduce the more general `input` field. 

before 

```yaml
spaces:  # A list of resource ids
    - space_id_one
    - space_id_two # Note: technically you could have more than one id, but nothing could use it
```

now, for an operator with two DiscoverySpace input params called space_one and space_two

```yaml
inputs: 
    space_one: 
          identifier: space_id_one
          kind: DiscoverySpace
    space_two:
          identifier: space_id_two
           kind: DiscoverySpace
```
        
In practice `kind` can be inferred when the YAML is read so users can omit it. 

For existing operators that previously had a YAML section like 

```yaml
spaces: 
    - space-1234-abcd
```    

Users will now set 
```yaml
inputs: 
   discoverySpace:
        identifier: space-1234-abcd
```

### Non-breaking changes

- The `spaces` field is kept as a computed field, so any code relying on querying it in metastore does not need to be changed
- All older operation resources are updated to have the new `input` field if read. Not breaking
- All existing operators resource parameters i.e. discoverySpace: DiscoverySpace, is supported out-of-box. No change needed
- If users pass the `spaces` field in YAML it will be auto-updated to new format. This will only be valid for operators talking a single space with a parameter called discoverySpace (all existing operators)

### Breaking changes

- The parameters field of all operators has been updated to be an instance of the operator class not keyword args. Old operators (pre this change) cannot be used with this change. Scripts or operators calling operator functions directly also must be updated. 

### Code Changes

The code changes fall into three groups:

1. Updates to allow operators to have multiple resource inputs of multiple types and for operations to store them (new input field, models for that field, validators are parsers for operator signature etc.)
2. Updates in orchestration flow to handle multiple inputs (instead of just looking for `spaces[0]`), including storing relationships, and handling cases where the rich resource type does not provide a way to acces the metastore
3. Updates due to change to use Operators parameter class and introduction of GenerericOperatorConfiguration (changing signatures, changing plugin versions etc).


michael-johnston:
@AlessandroPomponio docs to be updated once the implementation is deemed good enough.

ado/cli/resources/operation/create.py
```diff
@@ -36,16 +37,118 @@
...
+    Returns:
+        The input parameter identifier to set.
+
+    Raises:
+        ValueError: If the operator has no matching input, or multiple matching
+            inputs are unset / ambiguous.
+    """
+    from ado.modules.operators.collections import operator_metadata_for_reference
+
+    module = (operation_data.get("operation") or {}).get("module") or {}
```
AlessandroPomponio:
```suggestion
    module = operation_data.get("operation", {}).get("module", {})
```

ado/cli/resources/operation/create.py
```diff
@@ -36,16 +37,118 @@
...
+
+    existing_inputs = operation_data.get("inputs") or {}
+    unset = [d.identifier for d in matching if d.identifier not in existing_inputs]
+    if len(unset) == 1:
+        return unset[0]
+    names = [d.identifier for d in matching]
+    raise ValueError(
+        f"Operator {metadata.name!r} has multiple {kind.value!r} inputs "
+        f"{names}; set them explicitly in the operation YAML."
+    )
```
AlessandroPomponio:
```suggestion
    required_resource_inputs = metadata.required_resource_inputs
    inputs_of_kind = [i for i in required_resource_inputs if i.kind == kind]
    if not inputs_of_kind:
        raise ValueError(
            f"Operator {metadata.name!r} has no resource input of kind {kind.value!r}."
        )
    if len(inputs_of_kind) == 1:
        return inputs_of_kind[0].identifier

    present_inputs = operation_data.get("inputs", {})
    missing_inputs = [
        i.identifier for i in inputs_of_kind if i.identifier not in present_inputs
    ]
    if len(missing_inputs) == 1:
        return missing_inputs[0]
    input_identifiers_of_kind = [i.identifier for i in inputs_of_kind]
    raise ValueError(
        f"Operator {metadata.name!r} has multiple {kind.value!r} inputs "
        f"{input_identifiers_of_kind}; set them explicitly in the operation YAML."
    )
```

ado/cli/resources/operation/create.py
```diff
@@ -36,16 +37,118 @@
...
+        operation_data: Raw operation resource configuration dict (YAML).
+        kind: Resource kind to bind (e.g. discoveryspace, datacontainer).
+
+    Returns:
+        The input parameter identifier to set.
+
+    Raises:
+        ValueError: If the operator has no matching input, or multiple matching
+            inputs are unset / ambiguous.
+    """
```
AlessandroPomponio, line 59: I find this function to be a bit weird: it returns the identifier of a parameter of kind `kind` if:
- It's the only one matching the kind (no check on whether it was set or not)
- It's the only one unset matching the kind

What I mean with this is that this function is very specific but the name doesn't really convey it. This part of the docstring does a lot of heavy lifting:
> When several inputs share *kind*, prefers an input not yet present in ``operation_data["inputs"]``.

This function is used only in one place for now and it's private, so I'm not sure if I'd keep this separate, given how many caveats it has to be understandable

ado/cli/resources/operation/create.py
```diff
@@ -224,9 +314,15 @@ def reuse_requested_latest_identifiers(
             )
             raise typer.Exit(1)
 
-        # Set the appropriate field on the configuration
-        field_name = resource_field_mapping[resource_kind]
-        setattr(resource_configuration, field_name, [latest_id])
+        if resource_kind == CoreResourceKinds.ACTUATORCONFIGURATION:
+            operation_data["actuatorConfigurationIdentifiers"] = [latest_id]
```
AlessandroPomponio, line 320: Should actuator configurations still have their own handling at this point? I understand they would be a different type of "input", but they are still an input in the sense that they're passed to the actuator. This might also simplify the model a bit and keep it more consistent

ado/cli/resources/operation/create.py
```diff
@@ -191,31 +229,83 @@ def create_operation(parameters: AdoCreateCommandParameters) -> str | None:
...
+    if CoreResourceKinds.DATACONTAINER in parameters.with_resources:
+        dc_value = parameters.with_resources[CoreResourceKinds.DATACONTAINER]
+        if not isinstance(dc_value, str):
+            console_print(
+                f"{ERROR}--with datacontainer currently supports an existing "
+                "identifier only (not a YAML file).",
+                stderr=True,
+            )
+            raise typer.Exit(1)
+        _set_input_reference(operation_data, CoreResourceKinds.DATACONTAINER, dc_value)
```
AlessandroPomponio: The shorthand for Datacontainers in the cli is `dcr` and also this is an identifier and not a value
```suggestion
    if CoreResourceKinds.DATACONTAINER in parameters.with_resources:
        dcr_identifier = parameters.with_resources[CoreResourceKinds.DATACONTAINER]
        if not isinstance(dcr_identifier, str):
            console_print(
                f"{ERROR}--with datacontainer currently supports an existing "
                "identifier only (not a YAML file).",
                stderr=True,
            )
            raise typer.Exit(1)
        _set_input_reference(operation_data, CoreResourceKinds.DATACONTAINER, dcr_identifier)
```

ado/cli/resources/operation/create.py
```diff
@@ -191,31 +229,83 @@ def create_operation(parameters: AdoCreateCommandParameters) -> str | None:
     return output_operation_result(result=operation_output)
 
 
-def reuse_requested_latest_identifiers(
-    resource_configuration: DiscoveryOperationResourceConfiguration,
+def _apply_with_resources(
```
AlessandroPomponio, line 234: With multiple inputs of the same type being possible, this might need changing: as it stands, `--with` only supports one identifier/resource file

ado/cli/resources/operation/create.py
```diff
@@ -36,16 +37,118 @@
...
+    if len(unset) == 1:
+        return unset[0]
+    names = [d.identifier for d in matching]
+    raise ValueError(
+        f"Operator {metadata.name!r} has multiple {kind.value!r} inputs "
+        f"{names}; set them explicitly in the operation YAML."
+    )
+
+
+def _set_input_reference(
```
AlessandroPomponio, line 97: Maybe a better name for this would be `_add_operation_input_of_kind`

ado/cli/resources/operation/create.py
```diff
@@ -36,16 +37,118 @@
...
+    operation_data: dict,
+    kind: CoreResourceKinds,
+    identifier: str,
```
AlessandroPomponio, line 100: I see we have a model for this: ADOResourceReference, we should probably use it

ado/cli/resources/operation/template.py
```diff
@@ -53,8 +53,15 @@ def template_operation(parameters: AdoTemplateCommandParameters) -> None:
...
     if not parameters.operator_name:
+        from ado.core.resources import ADOResourceReference, CoreResourceKinds
+
         model_instance = DiscoveryOperationResourceConfiguration(
-            spaces=["your-spaces"],
+            inputs={
+                "discoverySpace": ADOResourceReference(
+                    identifier="your-discoveryspace-id",
+                    kind=CoreResourceKinds.DISCOVERYSPACE,
+                )
```
AlessandroPomponio, line 63: When I saw `discoverySpace` I thought this was going to be a required key, but after reading again I understand it's just a placeholder name.

This, however, made me think: why don't we have something like

- inputs
  - CoreResourceKinds.value
    - identifier
    - name

instead of:
- inputs
  - name
    - identifier
    - kind

This seems more structured to me, as it would group inputs by kind. The only "con" I can imagine is that I assume the name is required and it has to do with the signature, although this isn't really clear from the "name" name

ado/core/discoveryspace/space.py
```diff
@@ -1018,7 +1028,7 @@ def operation_context(
                 parameters={},
             ),
             metadata=config_metadata,
-            spaces=[self.uri],
+            inputs={"discoverySpace": self.reference},
```
AlessandroPomponio, line 1031: I don't think we should use `discoverySpace` as the default placeholder - we should use `CoreResourceKinds.DISCOVERYSPACE.value` (i.e., `discoveryspace`)

ado/core/operation/config.py
```diff
@@ -27,6 +31,24 @@
 
 if typing.TYPE_CHECKING:
     import ado.modules.operators.base
+    from ado.metastore.sqlstore import SQLStore
+
+#: Represents the default typed input parameter used for operators that work on a single discovery space.
+# i.e. a parameter `discoverySpace: DiscoverySpace`
+_DEFAULT_DISCOVERY_SPACE_INPUT_PROPERTY = ADOResourcePropertyDescriptor(
+    identifier="discoverySpace",
```
AlessandroPomponio, line 39: See comment above - I think we should use `discoveryspace`

ado/core/operation/config.py
```diff
@@ -27,6 +31,24 @@
...
+
+
+class GenericOperatorParameters(pydantic.BaseModel):
+    """Base class for operator parameter (configuration) models.
+
+    Operator-specific parameter classes should subclass this. Schemas are
+    closed: unknown fields are rejected (``extra="forbid"``).
+    """
+
+    model_config = ConfigDict(extra="forbid")
```
AlessandroPomponio, line 51: In GenericActuatorParameters I left this comment:

```
    # AP 5/11/24:
    # We allow extras because this model will be inherited
    # by actuator-specific models and will be used when
    # deserializing. If this wasn't the case, we would
    # ignore all the values
    model_config = ConfigDict(extra="allow")
```

Maybe I was wrong, but we should then investigate

ado/core/operation/config.py
```diff
@@ -295,6 +317,26 @@ class OperatorMetadata(pydantic.BaseModel):
...
+    ]
+    required_properties: Annotated[
+        list[str] | None,
+        pydantic.Field(
+            default=None,
+            description=(
+                "Target property identifiers this operator reads from a discovery space if any. "
+            ),
+        ),
+    ]
```
AlessandroPomponio: Should this field contain a callout to discovery space?

```suggestion
    required_properties: Annotated[
        list[str] | None,
        pydantic.Field(
            description=(
                "Target property identifiers this operator reads from a discovery space if any. "
            ),
        ),
    ] = None
```

ado/core/operation/config.py
```diff
@@ -569,13 +611,13 @@ class DiscoveryOperationResourceConfiguration(pydantic.BaseModel):
     actuatorConfigurationIdentifiers: Annotated[
         list[str], pydantic.Field(default_factory=list)
     ]
-    spaces: Annotated[
-        list[str],
+    inputs: Annotated[
+        dict[str, ADOResourceReference],
```
AlessandroPomponio, line 614: See comments above on possibly keying this by CoreResourceKinds.value

ado/core/operation/config.py
```diff
@@ -585,50 +627,227 @@ class DiscoveryOperationResourceConfiguration(pydantic.BaseModel):
         },
     )
 
+    @pydantic.computed_field  # type: ignore[prop-decorator]
+    @property
+    def spaces(self) -> list[str]:
```
AlessandroPomponio, line 631: According to https://pydantic.dev/docs/validation/latest/concepts/fields/#the-computed_field-decorator this will then be serialized - is this the behaviour we want?

ado/core/operation/config.py
```diff
@@ -655,3 +874,9 @@ class FunctionOperationInfo(pydantic.BaseModel):
             description="The namespace the operation should create ray workers/actors in"
         ),
     ] = None
+    projectContext: Annotated[
+        ProjectContext | None,
+        pydantic.Field(
+            description=("Project this operation runs in."),
+        ),
+    ] = None
```
AlessandroPomponio, line 881: This is very dangerous - are we sure this is never serialized?

ado/core/operation/config.py
```diff
@@ -585,50 +627,227 @@ class DiscoveryOperationResourceConfiguration(pydantic.BaseModel):
...
+        data = dict(data)
+        # We need to remove spaces as pydantic does not handle
+        # the situation of computed fields with extra=forbid
+        spaces = data.pop("spaces")
+        if "inputs" not in data and spaces:
+            data["inputs"] = {"discoverySpace": {"identifier": spaces[0]}}
+        return data
+
+    @pydantic.model_validator(mode="after")
+    def validate_inputs(self, info: pydantic.ValidationInfo) -> Self:
```
AlessandroPomponio, line 686: This is very long - can we maybe split it up somehow?

ado/core/operation/context.py
```diff
@@ -0,0 +1,102 @@
...
+        raise ValueError(f"Operator inputs disagree on project context: {details}.")
+
+    return carried[0][1]
+
+
+def assert_inputs_in_metastore(
+    inputs: dict[str, OperatorInputType],
+    metastore: SQLStore,
+) -> None:
+    """Assert that every operator input resource belongs to *metastore*.
```
AlessandroPomponio, line 71: I wouldn't call this assert since we don't do assertions in non-test code. I'd call this `verify_inputs_exist_in_metastore`, although i'm not sure if this function belongs here or in the metastore code

ado/core/operation/inputs.py
```diff
@@ -0,0 +1,264 @@
...
+    DiscoverySpace: lambda resource: resource.project_context,  # type: ignore[attr-defined,return-value]
+}
+
+
+def _unwrap_annotated(hint: object) -> object:
+    """Return the underlying type if *hint* is ``Annotated[T, ...]``, else *hint*."""
+    if get_origin(hint) is Annotated:
+        args = get_args(hint)
+        return args[0] if args else hint
+    return hint
```
AlessandroPomponio, line 63: This function seems generic enough, I'd put it in `ado/utilities/annotations.py` or something

ado/core/operation/inputs.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+def _resolve_datacontainer(
+    reference: ADOResourceReference, metastore: SQLStore
+) -> DataContainerResource:
+    return metastore.getResource(  # type: ignore[return-value]
+        identifier=reference.identifier,
+        kind=CoreResourceKinds.DATACONTAINER,
+        raise_error_if_no_resource=True,
+    )
```
AlessandroPomponio, line 210: Do we need these?

ado/modules/operators/_explore_orchestration.py
```diff
@@ -323,13 +324,16 @@ def finalize_callback(operation_resource: OperationResource) -> None:
 
         return finalize_callback
 
+    explore_inputs = {"discoverySpace": discovery_space}
```
AlessandroPomponio, line 327: same comment about discoveryspace

ado/modules/operators/base.py
```diff
@@ -45,35 +49,40 @@
...
 
-def validate_operator_function_signature(fn: typing.Callable) -> None:
-    """Validate that *fn* conforms to the :class:`OperatorFunction` Protocol.
 
-    Validates the positional parameter count against the Protocol, then
-    checks that every positional parameter and the return value carry a type
-    annotation whose type matches the Protocol.  Type hints are mandatory:
-    an operator function without them cannot be verified to be correct, so
-    a ``ValueError`` is raised rather than silently skipping the check.
+def validate_operator_call_shape(
```
AlessandroPomponio, line 65: I think there are multiple similar function checking parameters etc - is there no way to make them (partially) reusable?

AlessandroPomponio (changes requested):
Initial set of comments

ado/cli/resources/operation/create.py
michael-johnston: --with does need to change

ado/cli/resources/operation/create.py
michael-johnston: Maybe - Have to think about it a bit.

ado/cli/resources/operation/template.py
michael-johnston: As you say this is to do with the signature. 

This structure makes its the same order as function parameters, and most like calling the function

```python
def my_op(referenceSpace: DiscoverySpace, comparisonSpace: DiscoverySpace)
```

`calling` this becomes; 

```yaml
- referenceSpace: 
        value: space-1345-abcd
- comparisonSpace:
        value: space-4567-abcd
```

Note "kind" is not required to be added by user as on creation we can discover it from the signature. It's a convenience parameter for storing the kind so later when making metastore quires we don't have to look it up. (since our queries require the kind)

If we do it the other way we make the user have to write the "CoreResourceKinds.value" and manually group the parameters by that.

ado/core/discoveryspace/space.py
michael-johnston: ok. Here its not a default btw, this is for script operators.

ado/core/operation/config.py
michael-johnston: This is badly named. Its not a "default" 

It's for backwards compatibility. Better name would be "DISCOVERY_SPACE_INPUT_PROPERTY_FOR_OLDER_OPERATORS"

not the older operators are not legacy as how they are written is completely valid for new code. It's just at the time they *had* to call their first parameter `discoverySpace. `

ado/core/operation/config.py
michael-johnston: I haven't seen this problem in any of the tests. I assumed the subclass was used to deserialise.

ado/core/operation/config.py
michael-johnston: Yes. This allows querying metastore by "spaces" still. 

The behaviour we would prefer is on deserialising that computed fields are ignored automatically by pyndatic - however they don't do this. There are various issues discuss why and advocating for/against.

ado/core/operation/config.py
michael-johnston: 🤷 

Maybe we can move out the script operator part - it's the balance between length and creating functions that are only called from one place.

ado/core/operation/config.py
michael-johnston: Yes, this is not a model for serialising. It's for passing data around. 

There is a related (but not to this PR) issue of whether we should allow operators access to the database, and by extension the credentials. They can get this already via DiscoverySpace.

ado/core/operation/context.py
michael-johnston: ok

ado/core/operation/inputs.py
michael-johnston: its just to simplify the code getting the rich types for each object so it doesn't expose the different ways the rich types are constructed e.g. that code just does

```yaml
rich_object = resolve_$TYPE($RESOURCE_REFERENCE)
```

and then its easier to read

ado/modules/operators/base.py
michael-johnston: They check different things. This is checking the signature of a function is as required. (**signature checking**). Only requires the function signature

i.e. the operator function you wrote has signature X, is this the correct structure?

Other checks are that the type of the data requested to be passed to a function matches the types (**argument type checking**). Requires function signature and list of type you want to pass to it (static)

i.e your operation YAML want to pass resource sA,B,C of type X,Y,Z to operator MYOP - does this match what the operator wants

A mother validation is at that the actual resource you asked for exist (**data existence checking**) Requires function signature, list of types, and database.

your operation YAML wants to get resource with ids A,B,C - do these actually exist in the db if the project you are trying to run in.

ado/modules/operators/_explore_orchestration.py
michael-johnston: I don't think this is actually required here.

ado/cli/resources/operation/template.py
AlessandroPomponio: I would still prefer that, especially since there aren't expressivity guarantees from parameter names or identifiers (ref https://github.com/IBM/ado/issues/265)

ado/core/operation/config.py
AlessandroPomponio: Okay, maybe something changed. I remember I had to enable duck typing serialization as well

ado/cli/resources/operation/template.py
michael-johnston: not convinced this is a problem :-)

AlessandroPomponio: It's just my opinion - we should at least ask the others what they prefer

ado/core/operation/config.py
AlessandroPomponio: I see, this would be where a different structure of the inputs dictionary could be more convenient when querying

ado/core/operation/config.py
AlessandroPomponio: Unless this ends up causing issues with Ray, I would at least add an exclude to prevent accidental serializations: https://pydantic.dev/docs/validation/latest/concepts/fields/#excluding-fields

ado/core/operation/inputs.py
AlessandroPomponio: Should we maybe have a `from_identifier` method in the various resources?

ado/cli/resources/operation/create.py
michael-johnston: So the issue here is these changes are about operator plugin inputs - the inputs that go to a function that someone writes. 

actuator configurations while passed to the operation to set up actuators (when there are actuators) aren't operator plugin inputs. They are not passed to the plugin. 

At the moment I can't think of a sensible way to combine the two things so actuator configurations should still be considered special.

ado/core/operation/inputs.py
michael-johnston: This actually becomes a bit messy as it means the resources need to know about the rich type, and the normal flow of definitions is resource -> rich type (c.f. ADODiscoverySpaceResource -> DiscoverySpace) i.e. it would lead to circular imports

ado/cli/resources/operation/create.py
michael-johnston: This exists due to the current implementation of `--with

before changing lets figure out how with will work.

PR #1252, opened 2026-07-21
perf(core): skip matchingEntities() call in space stats when space is fully sampled

### LINKED ISSUE #1251: perf(core): skip matchingEntities() call when space is fully sampled

AlessandroPomponio:
## perf: skip `matchingEntities()` when space is fully sampled in stats

In `space_statistics_for_spaces`, the call to `space.matchingEntities()` — which is comparatively expensive — is now skipped when the space has been fully sampled. When all entities in the space have been measured (`len(sampledEntities()) == size_of_entity_space`), sampled entities are identical to matching entities, so the result is reused directly. The `sampledEntities()` call is moved up to make it available for this check.

### DIFF
--- a/ado/core/discoveryspace/stats.py
+++ b/ado/core/discoveryspace/stats.py
@@ -318,7 +318,17 @@ def space_statistics_for_spaces(
         else:
             number_unmeasured = size_of_entity_space - number_measured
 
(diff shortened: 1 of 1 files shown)


mgazz (approved):
LGTM

PR #1259, opened 2026-07-22
feat(core): support per-operation grouping when querying entity identifiers

### LINKED ISSUE #1258: feat(core): support per-operation grouping when querying entity identifiers
## Summary

It would be useful if querying for entity identifiers across one or more operations could optionally return results grouped by operation, rather than always returning a flat set of all identifiers combined.

## Motivation

Currently, callers that need to know which entities belong to which operation must perform that grouping themselves after the fact, often repeating logic or making additional queries. Supporting a grouped return mode natively would make this pattern simpler and less error-prone, while keeping the existing flat behaviour intact for all current callers.

## Scope

The change would touch the sample store querying layer and the discovery space interface that delegates to it. Both would gain an optional parameter controlling whether results are returned as a flat set or as a mapping from operation to its set of entity identifiers. Corresponding tests should be added to cover the new grouped return path, including multi-operation scenarios.

AlessandroPomponio:
## Add `group_by_operation` flag to entity identifier queries

Adds an optional `group_by_operation: bool = False` parameter to `SQLSampleStore.entity_identifiers_in_operations` and `DiscoverySpace.entity_identifiers_in_operations`.

When `False` (default), behaviour is unchanged — a flat `set[str]` of entity IDs is returned across all specified operations. When `True`, a `dict[str, set[str]]` is returned mapping each operation ID to its own set of entity identifiers.

The SQL query is also refactored from a raw `sqlalchemy.text` call to a type-safe SQLAlchemy Core `select` statement, and the result-processing loop now drives both return modes from a single pass over the rows. A new test (`test_entity_identifiers_in_operations_grouped`) covers the grouped path for multiple operations and for a single string operation ID.

### DIFF
--- a/ado/core/discoveryspace/space.py
+++ b/ado/core/discoveryspace/space.py
@@ -1061,11 +1061,27 @@ class DiscoverySpace:
 
     @_perform_preflight_checks_for_sample_store_methods
     def entity_identifiers_in_operations(
-        self, operation_ids: str | set[str]
-    ) -> set[str]:
-        """Return entity identifiers sampled in the given operation(s)."""
+        self,
+        operation_ids: str | set[str],
+        group_by_operation: bool = False,
+    ) -> set[str] | dict[str, set[str]]:
+        """Return entity identifiers sampled in the given operation(s).
+
+        Args:
+            operation_ids: A single operation identifier or a set of operation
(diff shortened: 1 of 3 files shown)


christian-pinto (approved):
LGTM

PR #1261, opened 2026-07-22
fix(vllm-performance): add USER env var to prevent uid not found on OpenShift

mgazz:
## Problem

When deploying vllm on OpenShift or clusters with restricted SCCs, the container
runs with a dynamically assigned UID. This causes
a `KeyError: 'getpwuid(): uid not found: 1000770000'` error at startup.

## Fix

Add `USER=vllm` as an environment variable in the vllm container deployment
template. This satisfies the `getpwuid()` lookup.


AlessandroPomponio (approved):
LGTM thanks

PR #1263, opened 2026-07-22
fix(core): make experiment catalog query reliable and cache results

### LINKED ISSUE #1262: fix(core): make experiment catalog query reliable and cache results
## Summary

The current implementation of the experiment catalog lookup in the SQL sample store has a fragile query strategy that may return incorrect or incomplete results depending on which entity happens to be selected first. Additionally, the result is recomputed on every call with no caching, and the same query logic is duplicated in more than one place.

## Motivation

Because the catalog drives experiment registration and replay behaviour, an unreliable lookup can silently produce wrong results — for example missing experiments that are present in the store but not associated with the first entity retrieved. The lack of caching means unnecessary database round-trips on repeated calls, and duplicated query logic makes the code harder to maintain and leaves two places where the same bug can recur.

## Scope

The following areas should be considered for improvement:

- **Catalog query strategy** — the query should be driven by measurement results rather than by an arbitrary entity, so that all distinct replay experiments in the store are correctly identified regardless of data ordering.
- **Result caching** — the computed catalog should be cached at the instance level and invalidated only when replay results are written, avoiding redundant database queries on repeated calls.
- **Code deduplication** — the secondary entry-point that reads the catalog from a store reference should delegate to the primary implementation rather than duplicating the query logic.
- **Tests** — existing tests should be updated to reflect the corrected, deterministic behaviour, and new tests should cover the case where the first entity in the store has no replay results, as well as the caching guarantee.

AlessandroPomponio:
## Fix: reliable experiment catalog for sample stores

The experiment catalog built from a SQL sample store could return incorrect or incomplete results because it queried only a single arbitrary entity (`LIMIT 1`). If that entity had no replay measurement result, the catalog would be wrong or empty.

This PR rewrites the query to scan only entities that actually have replay measurement results, builds the catalog from all of them, and caches the result on the store instance to avoid redundant database hits. Cache invalidation is handled automatically when new entities or replayed measurements are added.

Tests are added to cover the previously broken scenario and verify caching behaviour. A stale comment describing the old non-deterministic behaviour is removed.

---

*This code was generated using [IBM Bob](https://bob.ibm.com/) and was reviewed manually.*

### DIFF
--- a/ado/core/samplestore/sql.py
+++ b/ado/core/samplestore/sql.py
@@ -72,6 +72,11 @@ if TYPE_CHECKING:
 _source_tables_verified: set[tuple[str, str]] = set()
 _reflected_metadata_cache: dict[tuple[str, str], sqlalchemy.MetaData] = {}
 
+# Sentinel for the instance-level experiment-catalog cache.
+# Using a distinct object lets us cache ``None`` (empty store) without
+# confusing it with "not yet computed".
+_UNSET = object()
+
 
 class SQLSampleStoreConfiguration(pydantic.BaseModel):
     identifier: Annotated[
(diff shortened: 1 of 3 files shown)


AlessandroPomponio:
As a side-note, this PR slashes the time it takes to initialise the DiscoverySpaces for the cplex_mip context when running `ado show stats spaces` from ~130s down to ~12s

michael-johnston (approved):

PR #1268, opened 2026-07-23
perf(core): do not load entity measurements when computing space stats

### LINKED ISSUE #1114: perf(core): improve performance of matchingEntities
## Summary

The `matchingEntities` method currently fetches every entity from the sample store and checks each one against the discovery space in Python. When many discovery spaces share a single sample store this becomes expensive, and performance degrades roughly proportionally to the number of spaces times the total number of entities. It would be worth revisiting this approach to make better use of information that is already available in the system.

## Motivation

As sample stores grow to tens of thousands of entities shared across multiple discovery spaces, the current linear-scan approach becomes a meaningful bottleneck. The system already holds space-hierarchy relationships between discovery spaces (sub-space, equal-space, super-space), which could be used to categorise sibling spaces' entities — eliminating the need to check every entity individually. Leveraging these relationships could reduce the number of per-entity checks to a much smaller candidate subset, with no changes to the public API or any other subsystem.

## Scope

- The core logic for retrieving entities matching a discovery space could be updated to query sibling spaces sharing the same sample store, classify them by their spatial relationship to the current space, and only run per-entity checks on the subset of entities that genuinely require them.
- Test coverage for the matching-entities behaviour would benefit from cases that exercise the different space-hierarchy relationships (sub-space, super-space, undefined overlap, no siblings), ensuring correctness is preserved and regressions are caught.

### LINKED ISSUE #1239: perf(core): use db-based method to count measured entities with measurements for a set of experiments

### LINKED ISSUE #1267: perf(core): avoid loading measurement blobs when computing space statistics
## Summary

The `ado show stats space` command currently loads full measurement result blobs into memory as part of computing entity statistics, even though the statistics themselves only require knowing *which* experiments each entity has results for — not the measurement values. This issue proposes a targeted optimisation to eliminate that unnecessary data loading.

## Motivation

Space statistics are computed by iterating over sampled and matching entities and inspecting their measurement results in Python. Because the current retrieval path loads the full measurement JSON alongside each entity, computing even simple counts causes significant unnecessary I/O and memory pressure — especially in large spaces. The same counts can be derived entirely from lightweight SQL projections that never deserialise measurement blobs.

## Scope

The changes suggested by this issue touch three areas:

- **SampleStore interface and SQL implementation**: introduce SQL-backed methods that answer measurement-presence and experiment-coverage questions for a set of entity identifiers, without loading measurement result blobs. The SQL layer already has the building blocks needed; these would be new, focused query paths.
- **DiscoverySpace entity loading**: the methods used to retrieve sampled and matching entities could be made optionally blob-free, so that callers in the stats path can get entity constitutive properties without triggering measurement fetching. Existing callers should remain unaffected.
- **Space statistics computation**: the statistics function should be updated to use the new lightweight paths, replacing the current Python-side loops that depend on loaded measurement data with set arithmetic over entity identifiers and the results of the new SQL methods.

AlessandroPomponio:
## Summary

Eliminates full measurement-blob loading when computing discovery space statistics (`ado get stats`). Previously, the stats path hydrated every entity's measurement results (full JSON blobs) just to count measured entities and check experiment references. The change fetches only the data actually needed — entity identifiers and experiment references — via targeted SQL queries, reducing memory pressure and query cost, especially for large sample stores.

## High-level Changes

- **`ado/core/samplestore/base.py` / `sql.py`**: Replaced the monolithic `space_entity_statistics()` method with two focused, composable primitives — `entity_identifiers_in_operations()` (returns entity IDs grouped by operation) and `entity_experiment_references()` (returns experiment references per entity from valid results only, without deserialising measurement blobs). Added `entities_with_valid_measurements()` as an additional lightweight helper.
- **`ado/core/discoveryspace/stats.py`**: Refactored `space_statistics_for_spaces()` into a two-pass approach: Pass 1 fetches entities without loading measurements (`require_measurements=False`) and accumulates entity IDs; a single batched `entity_experiment_references()` call per sample store follows; Pass 2 distributes the batch result to each space in pure Python. `lightweight_space_statistics()` now accepts pre-computed `entity_ids_by_space_id` instead of taking a `SampleStore` and issuing its own query.
- **`ado/core/discoveryspace/space.py`**: `sampledEntities()` and `matchingEntities()` gained a `require_measurements` keyword argument, allowing callers to opt out of loading measurement blobs when only constitutive properties are needed.
- **`ado/cli/utils/resources/formatters.py`**: Merged two separate metastore traversals (child operations + parent sample stores) into a single bidirectional `get_resources_by_relationship` call, then hydrates sample stores with a single `getResources` batch call instead of per-space hydration.
- **Tests**: Replaced `test_space_entity_statistics_*` tests with equivalent coverage for the new `entities_with_valid_measurements` and `entity_experiment_references` APIs.

## Impact

`ado get stats` for spaces with large sample stores will use significantly less memory and issue fewer/lighter SQL queries. No user-facing behaviour changes — statistics values are identical. The internal `space_entity_statistics()` API has been removed and replaced; any out-of-tree code calling it directly will need to be updated.

---

> **AI Disclosure:** This code and this PR description were generated using [IBM Bob](https://bob.ibm.com/) and were reviewed manually.

### DIFF
--- a/ado/cli/utils/resources/formatters.py
+++ b/ado/cli/utils/resources/formatters.py
@@ -413,52 +413,44 @@ def format_ado_get_stats_for_spaces(
 
     space_ids: set[str] = set(df["IDENTIFIER"])
 
-    # Query 1: for each space, get its child operations.
-    # Returns {space_id: {CoreResourceKinds.OPERATION: set[operation_id]}}
-    space_child_relationships: dict[str, dict[CoreResourceKinds, set[str]]] = (
+    # Single traversal in both directions (identifiers only) to get child
+    # operation IDs (down) and parent samplestore IDs (up) in one SQL query.
+    # Returns {space_id: {CoreResourceKinds.OPERATION: set[str],
+    #                      CoreResourceKinds.SAMPLESTORE: set[str]}}
+    space_relationships: dict[str, dict[CoreResourceKinds, set[str]]] = typing.cast(
+        "dict[str, dict[CoreResourceKinds, set[str]]]",
         sql_store.get_resources_by_relationship(
             kind=CoreResourceKinds.DISCOVERYSPACE,
             identifier=space_ids,
-            hierarchy_direction="down",
+            hierarchy_direction="both",
             max_hops=1,
             identifiers_only=True,
-        )
+        ),
     )
 
-    # Build space_id_to_operation_ids: {space_id: set[operation_id]}
-    space_id_to_operation_ids: dict[str, set[str]] = {}
-    for space_id in space_ids:
-        space_children_by_kind = space_child_relationships.get(space_id, {})
-        space_id_to_operation_ids[space_id] = space_children_by_kind.get(
-            CoreResourceKinds.OPERATION, set()
+    # Collect deduplicated samplestore IDs across all spaces, then hydrate once.
+    all_samplestore_ids: set[str] = {
+        samplestore_id
+        for space_id in space_ids
+        for samplestore_id in space_relationships.get(space_id, {}).get(
+            CoreResourceKinds.SAMPLESTORE, set()
         )
-
-    # Query 2: for each space, get its parent sample store(s),
-    # returning hydrated SampleStoreResource objects.
-    # Returns {space_id: {CoreResourceKinds.SAMPLESTORE: {samplestore_id: SampleStoreResource}}}
-    space_parent_relationships: dict[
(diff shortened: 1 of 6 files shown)


AlessandroPomponio:
This PR brings down the execution time of `ado show stats spaces` in `cplex_mip` from ~14m to ~1.5m

michael-johnston (approved):

PR #1274, opened 2026-07-24
feat(vllm-performance): add `prefix_repetition` benchmark dataset support

mgazz:
## feat(vllm-performance): add `prefix_repetition` benchmark dataset support

Adds support for the `prefix_repetition` benchmark dataset to the `vllm_performance` actuator. This dataset simulates KV-cache prefix-caching workloads by generating requests with a shared cached prefix followed by a unique suffix, making it useful for evaluating prefix-cache hit rates and their effect on throughput and latency.

### Changes

**`benchmark_models.py`** — Added four new optional fields to `BenchmarkParameters`: `prefix_repetition_prefix_len`, `prefix_repetition_suffix_len`, `prefix_repetition_num_prefixes`, and `prefix_repetition_output_len`.

**`execute_benchmark.py`** — Extended `execute_random_benchmark` to accept the four new parameters and, when `dataset="prefix_repetition"`, swap the `custom_args` dict passed to the vLLM benchmark runner from the `--random-*` flags to the corresponding `--prefix-repetition-*` flags.

**`experiment_executor.py`** — Threaded the new parameters through both `run_resource_and_workload_experiment` and `run_workload_experiment` call sites.

**`performance_testing.yaml`** — Added `prefix_repetition` as a valid `dataset` value across all five experiment definitions (`test-deployment-v1`, `test-endpoint-v1`, `test-deployment-guidellm-v1`, `test-endpoint-guidellm-v1`, `test-agentic-tool-calling`). Declared the four new properties with discrete domains and sensible defaults (prefix len: 2048, suffix len: 128, num prefixes: 10, output len: 128).

---

> **AI disclosure:** This code was generated using [IBM Bob](https://bob.ibm.com/) and was reviewed manually.


plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing.yaml
AlessandroPomponio, line 1: I believe the ranges are [inclusive, exclusive[, so double check that the excluded numbers are correct

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/vllm_performance_test/execute_benchmark.py
```diff
@@ -205,12 +205,16 @@ def execute_random_benchmark(
...
+    prefix_repetition_num_prefixes: int | None = None,
+    prefix_repetition_output_len: int | None = None,
 ) -> BenchmarkResult:
     """
-    Execute benchmark with random dataset
+    Execute benchmark with random or prefix repetition dataset.
     :param base_url: url for vllm endpoint
     :param model: model
-    :param dataset: data set name ["random"]
+    :param dataset: data set name ["random", "prefix_repetition"]
```
AlessandroPomponio: Given that it can be only two values it could be good to have it as a literal. Also, there's nothing checking that it's one of those two values

AlessandroPomponio (changes requested):
This requires a bump in the VERSION for vllm-performance

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/vllm_performance_test/execute_benchmark.py
christian-pinto: Also can those values be `None`? They all have an int default

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/vllm_performance_test/execute_benchmark.py
```diff
@@ -220,9 +224,26 @@ def execute_random_benchmark(
...
+    :param prefix_repetition_suffix_len: suffix tokens per request for prefix repetition dataset,
+    :param prefix_repetition_num_prefixes: distinct cached prefixes for prefix repetition dataset,
+    :param prefix_repetition_output_len: output tokens per request for prefix repetition dataset,
 
     :return: BenchmarkResult instance
     """
+    custom_args = {
+        "--random-input-len": number_input_tokens,
+        "--random-output-len": max_output_tokens,
+    }
```
christian-pinto: There is not point in first setting custom with the `random` dataset params and then rewriting if the dataset is `prefix_repetition`

christian-pinto (dismissed):

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiments/performance_testing.yaml
mgazz: Good catch, the upper bound values were off-by-one. Fixed in commit 53733410

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/vllm_performance_test/execute_benchmark.py
mgazz: good point 👍

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/vllm_performance_test/execute_benchmark.py
mgazz: No None does not make sense. I'll set to int and give defaults

mgazz:
Comments address as following:

- dataset typed as Literal[“random”, “prefix_repetition”] with assert_never for exhaustiveness
- prefix_repetition_* params changed from int | None to int with defaults
- custom_args built once in an if/elif, no redundant overwrite
- Domain ranges fixed (off-by-one on upper bounds)
- VERSION bumped

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/vllm_performance_test/benchmark_models.py
```diff
@@ -36,8 +36,14 @@ class BenchmarkParameters(pydantic.BaseModel):
...
     max_output_tokens: Annotated[int | None, pydantic.Field()] = None
+    prefix_repetition_prefix_len: Annotated[int | None, pydantic.Field()] = None
+    prefix_repetition_suffix_len: Annotated[int | None, pydantic.Field()] = None
+    prefix_repetition_num_prefixes: Annotated[int | None, pydantic.Field()] = None
+    prefix_repetition_output_len: Annotated[int | None, pydantic.Field()] = None
     burstiness: Annotated[float, pydantic.Field()] = 1.0
-    dataset: Annotated[str | None, pydantic.Field()] = "random"
+    dataset: Annotated[
+        Literal["random", "prefix_repetition"] | None, pydantic.Field()
+    ] = "random"
```
AlessandroPomponio: The fields here are still optional while in other places they aren't optional anymore

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/vllm_performance_test/execute_benchmark.py
```diff
@@ -220,9 +224,26 @@ def execute_random_benchmark(
...
+    :param prefix_repetition_output_len: output tokens per request for prefix repetition dataset (default 128),
 
     :return: BenchmarkResult instance
     """
+    custom_args = {
+        "--random-input-len": number_input_tokens,
+        "--random-output-len": max_output_tokens,
+    }
+
+    if dataset == "prefix_repetition":
```
AlessandroPomponio: There is still nothing actually checking that dataset is only "random" or "prefix_repetition"

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
AlessandroPomponio, line 592: It looks like the dataset can be more than just "random" and "prefix_repetition"

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
```diff
@@ -596,8 +596,11 @@ def run_resource_and_workload_experiment(
                     burstiness=benchmark_parameters.burstiness,
                 )
             else:
-                logger.info("Using vLLM random benchmark for deployment")
-                result = execute_random_benchmark(
+                logger.info("Using vLLM benchmark for deployment")
+                dataset: Literal["random", "prefix_repetition"] = (
+                    benchmark_parameters.dataset or "random"
+                )
```
AlessandroPomponio: This still allows values other than the two defined to go through, unless I'm missing validation somewhere else

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/vllm_performance_test/benchmark_models.py
christian-pinto: They are really never optional because they have a default parameterization, so you will always get a value when creating this model. Assuming you refer to the `prefix_*`. This is tru for other fields as well.

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/experiment_executor.py
christian-pinto: Adding validation inside `execute_vllm_benchmark`

AlessandroPomponio (approved):
LGTM thanks

PR #1279, opened 2026-07-24
fix(core): propagate root logger level to Ray workers via runtime env

### LINKED ISSUE #1277: bug(core): --loglevel does not propagate to ray_tune
## Issue Description

The `--loglevel` CLI arg does not propagate to the ado-ray-tune code. The only way to configure the loglevel of the python code inside `ado-ray-tune` is to export the environment variable `LOGLEVEL=INFO`.

This is problematic because code in ado-ray-tune, like the SimpleStopper which controls when the operation stops, print messages to the INFO channel but these messages are silently ignored leaving the user guessing why their operation stopped.

### How to reproduce

1. create a ado-ray-tune operation without exporting `LOGLEVEL=INFO`
2. observe that you do not get INFO level messages from ado-ray-tune

## Expected behaviour

`--loglevel` should propagate to the code that ado-ray-tune is running.

### Python/ado/system info

Please include the output of:

```terminaloutput
python --version=3.11.10
ado version=2.0.0
Your OS=MacOS
```

AlessandroPomponio:
## Summary

Fixes a gap where Ray workers spawned during local execution had no knowledge of the CLI's log level, causing operator logs to be silently dropped. The root logger level is now propagated to Ray workers via a runtime environment variable, and stopper output that previously went directly to stdout is routed through the logging framework instead.

## High-level Changes

- **Log level propagation**: `orchestrate.py` now passes the current root-logger level to Ray workers via `runtime_env` so that remote tasks and actors inherit the same verbosity as the CLI process.
- **Logging cleanup in `ray_tune` stoppers**: Stopping-criteria summaries in `InformationGainStopper` and `BayesianMetricDifferenceStopper` are emitted via `self.log.info` instead of `print` / `sys.stdout.flush`.
- **`ray_tune` version bump**: Patch version bumped from `2.0.4` to `2.0.5`.

## Impact

No breaking changes. Ray workers in local mode now respect the log level set by the CLI (e.g. `-l debug`), and Ray Tune operator output is consistently routed through the logging framework rather than stdout.

---

> **AI Disclosure**: The code and this PR description were generated using [IBM Bob](https://bob.ibm.com/) and were reviewed manually.


VassilisVassiliadis (approved):
LGTM

PR #1287, opened 2026-07-27
fix(core): derive experiment reference from result data instead of dedicated column

### LINKED ISSUE #1286: fix(core): migrate experiment reference reads from SQL column to results JSON
## Summary

The samplestore currently reads experiment references back from a denormalised SQL column that stores a plain string representation of the `ExperimentReference` type. This string was never intended to be parsed — it is a by-product of serialisation, not a designed interchange format. It would be worth migrating all read paths to use the structured JSON already stored in the measurement results table, which is the canonical and safely-deserializable source of truth for this data.

## Motivation

Parsing the legacy string representation is fragile and has caused subtle bugs: the string encodes a fully-qualified semantic version, so two releases of the same experiment that should be considered equivalent end up treated as distinct. The `ExperimentReference` type itself defines equality at the major-version level, but string parsing bypasses that model entirely. The structured JSON in the results table deserialises cleanly through the type's own Pydantic model, respects the intended equality semantics, and is already used correctly in at least one read path in the codebase. Consolidating all reads onto this source would eliminate the class of bugs caused by string parsing and remove dead filtering parameters that currently have no callers.

## Scope

The following areas would likely need attention:

- **Measurement count methods** — dead filtering parameters that reference the legacy column could be removed.
- **Samplestore statistics** — experiment counting could be moved to read from the results JSON, with deduplication consistent with the type's equality model.
- **Experiment listing for an operation** — could be rewritten to join through the results table and deserialise structured JSON rather than parsing column strings.
- **Measurement request reads** — methods that reconstruct measurement request objects could be updated to source the experiment reference from the results JSON.
- **ORM filter mappings** — any mapping that routes a filter key to the legacy string column could be removed to avoid silently querying stale data.

AlessandroPomponio:
## Summary

Removes the dependency on the `experiment_reference` denormalised column in the measurement-requests table for deriving experiment context. The experiment reference is now read directly from the JSON payload stored in the measurement-results table, using SQLAlchemy ORM expressions instead of raw SQL strings where applicable.

## High-level Changes

- **Experiment reference sourced from result data**: Queries that previously read `experiment_reference` from the requests table now extract it via `JSON_EXTRACT` on the results table, with a fallback path for the legacy nested format. Python-side deduplication (`ExperimentReference.__hash__`) replaces the previous `COUNT(DISTINCT experiment_reference)` SQL aggregation.
- **`experiment_filter` parameter removed**: The `experiment_reference` column is no longer used as a filter in `measurement_requests_count_for_operation` and `measurement_results_count_for_operation`, so the `experiment_filter` argument has been dropped from both methods.
- **`experimentReference` column mapping removed**: The `experimentReference` → `experiment_reference` entry has been removed from `MEASUREMENT_REQUEST_COLUMN_MAPPINGS` in the filtering layer.
- **Raw SQL replaced with ORM in `measurement_request_by_id`**: The hand-written SQL string for fetching a single request by ID has been replaced with an equivalent SQLAlchemy `select()` statement.
- **`samplestore_statistics` updated**: Now issues two queries instead of one; the second fetches distinct experiment-reference JSON blobs from the results table, and Python deduplicates them by major version.
- **Test updated**: The stats test now reflects that a simulation using an already-present experiment leaves `number_of_experiments` unchanged rather than incrementing it by one.

## Impact

Read paths that surface experiment information — statistics, request lookups, and the experiments-in-operation query — will derive that information from result payloads rather than the requests table. No schema migration is required for existing data as long as results contain the expected JSON fields. The `experiment_filter` parameter is a breaking change for any callers of the two count methods.

---

*The code and this PR description were generated using [IBM Bob](https://bob.ibm.com/) and reviewed manually.*

### DIFF
--- a/ado/core/samplestore/orm/measurements_filtering.py
+++ b/ado/core/samplestore/orm/measurements_filtering.py
@@ -24,7 +24,6 @@ MEASUREMENT_REQUEST_COLUMN_MAPPINGS = {
     "status": "status",
     "timestamp": "timestamp",
     "type": "type",
-    "experimentReference": "experiment_reference",
 }
 
 # Direct column mappings for measurement results
--- a/ado/core/samplestore/sql.py
+++ b/ado/core/samplestore/sql.py
@@ -279,7 +279,9 @@ class SQLSampleStore(ActiveSampleStore):
                 "insert_id", Integer, primary_key=True, autoincrement=True
             ),
             sqlalchemy.Column("uid", CHAR(36), nullable=False, unique=True, index=True),
-            sqlalchemy.Column("experiment_reference", Text(3145728), nullable=False),
+            sqlalchemy.Column(
+                "experiment_reference", Text(length=3145728), nullable=False
+            ),
             sqlalchemy.Column("operation_id", String(256), nullable=False),
             sqlalchemy.Column("request_index", Integer, nullable=False),
             sqlalchemy.Column("request_id", String(256), nullable=False),
@@ -1515,7 +1517,6 @@ class SQLSampleStore(ActiveSampleStore):
     def measurement_requests_count_for_operation(
         self,
         operation_id: str,
-        experiment_filter: str | None = None,
         status_filter: MeasurementRequestStateEnum | None = None,
     ) -> int:
 
@@ -1530,10 +1531,6 @@ class SQLSampleStore(ActiveSampleStore):
             query_text += "AND status = :status_filter "
             query_parameters["status_filter"] = status_filter.value
 
-        if experiment_filter:
-            query_text += "AND experiment_reference = :experiment_filter "
-            query_parameters["experiment_filter"] = experiment_filter
-
         try:
             with self.engine.begin() as connectable:
                 query = sqlalchemy.text(query_text).bindparams(**query_parameters)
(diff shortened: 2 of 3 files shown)


michael-johnston (commented):
This PR changes two things
- Not searching on experiment_reference column
- How the MeasurementRequest object is created 

Two questions: 
- do we actually need the first change? Seems to make things more complex
- in the second change we perform a complex looking operation via SQL (the function.coalesce). Seems like the same thing could be accomplished in measurement_requests_cursor_to_pydantic with no additional LOC (just changing the value passed to experimentReference parameter of MeaurementRequest)

AlessandroPomponio:
> * do we actually need the first change? Seems to make things more complex

Given that we have agreed that the reference is not meant to be parsed, this PR eliminates any kind of reliance on the `experiment_reference` column (which, among other things, would only match exact versioning for experiments with the new versioning system). So yes, this is needed.



> * in the second change we perform a complex looking operation via SQL (the function.coalesce)

The coalesce operation is needed because as it stands we support two formats for the measurement results, the "normal" one, where the experiment reference is part of every `measurement`, and the "optimized" one, where our new serialiser has been used and the `experimentReference` is at the top level of the model. Coalesce just selects the first field that matches, so it's not really complex.

michael-johnston (approved):

PR #1300, opened 2026-07-28
refactor(vllm-performance): update experiment naming and versioning; deprecate old experiments

### LINKED ISSUE #1203: refactor(vllm-performance): use updated experiment versioning
e.g., introduce new experiments that have the updated versioning while keeping the existing v1 ones and mark those as deprecated then slowly remove them

christian-pinto:
## Summary

This PR introduces a versioning scheme for experiments in the `vllm-performance` actuator. All existing experiments have been marked as deprecated, and a new set of cleanly named, versioned replacement experiments has been added. The actuator code and executor have been updated to recognise both old and new identifiers, ensuring backward compatibility while signalling that users should migrate to the new names.

Resolves #1203

## High-level Changes

- **Deprecated legacy experiments**: all existing experiment identifiers (e.g. `test-deployment-v1`, `test-endpoint-guidellm-v1`, and all geospatial variants) have been marked `deprecated: true` in the experiment YAML catalog.
- **New versioned experiments added**: a full set of replacement experiments with cleaner identifiers has been introduced (`vllm-bench-deployment`, `vllm-bench-endpoint`, `guidellm-bench-deployment`, `guidellm-bench-endpoint`, and all geospatial equivalents), each carrying an explicit `version: 1.0.0` field.
- **Actuator and executor updated**: the routing logic in the actuator and experiment executor now recognises both old and new experiment identifiers, so existing spaces and operations remain functional.
- **Documentation updated**: the actuator README and website docs reflect the new experiment names and deprecation notices.

## Impact

Users can continue using existing discovery spaces and operations without modification — all legacy experiment identifiers remain functional. New spaces should use the new versioned experiment identifiers. The `deprecated: true` flag allows tooling to surface warnings for spaces still referencing the old names.


AlessandroPomponio (commented):
In addition to fixing the test we must make sure we can still load data from the other ones and update them. we must also consider if it's possible to add a validator that changes entity data or something from the old experiment to the new one - given that nothing but a rename changed - 
We must also update the documentation and examples

AlessandroPomponio (approved):
LGTM thanks

michael-johnston:
> also consider if it's possible to add a validator that changes entity data or something from the old experiment to the new one 

We just have to pay the price of changing the names. We shouldn't allow retroactive edit of spaces and the data collected from them as this breaks many guarantees we try to keep (spaces don't change, the schema implied by a space (observed property names etc.) doesn't change, the data associated with the space is the data collected by the ops etc.)

PR #1319, opened 2026-07-31
fix(trim): Fix numberEntities_iterative_modeling in Trim operator

### LINKED ISSUE #1318: bug: Trim sampling the wrong number of samples when starting with no prior
## Issue Description

When a Trim operation is started on a space with no prior, TRIM will do the initial sampling using samplingBudget.minPoints. 
Then, when figuring out the number of samples for the iterative phase it will be set to ` len(source_df) - initial_source_space_size` which equals once more to `samplingBudget.minPoints` because `initial_source_space_size=0`. This causes the trim operation to end prematurely without ever converging.

I have a tentative fix in the PR linked to this branch. Have a look if it makes sense. 
### How to reproduce

Start a TRIM operation on a space with no prior available.

christian-pinto:
This PR attempts fixing a but that prevents the TRIM operator to converge when executed on a space with no prior.

The original calculation was

```python
numberEntities_iterative_modeling = (
        len(source_df) - initial_source_space_size
        if op_output_characterization_no_prior.operation
        else params.samplingBudget.maxPoints
```

This is fine when no priors, but when there was evaluate the number of samples to take in treatise modelling step to  be 

```
 len(source_df) - initial_source_space_size
 ```
 
 Which is incorrect - this is just the number of sample taken in the no_priors step minus whatever was there (probably 0) 
 
 It should be 
 
 ```python
 numberEntities_iterative_modeling = params.samplingBudget.maxPoints - (
    len(source_df) - initial_source_space_size
)
```

Here `(len(source_df) - initial_source_space_size)` is the samples taken in no_prior stage. If this stage is not executed the value of this is 0.

In this way 
- If no_prior stage skipped -> same taken in no_prior stage is 0 and the sampling budget for iterative modelling is maxPoints
- If prior stage executed -> subtract the samples made in the no prior stage from max points


danielelotito (dismissed):
This enables a clearer, more intuitive behavior

michael-johnston:
@danielelotito There is a problem with the changes and one of the trim tests that leads to 0 entities being based to the iterative modelling stage you need to take a look at

The test is `test_trim_example_operation_succeeds`

```
        # maxPoints is a shared new-sample budget across no-priors and iterative phases
        numberEntities_iterative_modeling = params.samplingBudget.maxPoints - (
            len(source_df) - initial_source_space_size
        )
>       trim_rwparams = RandomWalkParameters(
            samplerConfig=trim_sampler_config,
            batchSize=1,
            numberEntities=numberEntities_iterative_modeling,
            singleMeasurement=True,
        )
E       pydantic_core._pydantic_core.ValidationError: 1 validation error for RandomWalkParameters
E       batchSize
E         Value error, Number of entities to sample 0 cannot be less than batch size 1 [type=value_error, input_value=1, input_type=int]
E           For further information visit https://errors.pydantic.dev/2.13/v/value_error
```



You can run it locally with 

```
pytest -k "test_trim_example_operation_succeeds[sqlite]" tests/
```

danielelotito:
The failure was due to the fact that the existing test configured `minPoints=8 and maxPoints=8`. Therefore, the remaining budget was zero. TRIM nevertheless constructed `RandomWalkParameters(numberEntities=0, batchSize=1)` and Pydantic  rejected this because numberEntities cannot be less than batchSize.

I have opted towards implementing a more solid behavior:   the number of expected entities is inspected explicitly by the test, and iterative modeling is entirely and cleanly fixed if the budget is exhausted by no priors operation.

Now I cannot approve the pr due to my commits :)

michael-johnston:
@danielelotito can you bump the trim version as the style check is failing 

```
┌─ Plugin version check
│
├─ ✗ plugins/operators/trim: files changed but VERSION was not bumped
│
├─ The following plugins have code changes but no valid VERSION bump:
│  ├─ plugins/operators/trim/VERSION
└─ Please bump VERSION (higher than main) in each plugin listed above before merging
```

I can then approve it.

michael-johnston:
@danielelotito Are you able to merge this?

christian-pinto:
> @danielelotito Are you able to merge this?

@michael-johnston you need to approve the PR and it will merge automatically after that.

VassilisVassiliadis (approved):

PR #1324, opened 2026-07-31
refactor(core): use SQLalchemy ORM constructs instead of raw SQL in SampleStore

### LINKED ISSUE #1323: refactor(core): migrate SQLSampleStore raw SQL queries to SQLAlchemy Core ORM-style constructs
## Summary

The `SQLSampleStore` class currently contains a number of methods that issue raw SQL strings via `sqlalchemy.text()`, even though the backing SQLAlchemy `Table` objects are already available on the class and used correctly elsewhere. This issue proposes migrating those remaining methods to use SQLAlchemy Core `select()` / `insert()` / `func.*` constructs — the same style already adopted by the majority of the class. One of the methods to be migrated also contains a runtime bug where a table name is incorrectly passed as a bind parameter, which should be addressed as part of this work.

## Motivation

Mixing raw SQL f-strings with ORM-style Core queries in the same class creates an inconsistent codebase that is harder to reason about, maintain, and audit. The raw SQL fragments also carry suppressed linter warnings (`# noqa: S608`) that mask potential issues and add noise. Migrating to Core constructs would make the class fully consistent, remove dialect-specific SQL strings from method bodies, eliminate the suppressed warnings, and fix the existing runtime bug in the process.

## Scope

The changes should be confined to the SQL-level samplestore implementation. No schema changes, test additions, or fixture modifications are expected — the existing test suite already provides coverage for all affected methods. The work can be approached method-by-method, starting with the simplest count/select conversions, moving through inserts, and finishing with the more complex aggregation and join queries. A final lint and full test pass should close out the work.

AlessandroPomponio:
## Summary

Refactor `SQLSampleStore` to replace all hand-written raw SQL strings with SQLAlchemy Core expression constructs.

## High-level Changes

- All raw `sqlalchemy.text(f"...")` queries in `SQLSampleStore` replaced with SQLAlchemy Core `select`, `insert`, and expression API equivalents, using bound table/column references instead of f-string interpolation.
- Data passed to bulk `INSERT` operations for measurement results is now pre-deserialised to dicts (instead of being passed as a JSON string), to align with SQLAlchemy's ORM-style column binding.
- One exception retained: a MySQL 8.0+-specific `MEMBER OF(JSON_KEYS(...))` predicate, which has no SQLAlchemy Core equivalent, remains as a `sqlalchemy.text` fragment with bound parameters.
- All `# noqa: S608` suppressions for SQL injection warnings removed as they are no longer needed.

## Impact

No behaviour changes. The refactor eliminates f-string SQL interpolation across the sample store, improving safety and maintainability. Existing tests cover the affected methods.

---

## AI Disclosure

The code and this PR description were generated using [IBM Bob](https://bob.ibm.com/) and reviewed manually.

### DIFF
--- a/ado/core/samplestore/sql.py
+++ b/ado/core/samplestore/sql.py
@@ -201,18 +201,27 @@ class SQLSampleStore(ActiveSampleStore):
         # formats (compressed: top-level experimentReference; legacy: nested
         # inside measurements[0].property.experimentReference) so both are
         # handled without fetching and deserialising every row in Python.
-        query = sqlalchemy.text(f"""
-            SELECT DISTINCT entity_id
-            FROM {self._tablename}_measurement_results
-            WHERE COALESCE(
-                JSON_EXTRACT(data, '$.experimentReference.actuatorIdentifier'),
-                JSON_EXTRACT(data, '$.measurements[0].property.experimentReference.actuatorIdentifier')
-            ) IN ('replay', '"replay"')
-        """)  # noqa: S608 - self._tablename is not untrusted
+        from sqlalchemy import func, select
+
+        res_table = self._result_table
+        actuator_id_col = func.coalesce(
+            func.json_extract(
+                res_table.c.data, "$.experimentReference.actuatorIdentifier"
+            ),
+            func.json_extract(
+                res_table.c.data,
+                "$.measurements[0].property.experimentReference.actuatorIdentifier",
+            ),
+        )
+        stmt = (
+            select(res_table.c.entity_id)
+            .where(actuator_id_col.in_(["replay", '"replay"']))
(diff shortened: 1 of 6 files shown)


michael-johnston:
Is there any performance impact?

AlessandroPomponio:
@michael-johnston there shouldn't be, no. This makes sure we're using SQLAlchemy's ORM functionalities to generate the SQL to query the DB with. This makes sure we don't have SQL Injection issues and should allow us to add other SQLAlchemy supported DB backends without issues, as its dialect layer will take care of the differences in SQL which would otherwise trip raw SQL statements

michael-johnston (approved):
LGTM

PR #1325, opened 2026-08-01
fix(docs): broken link

michael-johnston:


AlessandroPomponio:
The issue here is that with the new versioned documentation, the documentation is published at `https://ibm.github.io/ado/VERSION/path/to/page/`. 

What we need to do here is to:

1. Find all absolute link references 
2. Update all the references in files living under `docs/` to use links to files, as they will be rendered by mkdocs
3. Update all references in files outside of these folders to link to the latest documentation

AlessandroPomponio:
Superseded by https://github.com/IBM/ado/pull/1329

PR #1346, opened 2026-08-05
fix(vllm-performance): Wait for deployment to be deleted before re-creating it

mgazz:
Fixes a race condition where retried benchmark runs received a `409 AlreadyExists` from Kubernetes because the previous deployment's cleanup was not waited on before re-creating.

### Changes

- **`k8s/__init__.py`** — added `K8sDeploymentDeletionTimeoutError`
- **`k8s/manage_components.py`** — added `wait_deployment_deleted()`: polls until the deployment object is gone or a deadline is reached
- **`k8s/create_environment.py`** — on creation error, calls `wait_deployment_deleted()` after deleting K8s resources and before re-raising, so retries never hit a stale object
- **`env_manager.py`** — `cleanup_failed_deployment()` converted to `async` and polls `check_deployment_exist()` with `await asyncio.sleep()` between checks; must be async (not `time.sleep`) because it runs on the event loop alongside concurrent benchmark tasks — blocking would stall them; slot release and conflict-manager signal moved to a `finally` block; env is only added to `deleting_environments` (GC) when deletion could not be confirmed, eliminating a previous `in_use`/`deleting` overlap
- **`tests/test_env_manager_gc.py`** — existing tests updated to `async`; new `TestCleanupFailedDeploymentWaits` suite added

---

> **AI disclosure:** This code was generated using [IBM Bob](https://bob.ibm.com/) and was reviewed manually.


plugins/actuators/vllm_performance/ado_actuators/vllm_performance/k8s/create_environment.py
```diff
@@ -90,7 +94,10 @@ def create_test_environment(
     :param kv_cache_dtype: KV cache data type (e.g. fp8, turboquant_k8v4); if None, vLLM default is used
     :param enable_prefix_caching: flag to enable prefix caching in vLLM
     :param check_interval: wait interval in seconds
-    :param timeout: timeout in seconds
+    :param timeout: timeout in seconds for deployment readiness
+    :param deletion_timeout: timeout in seconds to wait for a deployment to be
+        fully deleted after an error before re-raising. If deletion itself
+        times out the original error is still re-raised.
```
AlessandroPomponio: The explanation of the variable is confusing, I would just say something like "Timeout in seconds for Kubernetes to carry out the deployment deletion". I'm not sure the re-raising part is needed

mgazz: Fair point, I will update accordingly

AlessandroPomponio (approved):

PR #1355, opened 2026-08-12
feat(vllm-performance): extend gpu_memory_utilization supported values up to 0.98

mgazz:
Bumps the upper bound of gpu_memory_utilization from its previous limit to 0.98 in the vllm-performance actuator experiment config, and increments the plugin version accordingly.

PR #1358, opened 2026-08-18
fix(vllm-performance): Restore geospatial dataset names in `BenchmarkParameters`

mgazz:
**NOTE:** versioning considers that #1355 will be merged first

This patch fixes a routing bug in the `test-agentic-tool-calling` experiment and tightens the `dataset` field definition for the BFCL benchmark path.

### Changes

**Routing fix (`experiment_executor.py`):** The condition that selected the BFCL benchmark code path was `benchmark_parameters.dataset == "bfcl"`. This was fragile because any experiment could accidentally trigger it by setting `dataset=bfcl`. The condition is now `experiment.identifier == "test-agentic-tool-calling"`, making the routing explicit and experiment-scoped. Both `run_resource_and_workload_experiment` and `run_workload_experiment` are updated.

**Dataset forwarding (`execute_benchmark.py`):** `execute_bfcl_benchmark` now accepts a `dataset` parameter and validates that it equals `"bfcl"`, raising a `ValueError` otherwise. 

**Experiment definition (`performance_testing.yaml`):** The `test-agentic-tool-calling` experiment restricts the `dataset` property domain to `['bfcl']` only (was `['random', 'prefix_repetition', 'bfcl']`) and changes the default value from `random` to `bfcl`. Version bumped from `1.0.0` → `1.1.1`.

**Model loosening (`benchmark_models.py`):** The `dataset` field on `BenchmarkParameters` is widened from `Literal["random", "prefix_repetition", "bfcl"] | None` to `str | None`, removing the enum constraint at the model level (validation is now done at the call site for the BFCL path).

**Version bump:** `vllm_performance` plugin bumped to 1.14.1

**Docs:** A version reference in the actuators user guide is updated to reflect the new experiment version.

---

### AI Disclosure
This code was generated using [IBM Bob](https://bob.ibm.com/) and was reviewed manually.


christian-pinto (dismissed):

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/vllm_performance_test/execute_benchmark.py
```diff
@@ -320,6 +320,7 @@ def execute_bfcl_benchmark(
     base_url: str,
     model: str,
     bfcl_categories: str,
+    dataset: str | None = "bfcl",
```
christian-pinto: can the dataset ever be `None`?

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/vllm_performance_test/execute_benchmark.py
```diff
@@ -348,6 +350,9 @@ def execute_bfcl_benchmark(
 
     :return: BenchmarkResult instance
     """
+    if dataset != "bfcl":
+        raise ValueError(f"dataset must be 'bfcl', but got {dataset}")
+
```
christian-pinto: Do you need this check and the `dataset` field at all here? `dataset` can only be `bfcl`. Unless you are expecting at some point to be able to pass a different dataset to this function.

If not, I would remove the dataset parameter from the function and this check as well, then hardcode the `bfcl` value when invoking `execute_benchmark`.

christian-pinto (changes requested):

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/vllm_performance_test/execute_benchmark.py
mgazz: No, fair point. dataset_path can be None but not dataset. I'll update.

christian-pinto: See also my next comment. I believe this argument is not needed.

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/vllm_performance_test/execute_benchmark.py
mgazz: This is more a defensive check. But I can see how the experiment definition already ensures that the only supported datasets are selected (bfcl). 
I will remove.

plugins/actuators/vllm_performance/ado_actuators/vllm_performance/vllm_performance_test/execute_benchmark.py
mgazz: I think we can keep it because more tool calling datasets might get supported. It's the signature of the method that is misleading. `execute_tool_calling_benchmark` would be more general and reusable. 

In this case, to support a new tool calling benchmark we can just extend the set of supported dataset in the experiment.

christian-pinto (approved):

PR #1361, opened 2026-08-19
feat(active-learning): add PKH active-learning operator

### LINKED ISSUE #1362: feat: (active-learning): add PKH active-learning operator

Measuring an entity's target property in a finite Discovery Space is often slow or expensive - for example, running a benchmark, a physical experiment, or an expensive simulation. 
Building an accurate predictive surrogate model of the space requires measuring a large, often unnecessary fraction of it, because ADO's existing sampling strategies aren't designed to choose which entity to measure next based on how much it will actually improve the model (in the active learning sense). This wastes measurement budget on entities that add little information.

you want to happen.

A surrogate-model guided strategy to sequentially select entities that, once labeled through ADO, best reduces the distributional discrepancy between the currently labeled sample and the remaining unlabelled sample of the finite discovery space.

EmileAydar:
This PR adds `ado-active-learning`, a new operator plugin, and its first
operator: **PKH** (Predictive Kernel Herding), a regression-based
active-learning `characterize` operator for finite Discovery Spaces.

PKH adaptively selects one entity at a time to measure next, using a Random
Forest fit on the labels collected so far. Instead of measuring entities in a
fixed or random order, PKH selects the unmeasured entity with the largest mean, 
across trees, of the deficit between its full-pool and labelled-sample leaf masses, 
steering the labelled sample toward the full pool- i.e. it builds a sample that represents the
whole space with as few measurements as possible.

## Why add this to ADO

Discovery Spaces are often large and finite, and measuring an entity (running
an experiment/actuator) can be slow or expensive. ADO's existing operators for
this situation are either unguided (`random_walk`, which samples arbitrarily)
or optimization-focused (`ray_tune`/`rifferla`, which searches for optima
rather than characterizing the whole space). Neither targets "build a labelled
sample that faithfully represents the full pool, at minimum measurement
cost" - PKH fills that gap.

PKH is a direct implementation of the acquisition strategy from [1], which was
validated on predicting LLM-serving performance - a use case ADO already
supports via the `vllm_performance` actuator. This gives users of that (and
any other expensive) actuator a principled way to decide what to measure
next, instead of falling back to `random_walk`.

This PR also lays the shared groundwork
(`active_learning/regression/_shared.py`: parameter base, finite-pool
encoding, the sequential-selection state machine, RandomWalk/operator wiring)
for **FLORA**, a complementary active-learning operator with a different
objective (predictive accuracy rather than representativeness), submitted as
a follow-up PR stacked on this one. The package is organized as
`active_learning/regression/<strategy>/` so a future non-regression
active-learning strategy can be added as a sibling of `regression/` without
touching PKH or FLORA.

## How it works

PKH is a thin `characterize` operator that wraps ADO's built-in `random_walk`
explore operator, configured with a custom sampler (`PKHSampleSelector`).
After one deterministic pilot (5 randomly sampled entities) if no labels are already available, PKH repeats:

1. At an `epoch` boundary (where the hyperparameter `epoch` denotes the width of the refit schedule, set by default to 10 entities - i.e. the Random Forest is refit every 10 entities collected and labelled), fit a `RandomForestRegressor` on the labels collected so far. Otherwise, reuse the current forest.
2. Per tree, compare how much of the *full pool* falls in each leaf against
   how much of the *labelled sample* does.
3. Select the unmeasured entity with the largest mean leaf-mass deficit across trees.
4. Wait for ADO to record that entity's measurement, then repeat.

The forest is refit every `epochLength` selections (default 10), and leaf counts
are updated cheaply between refits.

## Example

```bash
uv sync --group test
ado get operators   # pkh now appears, type "characterize"

ado create space -f plugins/operators/active_learning/examples/discoveryspace.yaml --new-sample-store
ado create operation -f plugins/operators/active_learning/examples/operation_pkh.yaml --use-latest space
ado show measurements operation --use-latest --property-format target
```

## Non-Breaking Changes
Purely additive: a new plugin registered under a new `ado.operators` entry
point (```pkh```), added as a uv workspace member. No existing operator,
schema, or CLI behavior changes.

## Reference
```
[1] E. Aydar, C. Pinto, S. Venugopal, and D. Chatzopoulos, "Sampling
       Where It Matters: Predicting LLM Serving Performance " in 
       Proceedings of the Sixth European Workshop on Machine Learning and Systems 
       (EuroMLSys '26), ACM, 2026, pp. 13-22. doi: 10.1145/3805621.3807633.
```


michael-johnston:
@EmileAydar Thanks for this! 👍 

However I think this should be a custom sampler like we did for the 'NoPriorsSampler" rather than an operator  c.f. https://ibm.github.io/ado/latest/user-guide/operators/random-walk/#no-priors-sample-selector. 

That is, It looks like I could use this code as is via  RandomWalks custom sampler field. Then I wouldn't have nested operations etc. Is this correct? (check the no prior example and confirm if the same approach possible for PKH)

The issue is surfacing the available custom-samplers for RandomWalk, as except for the docs there is no way to do so at moment - also probably RandomWalk is not a good name anymore for this operator 😆 

Lets synch with Alessandro when he gets back so we see what best thing to do is.

EmileAydar:
Thanks @michael-johnston - that makes sense. I've checked the `NoPriorsSampleSelector` example again, and you are right that `PKHSampleSelector` already implements the custom-sampler interface.
The current `pkh` operator is essentially a thin facade that configures `random_walk` with that sampler, so PKH can indeed be invoked directly through  `random_walk` 's  `samplerConfig`,  which would get rid of the nested operations 👍 

The main difference is that `NoPriorsSampleSelector` computes its sampling order upfront, whereas PKH is feedback-driven: after selecting an entity, it must wait until that entity's target measurement is available in the Sample Store before choosing the next one, which for `random_walk` would require setting to `batchSize: 1` (though we could maybe consider the `epoch` hyperparameter of `pkh` as a batchSize param - this can be debated and would actually fit ADO better, but I'll wait for sync before more discussion), but the current sampler already implements that feedback loop, so it can still be hosted directly by `random_walk` without requiring a separate operator.

I agree that discoverability is then the important question: if custom samplers are only referenced by Python module and class, methods such as PKH will not appear in the operator listing or templates, and users may miss that ADO supports active-learning strategies at all. Perhaps this is an opportunity to surface named sampling strategies more explicitely - for example through a sampler registry/template mechanism and an active-learning section in the documentation - while keeping `random_walk` as the generic execution layer. That would also preserve clear strategy provenance in operation configs. 
And yes, I agree with you that the `RandomWalk` may have outgrown its name 😆 

Happy to sync with Alessandro and you, and to discuss more about this !

Issue #1369, opened 2026-08-25
feat: enable bulk ado edit


michael-johnston:
**Is your feature request related to a problem? Please describe.** 

If you want to edit N resources to add the same label you have to do N `ado edit` calls.
The cost is between 1 and 1.5 mins for 10 edits. This becomes problematic when cleaning up, or labelling in a large project e.g. labelling many resources for deletion can take 5mins+. 

**Describe the solution you'd like**. 

Enable a bulk edit e.g.

```
ado edit operation ID1 ID2 ID3 \
  -p "labels: {for_deletion: 'true', deletion_reason: error-no-entities}"
```

this follows similar pattern as establish by `ado delete`

Issue #1368, opened 2026-08-25
feat: enable csv output of ado get table


michael-johnston:
**Is your feature request related to a problem? Please describe.**

`ado get` does not provide a way to output the tables it produces to a file without rich elements like unicode box characters, column wrapping. This is problematic for writing agent skills that would like to store or access this table programatically. 

e.g. 

```
ado get ops -o stats --output-file op_stats.csv --details
```

**Describe the solution you'd like**. 

ado get supports `-o csv` and this also dumps the stats implicitly 

**Describe alternatives you've considered**. 

1. --output-file with `get` and an option that produces a table format  ( stats or table) writes the table to the file as a csv. Issue here is that other commands use -o csv to get CSV so its not consistent
2. Use `ado show stats -o csv` to get CSV output. Problem here is that (a) it computes heavy stats on spaces always is not required in some skills (b) documents are not supported. Adding lightweight stats for spaces or document as a type would seem to blur stats and get 
3. --stats is an explicit flag on get and adds the stats to -o format - table, csv and even YAML (e.g. in metadata) - in a consistent way

AlessandroPomponio:
This goes against our definition for the `-o | --output` flag, which expects the format in which to output the results, while `--output-file` is only meant to control whether to output to console or to a file.

So, while I'm in favour of adding `-o csv`, you wouldn't be able to get stats _and_ csv output, as it stands.

michael-johnston:
> So, while I'm in favour of adding -o csv, you wouldn't be able to get stats and csv output, as it stands.

I know - looking for a solution. I've given 4 above. If you can check them each and if none work suggests some more that's the best way forward. 

> This goes against our definition for the -o | --output flag, which expects the format in which to output the results, while --output-file is only meant to control whether to output to console or to a file.

This is a grey area, as the output location is already used by rich to e.g. strip colours.  

Also -o does not just set format it sets content c.f. --stats it a table format with additional content. 

So right now -o sets format and content and --output-file sends that to a file AND in certain cases performs modifications on the format when a file is detected over terminal. Some other flags e.g. --details add content.

PR #1374, opened 2026-08-25
docs: maintenance skill update

michael-johnston:
Fix a variety of issues with maintenance skill: 

- superseeded should only apply to non-explore operators.
- remove pre-release condition as most operations are run from local repos
- more efficient provisional resource identification (label provisional=true)
- improve identification of operations with status started and no-entities
- use `ado show` to identify stale and empty spaces instead of `ado get` due to issues with parsing tables output by ado get


michael-johnston:
Further improvements to the maintenance skill require changes in ado itself. These are captured in #1369 and #1368

VassilisVassiliadis (approved):
LGTM, I tested the read-only commands myself and they work as described in the docs.

PR #1382, opened 2026-08-28
feat(trim): remove NoPriorsParameters.targetOutput, use solely top level targetOutput

### LINKED ISSUE #1381: bug(trim): TRIM operation YAML contains 2 fields for setting the targetOutput, there should be just one
## Issue Description

The TRIM operation yaml contains two `targetOutput` fields, one for the no-priors sampler, and a second one for the iterative model building sampler.

There should be just one field that controls the target output of both samplers.

Example: https://github.com/IBM/ado/blob/main/examples/trim/example_yamls/op_pressure.yaml

```yaml
spaces:
  - space-c8717f-3a68bf
operation:
  module:
    operationType: characterize
    operatorName: trim
  parameters:
    # ... other parameters ...
    noPriorsParameters:
      targetOutput: pressure  # <---- this is superfluous
    targetOutput: pressure
```

## Additional information

The `noPriorsParameters.targetOutput` field exists because there used to be a second operation whose configuration included the pydantic basemodel `noPriorsParameters`. We just forgot to remove that field when we refactored the code to its latest state.

VassilisVassiliadis:
Fixes #1381


danielelotito (dismissed):
Good, I approve it, but you might have to bump the version as well

VassilisVassiliadis:
Ah yes, I forgot about that. Thanks for catching it

danielelotito (approved):
Yes, approved

PR #1384, opened 2026-08-28
fix(ordered-pip): handle moved parse_uri import across ray versions

VassilisVassiliadis:
Change happened in ray==2.57.0


srikumar003 (approved):
LGTM

PR #1391, opened 2026-09-01
feat(core): log and store experiment and actuator details in operation

### LINKED ISSUE #1347: feat: capture and log actuator versions and experiment versions used in operation
e.g., if there is a minor version change for the ~~actuator~~ experiment compared to what was defined in the space

michael-johnston:
1. Add experiment and actuator details to provenance section of operation resource

Example: 

```yaml
provenance:
  ado:
    distributionName: ado-core
    distributionVersion: 2.1.1.dev17+g6f6458d2.d20260901110925.dirty
  operators:
    ray_tune@2.0.6:
      distributionName: ado-ray-tune
      distributionVersion: 2.0.8.dev17+g78d8f3d6.d20260805124050.dirty
  actuators:
    custom_experiments:
      distributionName: ado-core
      distributionVersion: 2.1.1.dev17+g6f6458d2.d20260901110925.dirty
  experiments:
    experimentVersion: 1.0.0
 ```
    
 2.  Print if an experiment that matches on major version but has different minor/patch version to the experiment present when space was created was used
 
 Example: 
 
 ```
 Note: Will use solve_mip@1.2.1 to satisfy request for solve_mip@v1 (Major Version Match). The space was originally created with solve_mip@1.1.7
 ```

### DIFF
--- a/ado/core/operation/resource.py
+++ b/ado/core/operation/resource.py
@@ -19,6 +19,7 @@ from ado.core.resources import (
     ADOResourceStatus,
     CoreResourceKinds,
 )
+from ado.schema.reference import ExperimentReference
 
 
 class OperationResourceEventEnum(enum.Enum):
@@ -73,6 +74,28 @@ class OperationProvenanceInfo(ProvenanceInfo):
             default_factory=dict,
             description=(
                 "Mapping of operator identifier to the Python distribution that "
+                "provided it at the time this operation was created."
+            ),
+        ),
+    ]
+    experiments: Annotated[
+        list[ExperimentReference],
+        pydantic.Field(
+            default_factory=list,
+            description=(
+                "Catalog experiment references (including algorithm versions) that "
+                "were available to resolve the space experiments at the time this "
+                "operation was created. Space-frozen experiment versions remain on "
+                "the related discovery space."
+            ),
+        ),
+    ]
+    actuators: Annotated[
+        dict[str, PackageProvenance],
+        pydantic.Field(
+            default_factory=dict,
+            description=(
+                "Mapping of actuator identifier to the Python distribution that "
                 "provided it at the time this operation was created."
             ),
         ),
--- a/ado/core/operation/resource.py
+++ b/ado/core/operation/resource.py
@@ -83,10 +83,7 @@ class OperationProvenanceInfo(ProvenanceInfo):
         pydantic.Field(
             default_factory=list,
             description=(
-                "Catalog experiment references (including algorithm versions) that "
-                "were available to resolve the space experiments at the time this "
-                "operation was created. Space-frozen experiment versions remain on "
-                "the related discovery space."
+                "Experiments used in operation to satisfy the requirements of the measurement space "
             ),
         ),
     ]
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: ado/core/operation/resource.py [hunk 1]
```diff
@@ -77,6 +78,28 @@ class OperationProvenanceInfo(ProvenanceInfo):
...
     ]
+    experiments: Annotated[
+        list[ExperimentReference],
+        pydantic.Field(
+            default_factory=list,
+            description=(
+                "Catalog experiment references (including algorithm versions) that "
+                "were available to resolve the space experiments at the time this "
+                "operation was created. Space-frozen experiment versions remain on "
+                "the related discovery space."
```
AlessandroPomponio: I don't quite understand what this means: are these all the experiments that were registered when the operation ran?

Also what does "Space-frozen experiment versions" mean?

ado/modules/operators/setup.py
```diff
@@ -29,6 +32,30 @@
 moduleLog = logging.getLogger("setup")
 
 
+def find_fqi_differences(
```
AlessandroPomponio: what is fqi? Fully Qualified Identifier?

AlessandroPomponio: Yes - confirmed after scrolling.

AlessandroPomponio: After reading the function, it seems to me that it's meant to find, for each experiment in the measurement space, experiments that match on major version but do not match on the "fully qualified version". Maybe `find_measurement_space_registry_fqi_mismatch_pairs` would be a more descriptive name, even though it keeps the "fqi" word, which is not really readable in my opinion

ado/core/operation/resource.py
michael-johnston:
```suggestion
                "Experiments used in operation to satisfy the requirements of the measurement space "
```

michael-johnston: See below for better wording

ado/modules/operators/setup.py
michael-johnston: mismatch seems too severe because its suggests something is wrong which its not. 

That's why went for differences. Not sure about the long name. It seem quite a number of characters to still not fully explain what's written in the first lines of description.

AlessandroPomponio: "fqi_mismatch" to me seems very accurate

ado/modules/operators/setup.py
```diff
@@ -29,6 +32,30 @@
...
+    registry: ActuatorRegistry, measurement_space: MeasurementSpace
+) -> list[tuple[Experiment, Experiment]]:
+    """Returns experiments in catalog whose fqi is different to the one used to create space
+
+    The major version identifiers must match to be included.
```
michael-johnston:
```suggestion
    """Returns (registry experiment, measurement space experiment) pairs with same major version identifier but different fully qualified identifiers. 

```

ado/core/operation/resource.py
AlessandroPomponio: I like this much better

ado/modules/operators/setup.py
michael-johnston: What is the difference between a fqi mismatch and an fqi difference?

AlessandroPomponio: "Find Mismatch" to me tells me that you're looking for an exact match and getting the items that don't match
"Find difference" to me suggests that you're returning the difference between two things (e.g., the two versions)

AlessandroPomponio (approved):
LGTM thanks

PR #1393, opened 2026-09-02
refactor(autoconf): prepare the custom experiment to build models from data on hugging face 

### LINKED ISSUE #1390: refactor(autoconf): reorganize folder for the imminent upload on HF of ado-sftttrainer dataset
Pushing ado-sfttrainer Dataset to Hugging Face (HF) will require some refactoring in the autoconf plugin.

ADO Changes:

- Update the autoconf documentation to point to the Hugging Face dataset.
- Update the model training code to include a command that downloads the dataset from Hugging Face to a - specific location (provided the file is not there already).
- No models will be stored (neither in /ado nor on Hugging Face), and the model changelog will be deleted.

**Additional context**

Hugging Face (HF) Updates:
- The dataset will be hosted on HF, but no models will be uploaded.
- The code used to create the models will be referenced on HF, alongside a minimum reproducible series of commands, such as:
```bash
uv create venv --python 3.13
uv pip install ado
uv run plugins/.../autoconf/ml_classifier/training_script
```
- Text describing the dataset and its figures will be included.

**About models not being loaded**
Creating the models on the fly is actually easier: users run just one command from their venv, instead of downloading the model file, and displacing it in the right folder.
Moreover, by creating models in the venv that will consume them is safer and ensure that they work correctly, autogluon recommends this practice.

**Data already present of hf**
https://huggingface.co/datasets/ibm-research/LLM_Fine-Tuning_Performance
This repo contains the small subset of data used for the SKT paper.

danielelotito:
close #1390


danielelotito:
This PR requires updating the path to the sfttrainer dataset on hugging face (@srikumar003), then it will be ready for review.

On hf I suggest adding a section such as
```md
## Reproduce the AutoConf Model

Build the model in the same virtual environment that will run AutoConf:

``bash
python -m venv .venv
source .venv/bin/activate
python -m pip install "ado-autoconf==2.0.0"
autoconf_build_model
``

The command downloads `ado-sfttrainer-dataset.csv` when it is not already
present and writes the generated model into the installed AutoConf package. The
package pins AutoGluon 1.6.1.
```

srikumar003:
I have checked it out and it has a few issues:
1. The models are removed from the distribution. This will cause issues with downstream packages such as fms-autoconfig-k8s-controller, tuning-config-recommender, and even the examples that we supply in the README. All of these packages assume that autoconf will be able to load models and be ready for inference. 

By inserting a model building step that should be executed by the user, we cannot depend on `pip install ado-autoconf` to provide a method that will be inference-ready out of the box.

2. The `autoconf-build-model` command calls `plugins/custom_experiments/autoconf/autoconf/utils/autoconf_build/ml_classifier.py`  which assumes that it is running in the root directory of ado. The dataset is stored and models are created in a folder in `plugins/custom_experiments/autoconf/autoconf/` _relative to current location_. I tested this branch in a venv inside `ado/plugins/custom_experiments/autoconf` which led to creation of `plugins/custom_experiments/autoconf/autoconf/plugins/custom_experiments/autoconf/autoconf` where the models and dataset were stored

3. One input to `ml_classifier.py` is the URL of the HF dataset which should be sanitised to prevent any injection attacks

Of these, 1. is the only blocker (2 ad 3 are easily solvable). One solution is to upload the model that is built currently by autoconf to HuggingFace and have autoconf download it automatically. The flip side of this is that everytime we update the AutoGluon library version, we will have to upload a new version of the model to HuggingFace, as models built from one version of AutoGluon may not be compatible with another in the dependencies.

@michael-johnston what do you think ?

danielelotito:
The main issue with having the models either in the repository or on Hugging Face is that AutoGluon guidelines state that the virtual environment in which a model is created should be identical to the virtual environment that consumes it.

A secondary issue is that AutoGluon saves models in the unsafe .pkl (Pickle) format. Furthermore, all model consumers that depend on AutoGluon can also generate these models on the fly in just 2 to 3 minutes, depending on their hardware, and this solves the problem in my opinion.

**To diminish the dependencies and time required to have the models**
An alternative I am considering is creating the model in the ONNX format, which does not require AutoGluon. However, given our relatively small dataset, a comparable model could be trained on the fly in far less time than AutoGluon currently requires, albeit with a slight reduction in accuracy. Ultimately, the time required to install AutoGluon in a virtual environment is greater than the time it takes to download the dataset and train a standard GBDT model from scratch.
If we opt for this solution we can consider uploading the ONNX model on hugging face.

srikumar003:
I am in favour of uploading a GBDT model in ONNX format to HuggingFace which will fix the problem of `pip install ado-autoconf`. We can specify in the README how users can build and replace the default model with their own.

I have pushed a commit to sanitise the URLs

danielelotito:
As also explicit in  https://github.com/IBM/ado/issues/1390#issuecomment-5636678382 , 
this pull request is for shifting towards a more effective consumption of Autogluon models in AutoConf. This both upgrades Autoconf to version 1.6, enabling support of Python 3.14, and also fixes a possible issue with the fact that Autogluon models require consuming a model in the same environment that created them. At the first execution of the experiment, the data will be downloaded from the interface and a model will be built, keeping the same user experience as before without these problems. 

This PR brings the autoconf experiment to an improved state.

Further improvement can go in the direction of building an ONNX model and shipping it to Hugging Face (autoconf will then download it), which is a standard practice in the field. That would, in turn, remove the compulsory dependency on Autogluon of this repository.

plugins/custom_experiments/autoconf/autoconf/models/.gitignore
danielelotito, line 1: necessary. The code in model_paths.py computes DEFAULT_MODEL_ROOT as <package_root>/models/ and that path is used as the default location for trained models. Without the .gitignore, the models/ directory wouldn't exist in a fresh clone, which would break any code that tries to write or read models there without first creating the directory.

srikumar003 (dismissed):
I have checked out the PR and tested that model building on demand works along with the examples. 

LGTM

AlessandroPomponio (commented):
trim's pyproject.toml has been changed as well, so its VERSION must be bumped as well

AlessandroPomponio (approved):

PR #1394, opened 2026-09-03
feat(core): Decorate operations with ray job id

christian-pinto:
When a new operation is created while running inside a Ray cluster, the Ray job ID is now automatically captured and stored in the operation's metadata under the key `ray_job_id`. This makes it straightforward to correlate ado operations with the Ray jobs that produced them, aiding debugging and traceability in distributed/remote execution workflows. If Ray is not initialized or the job ID cannot be retrieved, the failure is silently logged at DEBUG level and operation creation proceeds normally.


tests/operators/test_operators.py
```diff
@@ -897,3 +897,63 @@ class _Cfg(pydantic.BaseModel):
...
 
     assert any("already registered" in r.getMessage() for r in caplog.records)
+
+
+# ---------------------------------------------------------------------------
+# create_operation_and_add_to_metastore — ray_job_id injection
+# ---------------------------------------------------------------------------
+
+
+def test_create_operation_sets_ray_job_id(
```
michael-johnston: You should piggy back on one of the existing end-to-end operation tests. The below, doesn't create a ray job so there is no id, also it doesn't initialise ray (the ado path for initialising ray requires going through modules.operators.orchestrate.orchestrate()

ado/modules/operators/base.py
```diff
@@ -493,6 +493,13 @@ def create_operation_and_add_to_metastore(
         provenance=OperationProvenanceInfo(operators=operators),
     )
 
+    if ray.is_initialized():
+        try:
+            ray_job_id = ray.get_runtime_context().get_job_id()
+            operation.metadata["ray_job_id"] = ray_job_id
+        except Exception:
+            moduleLog.debug("Could not retrieve Ray job ID", exc_info=True)
```
michael-johnston: I would move this up to _run_operation_harness because there you at least know an operation is launched.

michael-johnston: Also its technically possible for an operation not to be running in ray, so debug log should probably say something like "Operation is not running in ray" or "Operation is not running as a ray job - no ray job id".

michael-johnston (changes requested):

ado/modules/operators/_orchestrate_core.py
```diff
@@ -118,6 +118,18 @@ def _run_operation_harness(
...
 
+    if ray.is_initialized():
+        try:
+            ray_job_id = ray.get_runtime_context().get_job_id()
+            operation_resource.metadata["ray_job_id"] = ray_job_id
+        except Exception:
+            moduleLog.debug("Could not retrieve Ray job ID", exc_info=True)
+            moduleLog.warning(
+                "Could not retrieve Ray job ID - operation will proceed without it"
+            )
```
michael-johnston: Why a debug and warning log? I would just make the warning log the debug log (or maybe an info log)

ado/modules/operators/_orchestrate_core.py
```diff
@@ -118,6 +118,18 @@ def _run_operation_harness(
...
+    else:
+        moduleLog.debug("Operation is not running as a Ray job - no ray_job_id set")
```
michael-johnston:
```suggestion
        moduleLog.info("Operation is not running as a Ray job - no ray_job_id set")
```

michael-johnston (changes requested):
Some minor things

ado/modules/operators/_orchestrate_core.py
christian-pinto: The debug print will also include a stack trace, that's why I had them separate. I can merge them and remove the debug print. Or, keep the debug print with just the stack trace.

michael-johnston (approved):

AlessandroPomponio:
We need to make sure we choose PR names very carefully given that they're used in our changelog - this PR doesn't "decorate" anything in the Python sense of `@decorator`, it adds the ray job id in the operation resource metadata

PR #1404, opened 2026-09-07
docs(agents): add standalone run_experiment skill

michael-johnston:
Agents with less capable models were often missing run_experiment and how to execute it, or hallucinating its CLI. This is because it was in `using-ado-cli` which some models took to mean the "ado` tool and its subcommands." regardless of what the front-matter or skill said. 

Keeping this separate to the issue if we should move run_experiment under `ado` or not. 

Moving it to a separate tool outside of using-ado-cli which now descibes just `ado` command. 

This PR also changes the `--TIMEOUT` option to `--timeout` (which is what is documented in the website).


michael-johnston:
@AlessandroPomponio the PR name seems to failing the CI due to conventional commits. How can I change it so it passes?

AlessandroPomponio (approved):
LGTM thanks

Issue #1405, opened 2026-09-07
bug(skills): remote-execution does not handle experiments that require building wheels for dependencies


VassilisVassiliadis:
## Issue Description

The remote-execution skill does not guide Agents to:
1. check whether the experiment requires building wheels for dependencies that a simple pip install without any kind of authentication will fail e.g. python packages installed from private repositories on Git
2. prefix wheels uploaded to the Ray Remote cluster via `additionalFiles` with the `${RAY_RUNTIME_ENV_CREATE_WORKING_DIR}/` prefix.

# How to reproduce

1. Ask your Agent to create a remote ray operation involving an experiment whose python package has an optional dependency to a python package that lives in a private Github repository.
2. The first ray execution execution environment YAML file for `ado --remote ray-executon-environment.yaml create operation` command that Agent creates configures ado to install the python package from source but does not ask for the installation of the optional packages in the `sweep` group.
3. The Agent then identifies that the python package must be manually built and uploaded to the remote ray cluster via the `additionalFiles` section of the ray execution environment YAML.
4. The Agent correctly creates an entry for the wheel file  under the `packages.fromPyPI` map. However, the Agent does not prefix the wheel with `${RAY_RUNTIME_ENV_CREATE_WORKING_DIR}/` which the right way to inform Ray that the wheel is to be found in the working directory of the ray job.
5. The ray job fails with `error: Distribution not found at: file:///tmp/ray/session_2026-09-04_06-49-52_155261_1/runtime_resources/uv/142e00093bf867096e4366a5ad6f92bf8462f4b2/exec_cwd/my_wheel-0.9.7-py3-none-any.whl` error during the initial virtual environment setup. This is before it gets to the point that it calls the `ado create operation` command on the RayCluster.

The first ray_execution.yaml file that the Agent builds is:

```yaml
# ... other fields
packages:
  fromPyPI:
    - ado-core
    - ado-ray-tune
  fromSource:
    - .
```

The above is missing the `additionalFields` and `packages.fromPyPI` entry.

## Expected behaviour

The Agent should have discovered that my experiment requires building a wheel and it should have produced the following yaml: 

```yaml
# ... other fields
packages:
  fromPyPI:
    - ado-core
    - ado-ray-tune
    # this is the filename of the wheel the Agnet should have built, prefixed with ${RAY_RUNTIME_ENV_CREATE_WORKING_DIR}
[...]

PR #1410, opened 2026-09-09
fix(core): validate and symlink remote execution wheels in packages.fromPyPI

VassilisVassiliadis:
Contributes to #1405

### DIFF
--- a/.agents/skills/remote-execution/SKILL.md
+++ b/.agents/skills/remote-execution/SKILL.md
@@ -254,7 +254,7 @@ packages:
 additionalFiles:
   - /absolute/local/path/to/data_file.csv
   - path/to/my_data_dir/   # directories also supported
-  - dist/my_local_pkg-1.0.0-py3-none-any.whl # local wheel to upload
+  - my_wheels/my_local_pkg-1.0.0-py3-none-any.whl # local wheel to upload
 ```
 
 > [!NOTE] Local wheels
@@ -262,6 +262,12 @@ additionalFiles:
 > Local wheel files must be listed in `additionalFiles` to be uploaded, and
 > their bare filename specified in `packages.fromPyPI`. Multiple files in
 > `additionalFiles` cannot share the same basename.
+>
+> **Do not place custom wheels inside `dist/`** when any `fromSource` entry
+> targets the parent of the `dist/` directory. The build step runs
+> `uv build --wheel -o dist --clear`, which deletes the contents of `dist/`
+> before writing the new wheel. Store extra wheels in a separate directory
+> (e.g. `my_wheels/`).
 
 Use bare filenames (no path) in space/operation YAML; ray copies
 `additionalFiles` entries into the Ray working directory.
--- a/.agents/skills/remote-execution/SKILL.md
+++ b/.agents/skills/remote-execution/SKILL.md
@@ -243,6 +243,7 @@ packages:
     - ado-ray-tune
     - ray==2.52.1                         # pin to match cluster version if needed
     - /remote/path/to/package.whl         # path to a wheel already on the cluster
+    - my_local_pkg-1.0.0-py3-none-any.whl # local wheel (bare name, listed in additionalFiles)
   fromSource:
     - plugins/actuators/my_plugin  # relative to where ado --remote is run
 ```
@@ -253,7 +254,14 @@ packages:
 additionalFiles:
   - /absolute/local/path/to/data_file.csv
   - path/to/my_data_dir/   # directories also supported
+  - dist/my_local_pkg-1.0.0-py3-none-any.whl # local wheel to upload
 ```
+
+> [!NOTE] Local wheels
+>
+> Local wheel files must be listed in `additionalFiles` to be uploaded, and
+> their bare filename specified in `packages.fromPyPI`. Multiple files in
+> `additionalFiles` cannot share the same basename.
 
 Use bare filenames (no path) in space/operation YAML; ray copies
 `additionalFiles` entries into the Ray working directory.
--- a/tests/ado/test_remote_submission.py
+++ b/tests/ado/test_remote_submission.py
@@ -763,6 +763,25 @@ def test_remote_dispatch_wheel_only_in_additional_files(
     assert loaded["uv"] == ["ado-core"]
 
 
+def test_additional_files_in_from_source_dist_raises() -> None:
+    """additionalFiles entry inside a fromSource package's dist/ raises ValidationError.
+
+    When a fromSource package is built, uv build --clear wipes its dist/
+    directory. Any additionalFiles entry pointing into that same dist/ will
+    be deleted before Ray can upload it, so validation must reject this.
+    """
+    raw = {
+        "executionType": {"type": "cluster", "clusterUrl": "http://localhost:8265"},
+        "packages": {
+            "fromSource": ["plugins/my_plugin"],
+            "fromPyPI": ["dep_wheel-1.0-py3-none-any.whl"],
+        },
+        "additionalFiles": ["plugins/my_plugin/dist/dep_wheel-1.0-py3-none-any.whl"],
+    }
+    with pytest.raises(pydantic.ValidationError, match="fromSource"):
+        RemoteExecutionContext.model_validate(raw)
+
+
 def test_write_runtime_env_with_ray_config(tmp_path: pathlib.Path) -> None:
     """runtimeEnv maps to Ray config keys in runtime_env.yaml."""
     ctx = RemoteExecutionContext(
--- a/tests/ado/test_remote_submission.py
+++ b/tests/ado/test_remote_submission.py
@@ -8,6 +8,7 @@ import subprocess
 from io import BytesIO
 from unittest.mock import MagicMock, patch
 
+import pydantic
 import pytest
 import yaml
 from typer.testing import CliRunner
@@ -621,13 +622,13 @@ def test_write_runtime_env_env_vars_only(tmp_path: pathlib.Path) -> None:
 
 
 def test_write_runtime_env_local_wheel_in_pypi(tmp_path: pathlib.Path) -> None:
-    """Local .whl paths in fromPyPI are copied to working_dir and rewritten."""
+    """Local .whl paths in fromPyPI are symlinked to working_dir and rewritten."""
     local_whl = tmp_path / "my_local-1.0-py3-none-any.whl"
     local_whl.write_bytes(b"fake wheel")
 
     ctx = RemoteExecutionContext(
         executionType=ClusterExecutionType(clusterUrl="http://localhost:8265"),
-        packages=PackageConfiguration(fromPyPI=[str(local_whl), "ado-core"]),
+        packages=PackageConfiguration(fromPyPI=[local_whl.name, "ado-core"]),
     )
     dest = tmp_path / "runtime_env.yaml"
     working_dir = tmp_path / "working"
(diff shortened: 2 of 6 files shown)


tests/ado/test_remote_submission.py
```diff
@@ -621,13 +622,13 @@ def test_write_runtime_env_env_vars_only(tmp_path: pathlib.Path) -> None:
...
 def test_write_runtime_env_local_wheel_in_pypi(tmp_path: pathlib.Path) -> None:
-    """Local .whl paths in fromPyPI are copied to working_dir and rewritten."""
+    """Local .whl paths in fromPyPI are symlinked to working_dir and rewritten."""
     local_whl = tmp_path / "my_local-1.0-py3-none-any.whl"
     local_whl.write_bytes(b"fake wheel")
 
     ctx = RemoteExecutionContext(
         executionType=ClusterExecutionType(clusterUrl="http://localhost:8265"),
-        packages=PackageConfiguration(fromPyPI=[str(local_whl), "ado-core"]),
+        packages=PackageConfiguration(fromPyPI=[local_whl.name, "ado-core"]),
```
VassilisVassiliadis, line 631: Remote paths are propagated as is  to the ray_runtime.yaml file that ado uses when executing `ray job submit` and thus these paths are not expected to exist in the local machine. They are expected to exist on the RayCluster.

Currently, local paths also act as a syntax sugar for implicit `additionalFiles` entries. This is to maintain the old behaviour of ado which would always copy local wheels into the temporary directory it creates prior to submitting the ray job.

Finally, the wheels are no longer copied into this temporary directory. Instead, they are symlinked just like we do for any other `additionalFiles` entry.

.agents/skills/remote-execution/SKILL.md
```diff
@@ -253,8 +254,21 @@ packages:
...
+>
+> Local wheel files must be listed in `additionalFiles` to be uploaded, and
+> their bare filename specified in `packages.fromPyPI`. Multiple files in
+> `additionalFiles` cannot share the same basename.
+>
+> **Do not place custom wheels inside `dist/`** when any `fromSource` entry
+> targets the parent of the `dist/` directory. The build step runs
+> `uv build --wheel -o dist --clear`, which deletes the contents of `dist/`
+> before writing the new wheel. Store extra wheels in a separate directory
+> (e.g. `my_wheels/`).
```
michael-johnston, line 270: Why do we need this? Seems like AI anxiety.

michael-johnston (commented):
I'd forgotten that you could put local wheel path in pypi. 

So I see one change is the symlinking instead of copying. I think this could go in another PR as it's not directly related to the issue. 

What is the other change trying to solve? I'm no longer sure we had the problem we thought.

.agents/skills/remote-execution/SKILL.md
VassilisVassiliadis: If you use `additionalFiles` to place a different wheel under the same `dist/` directory of a `package.fromSource` entry then that custom wheel will be deleted before `ray job submit` starts. Then the environment on the RayCluster will not be created because that wheel will be missing.

VassilisVassiliadis:
> So I see one change is the symlinking instead of copying. 

This makes no difference whatsoever. I can revert back to copying. It's just something I did to keep the code a bit more consistent.

> What is the other change trying to solve?

ADO does ~not~ copy the wheels of `additionalFiles` into the temporary working directory it creates but does not prefix them  in the runtime_env.yaml with `${RAY_RUNTIME_ENV_CREATE_WORKING_DIR}/` even if they are present in `packages.fromPyPI`.

The changes are:
1. if an entry in `packages.fromPyPI` is a base wheel name then when generating the `runtime_env["uv"]` array of packages, prefix it with `${RAY_RUNTIME_ENV_CREATE_WORKING_DIR}/` --> this is the fix to the issue I ran into
2. if an entry in `packages.fromPyPI` is an absolute path, then this is wheel is expected to be found on the remote RayCluster. Users that wish to install a wheel from an absolute path on their local machine should instead use `additionalFiles` with a minimal entry in `packages.fromPyPI` that is just the filename of the wheel --> this makes the protocol for defining wheels more consistent.


I maintained the current behaviour of `ado --remote create operation` for listing relative paths, and bare wheel names, within `packages.FromPyPI` and having ado automatically resolve the relative path locally and include the referenced wheel in the temporary working dir it builds prior to running `ray job submit`. The only change I made there was to also use a symlink instead of copying in the file, just like we do for `additionalFiles`. When users opt for relative paths in `packages.fromPyPI` but without an associated entry in `additionalFiles` they get a warning.

Also, I maintained the behaviour of not installing a wheel just because it is present in `additionalFiles`. Placing a wheel in `additionalFiles` only ensures that it gets delivered to ray. You must then ask for it to be installed via `packages.fromPyPI`

Finally, I noticed that agents preferred to place custom wheels they built to satisfy the dependencies of the python package in the `packages.fromSource` entry inside the same `dist/` directory as the `fromSource` entry. When `ado` uses `uv build --clear` to build the wheel for the `fromSource` entry, the files in the `additionalFiles` array that were also under that same `dist/` directory would get automatically deleted. So I added both a check in the validation of the remote execution environment YAML as well as a sentence in the remote-execution Skill to avoid placing custom wheels in `dist/` directories whose parent directory is found in any `packages.fromSource` entry.

.agents/skills/remote-execution/SKILL.md
michael-johnston: In what situations would this happen? Does it happen?

michael-johnston:
So they way I see it now, the problem arose from the agent using an unsupported pattern when a supported pattern was available. So then there are two separate issues/solutions

1. Update so the agent uses the supported pattern (bug: execution skill does not cover how to specify local wheel use)
2. Change ado to supported the current unsupported (feat: ado should support bare wheel name in fromPYPI+entry additional files  instead of full path in fromPyPI)

This PR is doing 2 - should we actually change this is my question.

michael-johnston:
> Finally, I noticed that agents preferred to place custom wheels they built to satisfy the dependencies of the python package in the packages.fromSource entry inside the same dist/ directory as the fromSource entry. 

This seems strange, so to be clear what happens is
1. The see the local package you want to build wheel for has a dependency on a source repo
2. They clone the source repo (?) as they realise they need to build that wheel also
3. They build the dependent wheel but transfer it to the dist/ dir of the first package (?) 
   or run uv build --wheel -o ../local-package/dist --clear`

VassilisVassiliadis:
I'll run some more tests and get back to you.

VassilisVassiliadis:
My agent did not clone the dependency repo, they used `pip build ....` to build the wheel without downloading it. So one solution would be to have a sentence in the skills that would guuide the agent to first clone and then add the repos to the `packages.fromSource` array.

michael-johnston:
> y agent did not clone the dependency repo, they used pip build .... to build the wheel without downloading it.

But then I guess it ran pip build inside the dist dir or supplies that dir as an arg to pip build?

VassilisVassiliadis:
> But then I guess it ran pip build inside the dist dir or supplies that dir as an arg to pip build?

It used a pip arg to control where to place the output wheel

VassilisVassiliadis:
How about I close this PR and open a new one with the following changes:

- Update the pydantic validation to make `test_remote_dispatch_wheel_in_additional_files_under_root` and `test_remote_dispatch_wheel_in_additional_files_under_with_rel_path` consistent
    -  Rule: There cannot be 2 entries in either `packages.fromPyPI` or `additionalFiles` that end with `".whl"` and have the same filename.
- Update the pydantic validation to fix `test_remote_dispatch_from_source_and_additional_files_with_wheel_in_dist_dir`
    - Rule: `additionalFiles` entries cannot have a parent directory that is the "dist/" child of any of the `packages.fromSource` entries`

This way, we don't need to change the skills or business logic of ado at all.

If the agents make a mistake like `test_remote_dispatch_from_source_and_additional_files_with_wheel_in_dist_dir` or `test_remote_dispatch_wheel_in_additional_files_under_with_rel_path` they'll just get a validation error and quickly fix the problem they ran into.

I'll then open a different issue for handling cases where the python package that contains the ADO experiment has dependencies to packages that are not on public PyPI.

----

I started second guessing this PR so I wrote a couple of tests to investigate the behaviour of the `main` branch:

1. `test_remote_dispatch_wheel_in_additional_files_under_with_rel_path` (Fail)
    - `{packages': {'fromPyPI': ['my_wheel-0.9.7-py3-none-any.whl', 'ado-core']}, 'additionalFiles': ['wheels/my_wheel-0.9.7-py3-none-any.whl']}`
    - it uploads a wheel via `additionalFiles` into the working directory of the remote ray job and also asks ado to install it in the venv
    - **Failure reason**: the resulting Ray environment YAML contains `my_wheel-0.9.7-py3-none-any.whl` instead of `${RAY_RUNTIME_ENV_CREATE_WORKING_DIR}/my_wheel-0.9.7-py3-none-any.whl`
    - I think this is abusing the `additionalFiles` feature. The fix is to update the pydantic validation such that it raises a ValueError for RemoteExecutionContext configurations that have wheel files in `packages.fromPyPI` and `additionalFiles` which have identical filenames. This is actually what's happening in the test below
2. `test_remote_dispatch_wheel_in_additional_files_under_root` (Fail)
    - `{packages': {'fromPyPI': ['my_wheel-0.9.7-py3-none-any.whl', 'ado-core']}, 'additionalFiles': ['my_wheel-0.9.7-py3-none-any.whl']}`
    - Same as above, only difference is that the `additionalFiles` entry points to a wheel in the root of the working directory of `ado --remote create operation`
    - **Failure reason**: `ValueError: File basename collision: two file arguments resolve to the same filename 'my_wheel-0.9.7-py3-none-any.whl'. Rename one of the files before submitting remotely.`
        - I think this Failure is consistent with the way `ado --remote create operation` is meant to work
        - This is what I would expect for `test_remote_dispatch_wheel_in_additional_files_under_with_rel_path` too
        
3. `test_remote_dispatch_wheel_in_additional_files_under_root_without_additional_files` (Success)
    - `{packages': {'fromPyPI': ['my_wheel-0.9.7-py3-none-any.whl', 'ado-core']}}`
    - Installs just a wheel which is in the root of the working dir for the `ado --remote create operation` command
    - **Success**
4. `test_remote_dispatch_wheel_in_additional_files_under_root_without_additional_files_with_rel_path` (Success)
    - `{packages': {'fromPyPI': ['wheels/my_wheel-0.9.7-py3-none-any.whl', 'ado-core']}}`
    - Same as above, only now the wheel is in a relative path under the working dir
    - **Success**
5.  `test_remote_dispatch_from_source_and_additional_files_with_wheel_in_dist_dir` (Failure)
    - `{packages': {'fromPyPI': ['dist/second_wheel-0.9.7-py3-none-any.whl', 'ado-core'], 'fromSource': ['.']}, 'additionalFiles': ['dist/second_wheel-0.9.7-py3-none-any.whl']}` 
    - Installs a wheel from its source, also manually places a different wheel inside the same "dist" directory that ado will use to build the wheel for `fromSource=["."]`.
    - **Failure**: `FileNotFoundError: additionalFiles entry 'dist/second_wheel-0.9.7-py3-none-any.whl' does not exist`
        - The manually placed wheel gets deleted by ado during the building of the `fromSource=["."]` entry


<details><summary>The code for the unit tests is here</summary>

I placed these tests under `tests/ado/test_other_remote_submission.py` and tested it via `uv run pytest tests/ado/test_other_remote_submission.py`

```python
import importlib
import pathlib
import subprocess
from io import BytesIO
from unittest.mock import MagicMock, patch

import pytest
import yaml

from ado.cli.utils.remote.dispatch import (
    _build_source_wheels,
    _symlink_additional_files,
    _write_runtime_env,
)
from ado.core.remotecontext.config import (
    ClusterExecutionType,
    PackageConfiguration,
    RemoteExecutionContext,
)


def test_remote_dispatch_wheel_in_additional_files_under_with_rel_path(tmp_path: pathlib.Path) -> None:
    wheel_dir = tmp_path / "wheels"
    wheel_dir.mkdir()
    wheel_file = wheel_dir / "my_wheel-0.9.7-py3-none-any.whl"
    wheel_file.write_bytes(b"wheel content")

    ctx = RemoteExecutionContext(
        executionType=ClusterExecutionType(clusterUrl="http://localhost:8265"),
        packages=PackageConfiguration(
            fromPyPI=["my_wheel-0.9.7-py3-none-any.whl", "ado-core"]
        ),
        additionalFiles=["wheels/my_wheel-0.9.7-py3-none-any.whl"],
    )
    dest = tmp_path / "runtime_env.yaml"
    working_dir = tmp_path / "working"
    working_dir.mkdir()

    seen_basenames: set[str] = set()
    _symlink_additional_files(
        ctx.additionalFiles, tmp_path, working_dir, seen_basenames
    )
    _write_runtime_env(ctx, [], dest, tmp_path, working_dir, seen_basenames)

    assert (working_dir / wheel_file.name).is_file()
    loaded = yaml.safe_load(dest.read_text())
    assert f"${{RAY_RUNTIME_ENV_CREATE_WORKING_DIR}}/{wheel_file.name}" in loaded["uv"]
    assert "ado-core" in loaded["uv"]

    assert len(loaded["uv"]) == 2


def test_remote_dispatch_wheel_in_additional_files_under_root(tmp_path: pathlib.Path) -> None:
    wheel_file = tmp_path / "my_wheel-0.9.7-py3-none-any.whl"
    wheel_file.write_bytes(b"wheel content")

    ctx = RemoteExecutionContext(
        executionType=ClusterExecutionType(clusterUrl="http://localhost:8265"),
        packages=PackageConfiguration(
            fromPyPI=["my_wheel-0.9.7-py3-none-any.whl", "ado-core"]
        ),
        additionalFiles=["my_wheel-0.9.7-py3-none-any.whl"],
    )
    dest = tmp_path / "runtime_env.yaml"
    working_dir = tmp_path / "working"
    working_dir.mkdir()

    seen_basenames: set[str] = set()
    _symlink_additional_files(
        ctx.additionalFiles, tmp_path, working_dir, seen_basenames
    )
    _write_runtime_env(ctx, [], dest, tmp_path, working_dir, seen_basenames)

    assert (working_dir / wheel_file.name).is_file()
    loaded = yaml.safe_load(dest.read_text())
    assert f"${{RAY_RUNTIME_ENV_CREATE_WORKING_DIR}}/{wheel_file.name}" in loaded["uv"]
    assert "ado-core" in loaded["uv"]

    assert len(loaded["uv"]) == 2

def test_remote_dispatch_wheel_in_additional_files_under_root_without_additional_files(tmp_path: pathlib.Path) -> None:
    wheel_file = tmp_path / "my_wheel-0.9.7-py3-none-any.whl"
    wheel_file.write_bytes(b"wheel content")

    ctx = RemoteExecutionContext(
        executionType=ClusterExecutionType(clusterUrl="http://localhost:8265"),
        packages=PackageConfiguration(
            fromPyPI=["my_wheel-0.9.7-py3-none-any.whl", "ado-core"]
        )
    )
    dest = tmp_path / "runtime_env.yaml"
    working_dir = tmp_path / "working"
    working_dir.mkdir()

    seen_basenames: set[str] = set()
    _symlink_additional_files(
        ctx.additionalFiles, tmp_path, working_dir, seen_basenames
    )
    _write_runtime_env(ctx, [], dest, tmp_path, working_dir, seen_basenames)

    assert (working_dir / wheel_file.name).is_file()
    loaded = yaml.safe_load(dest.read_text())
    assert f"${{RAY_RUNTIME_ENV_CREATE_WORKING_DIR}}/{wheel_file.name}" in loaded["uv"]
    assert "ado-core" in loaded["uv"]

    assert len(loaded["uv"]) == 2

def test_remote_dispatch_wheel_in_additional_files_under_root_without_additional_files_with_rel_path(tmp_path: pathlib.Path) -> None:
    wheel_dir = tmp_path / "wheels"
    wheel_dir.mkdir()
    wheel_file = wheel_dir / "my_wheel-0.9.7-py3-none-any.whl"
    wheel_file.write_bytes(b"wheel content")

    ctx = RemoteExecutionContext(
        executionType=ClusterExecutionType(clusterUrl="http://localhost:8265"),
        packages=PackageConfiguration(
            fromPyPI=["wheels/my_wheel-0.9.7-py3-none-any.whl", "ado-core"]
        )
    )
    dest = tmp_path / "runtime_env.yaml"
    working_dir = tmp_path / "working"
    working_dir.mkdir()

    seen_basenames: set[str] = set()
    _symlink_additional_files(
        ctx.additionalFiles, tmp_path, working_dir, seen_basenames
    )
    _write_runtime_env(ctx, [], dest, tmp_path, working_dir, seen_basenames)

    assert (working_dir / wheel_file.name).is_file()
    loaded = yaml.safe_load(dest.read_text())
    assert f"${{RAY_RUNTIME_ENV_CREATE_WORKING_DIR}}/{wheel_file.name}" in loaded["uv"]
    assert "ado-core" in loaded["uv"]

    assert len(loaded["uv"]) == 2


def test_remote_dispatch_from_source_and_additional_files_with_wheel_in_dist_dir(tmp_path: pathlib.Path) -> None:
    # 1. Create a minimal Python package at tmp_path so uv build can produce a wheel.
    (tmp_path / "src").mkdir()
    (tmp_path / "src" / "my_source_pkg").mkdir()
    (tmp_path / "src" / "my_source_pkg" / "__init__.py").write_text("")
    (tmp_path / "pyproject.toml").write_text(
        "[project]\n"
        'name = "my-source-pkg"\n'
        'version = "1.0.0"\n'
        'requires-python = ">=3.9"\n'
    )

    # 2. Place the second wheel that additionalFiles / fromPyPI reference.
    dist_dir = tmp_path / "dist"
    dist_dir.mkdir()
    second_wheel = dist_dir / "second_wheel-0.9.7-py3-none-any.whl"
    second_wheel.write_bytes(b"fake wheel content")

    ctx = RemoteExecutionContext(
        executionType=ClusterExecutionType(clusterUrl="http://localhost:8265"),
        packages=PackageConfiguration(
            fromPyPI=["dist/second_wheel-0.9.7-py3-none-any.whl", "ado-core"],
            fromSource=["."],
        ),
        additionalFiles=["dist/second_wheel-0.9.7-py3-none-any.whl"],
    )

    working_dir = tmp_path / "working"
    working_dir.mkdir()
    dest = tmp_path / "runtime_env.yaml"

    # 3. Build wheels from source (runs real uv build inside tmp_path).
    wheel_names = _build_source_wheels(
        ctx.packages.fromSource, working_dir, repo_root=tmp_path
    )

    # 4. Symlink additionalFiles into the working directory.
    seen_basenames: set[str] = set(wheel_names)
    _symlink_additional_files(
        ctx.additionalFiles, tmp_path, working_dir, seen_basenames
    )

    # 5. Write runtime_env.yaml.
    _write_runtime_env(ctx, wheel_names, dest, tmp_path, working_dir, seen_basenames)

    # 6. The source-built wheel must be present in working_dir and in the uv list.
    assert len(wheel_names) == 1
    source_wheel_name = wheel_names[0]
    assert (working_dir / source_wheel_name).is_file()

    # 7. The second wheel must be symlinked into working_dir (via additionalFiles).
    assert (working_dir / second_wheel.name).is_file()

    loaded = yaml.safe_load(dest.read_text())
    uv_entries = loaded["uv"]

    assert f"${{RAY_RUNTIME_ENV_CREATE_WORKING_DIR}}/{source_wheel_name}" in uv_entries
    assert f"${{RAY_RUNTIME_ENV_CREATE_WORKING_DIR}}/{second_wheel.name}" in uv_entries
    assert "ado-core" in uv_entries
    assert len(uv_entries) == 3
```

</details>

VassilisVassiliadis:
The above tests reflect the intended behaviour of `ado --remote create operation`.

The fix is not to add validation code but to add documentation and update the `remote-execution` Agentic Skill such that we avoid running into this problem altogether.

PR #1411, opened 2026-09-10
fix(core): ensure correct identifier is used in prints for parameterized experiments

michael-johnston:
If an experiment was parameterized, the parameterization column of the Experiments table showed None instead of the parameterization as expected.


AlessandroPomponio (approved):
LGTM thanks

PR #1413, opened 2026-09-11
fix(core): prevent error in live table display from failing the operation

michael-johnston:
Transient read errors in live console output code would crash entire operation. This PR

- Handles read exceptions in live console, re-rendering last obtained table instead of raising
- At top level if live console exits due to exception, it is captured and code waits on operation to exit as it is the primary process (either it also exits or it doesn't and we don't stop e.g. long sampling operation)


AlessandroPomponio:
Should we not retry a few times (with a backoff) first instead of giving up at the first failure?
Also, is it better to show stale data or to always say we can't show the latest results?

michael-johnston:
> Should we not retry a few times (with a backoff) first instead of giving up at the first failure?
Also, is it better to show stale data or to always say we can't show the latest results?

Its currently fires every 2 seconds, so retrying or switching from table to say "we can't show latest results" seems pointless. We also have no idea at any given point if there was new data, and it's much more likely at any particular time that the table has stayed the same.

AlessandroPomponio (approved):

AlessandroPomponio (approved):

PR #1414, opened 2026-09-11
build(vllm-performance): require vllm>=0.27.0

AlessandroPomponio:


pyproject.toml
```diff
@@ -121,6 +121,17 @@ RAY_ENABLE_UV_RUN_RUNTIME_ENV = "0"
...
+# ado-vllm-performance requires vllm>=0.27.0 (security), which depends on numba==0.65.0.
+# ado-profile-space requires fg-data-profiling, which depends on numba>=0.60,<0.63.
+# These numba version ranges are incompatible, so the two packages are declared as conflicting.
+conflicts = [
+  [
+    { package = "ado-vllm-performance", extra = "vllm" },
+    { package = "ado-profile-space" },
+  ],
+]
+
```
AlessandroPomponio: If we declare the conflict like this, our test group ends up skipping `ado-profile-space` anyway, so we can't do it.

@michael-johnston `ado-profile-space` is used in `test_operator_callable_for_harness_unwraps_decorated_operator`. Can we use something else instead of it? More in general, it's using `fg-data-profiling`, which is an unmaintained package. We should think about what to do with it

michael-johnston: Sure can use anything else. 

In general updating this is in the pipeline, that will remove the fg-data-profiling.

michael-johnston: While waiting can remove ado-profile-space from the test group?

AlessandroPomponio: Sure, I will remove it from the test group and from the test cases

christian-pinto (approved):

PR #1415, opened 2026-09-11
docs(agents): document manual cloning of git+ssh dependencies for fromSource

### LINKED ISSUE #1405: bug(skills): remote-execution does not handle experiments that require building wheels for dependencies

VassilisVassiliadis:
Fixes #1405 

This fixes a gap in the `remote-execution` skill about experiments whose python package has a dependency to other packages that are installed via `pip install git+ssh://`.

These packages must be:

* configured as optional dependencies
   * I haven't explicitly checked, but I think that if they are listed as required dependencies with an `extras` group for installing them from git there's the risk of `pip install` attempting to fetch a package with the same name from public PyPI.
* git cloned from the Git server locally
* added to `packages.fromSource`


.agents/skills/remote-execution/SKILL.md
```diff
@@ -190,6 +190,24 @@ If you are still seeing stale wheels, confirm the plugin's `pyproject.toml`
...
 
+### Dependencies to packages obtained via git and ssh must be manually cloned and added to packages.fromSource
+
+If you use a `fromSource` entry you must also:
+
+1. Read its `pyproject.toml` or `setup.py` to collect every `git+ssh://` dependency.
+   - To find which extras apply: locate the `[project.entry-points."ado.custom_experiments"]`
+     section; the module listed there is the ado experiment. Search that module (and any modules
+     it imports from the same package) for imports of packages that appear in the
+     `[project.optional-dependencies]` extras. Include those extras in the scan.
```
michael-johnston: I don't understand what this bit is for?

VassilisVassiliadis: These are the steps to identify which of the optional groups in the pyproject.toml file are necessary for the experiment to function.

The problem I'm trying to solve here is how to select which of the python dependencies of the top-level package that are installed via`git+ssh://` are required for the experiment to work. The challenge is that there can be several groups with this kind of dependencies.

An alternative approach would be to ask the agent to:
1. identify all `git+ssh://` dependencies
2. ask the user which of those should be included

.agents/skills/remote-execution/SKILL.md
```diff
@@ -190,6 +190,24 @@ If you are still seeing stale wheels, confirm the plugin's `pyproject.toml`
 has the `format-jinja` block from the
 [plugin-development](../plugin-development/SKILL.md) skill.
 
+### Dependencies to packages obtained via git and ssh must be manually cloned and added to packages.fromSource
```
michael-johnston:
```suggestion
### Handling Version Control Source (VCS) Dependencies
```

VCS Dependency is the term used this type of deps in general.  PEP 440 and PEP 508 also define "direct reference" but this includes zips etc.

.agents/skills/remote-execution/SKILL.md
```diff
@@ -190,6 +190,24 @@ If you are still seeing stale wheels, confirm the plugin's `pyproject.toml`
...
+
+If you use a `fromSource` entry you must also:
```
michael-johnston: For e
```suggestion
It's necessary to detect and handle any VCS dependencies used by `fromSource` packages. The process is:
```

.agents/skills/remote-execution/SKILL.md
```diff
@@ -190,6 +190,24 @@ If you are still seeing stale wheels, confirm the plugin's `pyproject.toml`
...
+
+1. Read its `pyproject.toml` or `setup.py` to collect every `git+ssh://` dependency.
```
michael-johnston: Suggest generalising to 
```suggestion
1. Obtain list of all VCS dependencies in the packages project configuration files e.g. pyproject, setup.py
```

.agents/skills/remote-execution/SKILL.md
```diff
@@ -190,6 +190,24 @@ If you are still seeing stale wheels, confirm the plugin's `pyproject.toml`
...
+   - To find which extras apply: locate the `[project.entry-points."ado.custom_experiments"]`
+     section; the module listed there is the ado experiment. Search that module (and any modules
+     it imports from the same package) for imports of packages that appear in the
+     `[project.optional-dependencies]` extras. Include those extras in the scan.
+2. If no local clone exists strictly under the `dependencies/` directory,
+   clone it immediately:
+   ```bash
+   # given: my-utils @ git+ssh://git@github.com/ibm/my-utils.git@branch_name
+   mkdir -p dependencies && git clone git@github.com:ibm/my-utils.git -b branch_name dependencies/my-utils
+   ```
```
michael-johnston:
```suggestion
2. Clone each VCS dependency under the `dependencies/` directory. For example
   ```bash
   # given: my-utils @ git+ssh://git@github.com/ibm/my-utils.git@branch_name
   mkdir -p dependencies && git clone git@github.com:ibm/my-utils.git -b branch_name dependencies/my-utils
   ```
```

.agents/skills/remote-execution/SKILL.md
```diff
@@ -190,6 +190,24 @@ If you are still seeing stale wheels, confirm the plugin's `pyproject.toml`
...
+3. Add the cloned path to `fromSource`. Never source wheels or source trees from `site-packages`.
```
michael-johnston: Not sure why the "Never source wheels ..." is here. Seems like AI anxiety. 

```suggestion
3. Add each cloned path to `fromSource`. 
```

.agents/skills/remote-execution/SKILL.md
```diff
@@ -190,6 +190,24 @@ If you are still seeing stale wheels, confirm the plugin's `pyproject.toml`
...
+### Dependencies to packages obtained via git and ssh must be manually cloned and added to packages.fromSource
+
+If you use a `fromSource` entry you must also:
+
+1. Read its `pyproject.toml` or `setup.py` to collect every `git+ssh://` dependency.
+   - To find which extras apply: locate the `[project.entry-points."ado.custom_experiments"]`
+     section; the module listed there is the ado experiment. Search that module (and any modules
+     it imports from the same package) for imports of packages that appear in the
+     `[project.optional-dependencies]` extras. Include those extras in the scan.
+2. If no local clone exists strictly under the `dependencies/` directory,
```
michael-johnston: I suggest an additional point in the workflow before current 2

```
2. Determine if public access: Remove any VCS dependency that refers to an open repository on e.g. public GitHub, from the list. 
If no VCS dependencies remain, stop.

File: docs/user-guide/advanced/remote-execution.md [hunk 8]
```diff
@@ -276,6 +276,13 @@ envVars:
 1. Build python wheels for those packages
 2. Instruct Ray to install the wheels as part of the Ray job submission
 
+> [!NOTE] git+ssh dependencies
+>
+> Ray cluster nodes are not guaranteed to have SSH access to git servers.
+> If your package has `git+ssh://` dependencies, make them optional in your
+> top-level `pyproject.toml` and clone their source locally so you can add
+> the local paths to `packages.fromSource` instead.
```
michael-johnston: What about using uv sources like we do?

docs/user-guide/advanced/remote-execution.md
```diff
@@ -276,6 +276,13 @@ envVars:
 1. Build python wheels for those packages
 2. Instruct Ray to install the wheels as part of the Ray job submission
 
+> [!NOTE] git+ssh dependencies
```
michael-johnston:
```suggestion
> [!NOTE] Version Control System dependencies
```

docs/user-guide/advanced/remote-execution.md
```diff
@@ -276,6 +276,13 @@ envVars:
...
+>
+> Ray cluster nodes are not guaranteed to have SSH access to git servers.
```
michael-johnston:
```suggestion
> Ray cluster nodes are not guaranteed to have access to any VCS's specified in python package dependencies. 
```

.agents/skills/remote-execution/SKILL.md
VassilisVassiliadis: I added this bit because my agent found the wheel of a dependency in my .venv and tried to upload it with `packages.fromPyPI` but I think that this is an antipattern.

docs/user-guide/advanced/remote-execution.md
VassilisVassiliadis: Yes that would work. 

Here's an example:

```
dependencies = [
    "private-dependency"
]

[tool.uv.sources]
private-dependency = { git = "ssh://git@github.com/ibm/private-dependency.git", tag = "v0.4.0" }
```

- `uv pip install .` automatically installs `private-dependency` from git
- `pip install .` complains that it cannot find `private-dependency` on public PyPI but installing both the dependency and `.` in one go works: `pip install "private-dependency @ ssh://git@github.com/ibm/private-dependency.git@v0.4.0" .`

.agents/skills/remote-execution/SKILL.md
VassilisVassiliadis: This is the one change I didn't apply

.agents/skills/remote-execution/SKILL.md
```diff
@@ -190,6 +190,22 @@ If you are still seeing stale wheels, confirm the plugin's `pyproject.toml`
...
 [plugin-development](../plugin-development/SKILL.md) skill.
 
+### Handling Version Control Source (VCS) Dependencies
+
+It's necessary to detect and handle any VCS dependencies used by `fromSource` packages. The process is:
+
+1. Obtain list of all VCS dependencies in the packages project configuration files e.g. pyproject, setup.py
+2. Determine if public access: Remove any VCS dependency that refers to an open repository on e.g. public GitHub, from the list.
+If no VCS dependencies remain, stop.
+3. Clone each VCS dependency under the `dependencies/` directory. For example
```
VassilisVassiliadis, line 200: I didn't comment on this detail before, but if I skip the instructions for using a dedicated `dependencies` directory my agent goes hay wire and starts scanning my entire home to see if I have cloned the dependency before.

.agents/skills/remote-execution/SKILL.md
michael-johnston: It doesn't seem like the correct place to have "source trees" since nothing in this section is about source trees. 

It also doesn't seem the correct place for wheels as nothing here is about wheels (the workflow is a set of steps and the prior step is not about wheels).

If this guideline should be added it should be in the fromSource and fromPyPI  (wheels subsection)

docs/user-guide/advanced/remote-execution.md
michael-johnston: How about`pip install . $WHEEL_FOR DEP` or `uv pip install WHEELA WHEELB` when uv can't access github

?

.agents/skills/remote-execution/SKILL.md
michael-johnston: Is it not in step 3?

VassilisVassiliadis: It is indeed. This is the motivation for including it in the first place.

.agents/skills/remote-execution/SKILL.md
VassilisVassiliadis: The skill works without the bit about `site-packages`. I removed it!

michael-johnston (dismissed):

michael-johnston (approved):

PR #1421, opened 2026-09-14
docs(website): add section on using library-skills

### LINKED ISSUE #1398: feat:  docs: add doc on installing ado skills via library-skills

ado supports distribution of its skills via library-skills to help agent based development with ado. 
However we don't have any documentation on this approach on the website. 

A section in the install guide on the library-skills approach

AlessandroPomponio:
## Summary

Improves the "Getting set up for agent-assisted workflows" section of the user guide by documenting a new, simpler way to install ado's agent skills without cloning the full repository.

## High-level Changes

- **Docs (`ado-and-agents.md`)**: Added a new recommended setup option ("Option 1") that uses the `library-skills` tool to install ado's agent skills directly from the installed `ado-core` package via `uvx library-skills`. The existing clone-based setup is retained as "Option 2".

## Impact

Documentation-only change. No code, tests, or configuration are affected. Users with `ado-core` as a dependency can now set up agent skills without cloning the ado repository.

---

> **AI Disclosure**: The code and this PR description were generated using [IBM Bob](https://bob.ibm.com/) and reviewed manually.


docs/user-guide/ado-and-agents.md
```diff
@@ -35,6 +35,28 @@ plain language:
...
+If you have `ado-core` as a dependency in your project, you can install ado's
+skills directly into your workspace using
+[library-skills](https://library-skills.io/use/), without cloning the
+repository.
+
+From your project directory, run:
+
+```shell
+uvx library-skills
+```
```
michael-johnston, line 51: Do you not need to install library skills? Also perhaps we should tell them to install ado-ray-tune, so they have it? 

```suggestion
From your project directory, run:

```shell
uv add ado-core library-skills
uvx library-skills --yes --all
```
```

`--yes --all` as without they can be presented with a large number of skills they have to read and agree to install. This seems more one-shot YOLO way most people seem to prefer. 

Can add some extra info on the flags for the cautious types.

File: docs/user-guide/ado-and-agents.md [hunk 2]
```diff
@@ -35,6 +35,28 @@ plain language:
...
 ## Getting set up for agent-assisted workflows
 
+There are two ways to get ado's skills into your agent-enabled IDE.
+
+### Option 1: install via library-skills (recommended)
+
+If you have `ado-core` as a dependency in your project, you can install ado's
+skills directly into your workspace using
+[library-skills](https://library-skills.io/use/), without cloning the
+repository.
```
michael-johnston, line 45: Since these are instructions for setting up an agent assisted worked - essentially installing ado, it seems strange to say "if you have "ado-core" as a dependency" . Also the "install ado's skills directly ..." info is basically repeated below.   Probably none of these lines is necessary

```suggestion
```

docs/user-guide/ado-and-agents.md
```diff
@@ -35,6 +35,28 @@ plain language:
...
+From your project directory, run:
+
+```shell
+uvx library-skills
+```
+
+The tool scans your installed environment, finds the skills bundled with
+`ado-core`, and creates symbolic links under `.agents/skills` (or
+`.claude/skills` for Claude Code). Your agent will then pick them up
+automatically.
```
michael-johnston, line 56: Maybe: 

```suggestion
`library-skills adds `ado`s skills under your projects `.agents/skills` (or
`.claude/skills` for Claude Code). Your agent will then pick them up
automatically.
```

michael-johnston (changes requested):

docs/user-guide/ado-and-agents.md
AlessandroPomponio: > Do you not need to install library skills?

No, uvx is like npx, it pulls the dependency on demand

> Also perhaps we should tell them to install ado-ray-tune, so they have it?

It's not relevant for this section

I also don't like `--yes --all` because it'll install all skills provided by *all* dependencies

docs/user-guide/ado-and-agents.md
AlessandroPomponio: I disagree - people need to have `ado-core` in the dependencies of their project (i.e., in their setup.py/pyproject.toml), otherwise it won't work, that's why I have a specific callout

michael-johnston (approved):

michael-johnston:
I noticed this PR did not add the instructions to the [install section](https://ibm.github.io/ado/latest/user-guide/getting-started/#installing) of Getting Started

This means this methods, which is noted as the [recommended method in ado and agents ](https://ibm.github.io/ado/latest/user-guide/ado-and-agents/#option-1-install-via-library-skills-recommended) is not the one users first encounter

PR #1425, opened 2026-09-14
feat: handle invalid warm-start and return best-bound

michael-johnston:
This PR adds the following to the solve_,mip experiment

* Fail gracefully if warm start file cannot be used instead of ignoring and continuing to solve as if it was never given
* return best_bound (dual) as a scalar metric. Previously only in time-series. 
* Set version of experiment to 0.2


plugins/custom_experiments/cplex_mip/tests/test_solve_mip.py
```diff
@@ -51,6 +51,10 @@ def test_actuator_identifier(self, experiment: Experiment) -> None:
         """Actuator identifier must be 'custom_experiments'."""
         assert experiment.actuatorIdentifier == "custom_experiments"
 
+    def test_experiment_version(self, experiment: Experiment) -> None:
+        """Experiment version must be 0.2.0 (scalar best_bound target property)."""
+        assert experiment.version == "0.2.0"
```
AlessandroPomponio, line 56: Is this a valid test worth having?

plugins/custom_experiments/cplex_mip/VERSION
```diff
@@ -1 +1 @@
-1.0.3
+1.2.0
```
AlessandroPomponio, line 1:
```suggestion
1.1.0
```

plugins/custom_experiments/cplex_mip/cplex_mip_experiments/solve_mip.py
```diff
@@ -291,6 +296,74 @@ def _normalize_cplex_value(value: float | None) -> float | None:
...
+            self._buf = self._buf.rsplit("\n", 1)[-1]
+        return len(data)
+
+    def flush(self) -> None:
+        """Flush the underlying stream and inspect any remaining buffered text."""
+        if self._buf:
+            self._watcher.observe(self._buf)
+        flush = getattr(self._dest, "flush", None)
+        if callable(flush):
+            flush()
```
AlessandroPomponio, line 364: Can't we reuse (possibly subclassing) one of the default logger handlers instead of doing all of this?
https://docs.python.org/3/library/logging.handlers.html

AlessandroPomponio (dismissed):

plugins/custom_experiments/cplex_mip/tests/test_solve_mip.py
michael-johnston: Thought I had removed it

plugins/custom_experiments/cplex_mip/cplex_mip_experiments/solve_mip.py
michael-johnston: Lets see

michael-johnston: The issue is the CPLEX does not use python logging system. You have to give it file-like objects.

e.g.

solve_mip.py
Ln 766–769
```python
model.set_log_stream(log_stream)
model.set_error_stream(warn_stream)
model.set_warning_stream(warn_stream)
model.set_results_stream(log_stream)
```

set_log_stream / set_warning_stream expect something with write() and flush(), like sys.stdout.  You cannot pass a handler to CPLEX.

That is the root of the problem.  We have to observe the solver’s native output stream, looking for:

Warning: No solution found from 1 MIP starts.

and aborting the solve when it appears.

This code also handles Chunked Writes which loggers don't need to.

That is, CPLEX can flush that warning in pieces ("Warning: No solution found from" then " 1 MIP starts.\n"). The tee’s buffer exists specifically for that.

michael-johnston:
@AlessandroPomponio bump

AlessandroPomponio (approved):

PR #1426, opened 2026-09-15
feat(cli): add ado show measurements samplestore

### LINKED ISSUE #1422: feat(cli): support ado show measurements samplestore

AlessandroPomponio:
## Summary

Extends `ado show measurements` to support samplesestores as a first-class resource type, alongside the existing `discoveryspace` and `operation` targets. Users can now query and export all measured entities directly from a samplestore using the familiar `ado show measurements samplestore <id>` command.

## High-level Changes

- **New CLI target**: `samplestore` (alias `store`) added to `ado show measurements`, backed by a new handler that fetches all measured entities from the store and formats them using the existing output pipeline.
- **Constrained `--include` mode**: Because a bare samplestore has no entity-space knowledge, only `--include measured` is accepted; space-relative modes (`sampled`, `unsampled`, `matching`, `missing`) are rejected with a clear error.
- **Tests**: Four integration tests cover the happy path, an empty store, a non-existent ID, and the unsupported `--include` rejection.
- **Docs & skills updated**: CLI reference, `query-ado-data` skill, and `using-ado-cli` skill all reflect the new target and its constraints.

## Impact

Purely additive — no existing behaviour for `discoveryspace` or `operation` targets is changed. The new samplestore path is self-contained, so there is no risk of regression for current users.

---

> **AI Disclosure** — The code and this PR description were generated using [IBM Bob](https://bob.ibm.com/) and reviewed manually.


michael-johnston:
@AlessandroPomponio quick one

> Constrained --include mode: Because a bare samplestore has no entity-space knowledge, only --include measured is accepted; space-relative modes (sampled, unsampled, matching, missing) are rejected with a clear error.

Why support any include mode? You could just do the same as for operations. 

Current docs 

>  │ --include          <measured|matching|missing|unmeasured>  The type of entities to include. Ignored for operations. [default: measured]                 │

could be

> │ --include          <measured|matching|missing|unmeasured>  The type of entities to include. Only applies to spaces [default: measured]                 │

AlessandroPomponio:
@michael-johnston changes implemented

michael-johnston (approved):

PR #1434, opened 2026-09-16
fix: formatting on error in CLI output

michael-johnston:
Before this PR CLI errors were printed while the Rich Status spinner was still live, so there were written onto the same terminal line.

For example 
`ado describe experiment` started Status("Initializing Actuator Registry") and then looked up the experiment inside that context. For a missing experiment such as solve_mip, _ado_lookup_cli_experiment prints:

```
ERROR: Experiment solve_mip does not exist
```

via a separate Rich Console on stderr. That bypasses Live’s stdout/stderr interception. The spinner is an in-place \r line, so the error is appended to it:

```
⠋ Initializing Actuator RegistryERROR:  Experiment solve_mip does not exist
```

Fix proposed and implemented here is to initialize the actuator registry under the spinner, then stop it before lookup and error printing.

The same lookup-while-spinner-is-live pattern is also fixed in ado get experiment <id> and ado template space --from-experiment.


michael-johnston:
This may introduce additional newlines in SUCCESS situation - checking.

AlessandroPomponio:
Most of the times in these situations I believe i pass the spinner down to the function that handles the errors and i make that function stop it when needed

michael-johnston:
> Most of the times in these situations I believe i pass the spinner down to the function that handles the errors and i make that function stop it when needed

Can you point to an instance of this pattern?

AlessandroPomponio:
https://github.com/IBM/ado/blob/fe347573d4573fe5eaf0c44cf8b1e1d671272a37/ado/cli/utils/resources/formatters.py#L237

michael-johnston:
>  I believe i pass the spinner down to the function that handles the errors and i make that function stop it when needed

I can't find anywhere where that function calls stop on the spinner when there is an error?? 

The existing pattern seems to be local to command handlers, e.g. [ado/cli/resources/actuator/get.py](ado/cli/resources/actuator/get.py)

michael-johnston:
@AlessandroPomponio see above comment

AlessandroPomponio:
It seems like I'm passing down the spinner/status objects to update them more than to stop them, but I think the same applies

michael-johnston:
So just to be clear here the error printing code in question is in `_ado_lookup_cli_experiment`

It's used in three of the four places the problem being addressed occurs. This doesn't take a rich status as arg. 

The problem arises because the call to this function is inside a `with Status:` block. 

The solution implemented is just to move `_ado_lookup_cli_experiment` out of the spinner  context - because it's cheap and doesn't really need to be in spinner. 


What you are proposing is to add a status parameter to _ado_lookup_cli_experiment and pass the spinner in. And update the error printing code in  _ado_lookup_cli_experiment to handle stopping the spinner. 

 
Is this correct and if so is it really necessary? Because it seems like a new pattern different to making a long running function rich Status aware (for updates) - instead making a short running function status aware in to handle error printing if it happens to be in a rich context.

tests/ado/describe/test_ado_describe.py
```diff
@@ -73,6 +76,93 @@ def test_describe_peptide_mineralization_experiment() -> None:
...
+    """Describe of an unknown experiment exits with a clear error."""
+    runner = CliRunner()
+    result = runner.invoke(ado, ["describe", "experiment", "solve_mip"])
+    assert result.exit_code == 1
+    if os.environ.get("CI", "false") != "true":
+        assert "does not exist" in result.output
+        assert "solve_mip" in result.output
+
+
+@pytest.mark.skipif(sys.platform == "win32", reason="requires a Unix PTY")
```
AlessandroPomponio: is this really needed? does the runner output not contain it?

ado/cli/resources/experiment/describe.py
```diff
@@ -17,11 +17,14 @@
 
 
 def describe_experiment(parameters: AdoDescribeCommandParameters) -> None:
+    """Print a human-readable description of a catalogued experiment.
 
+    The actuator registry is initialised under a Status spinner. Lookup and
+    error printing happen after the spinner stops so a missing experiment is
+    not written onto the live spinner line.
+    """
```
AlessandroPomponio:
```suggestion
    """Print a human-readable description of a catalogued experiment."""
```

tests/ado/describe/test_ado_describe.py
michael-johnston: Runner output contain what?

AlessandroPomponio: The "Initializing Actuator RegistryERROR" string

michael-johnston: this comment shows on line 89, but I now guess you referring to the test "test_describe_nonexistent_experiment_error_not_on_spinner_line"?

michael-johnston: or is it the last assertion in this test on line 163?

AlessandroPomponio: yes it's here but it's for the whole function - all the pty business

michael-johnston: It's because the error only occurs on terminal (TTYs) and CliRunner.invoke does not run in a terminal. Click replaces stdout/stderr with in-memory buffers. Then what happens is that Rich treats console.is_terminal as false and does not run the live Status display. PTY is so we can test against an actual terminal and see the bug. 

The bug itself is to do with TTY cursors 
1. Status writes ⠋ Initializing Actuator Registry on the current line.
2. console_print(..., stderr=True) uses a different Rich Console, so it bypasses Live and writes ERROR: ... at that same cursor.

On a real terminal those two share one cursor, so you get:

```
Initializing Actuator RegistryERROR:  Experiment solve_mip does not exist
```

In result.output they are just two captured streams. There is usually no spinner text at all, and even if both strings appear they are not likely to be concatenated the way a user sees them on a terminal. So an assertion about if the above string is in output would likely (always?) be True.

AlessandroPomponio (approved):

PR #1437, opened 2026-09-16
feat(core): record Ray job submission id and metadata in operation resource

michael-johnston:
Previously the job id was stored. However, the submission id is preferred for use with ray CLI, `ray job status` etc. This is because it exists before the job starts, whereas job id is only assigned once the jobs starts. Hence, submission id allows quering or cancle jobs in waiting states etc.

### DIFF
--- a/ado/modules/operators/_orchestrate_core.py
+++ b/ado/modules/operators/_orchestrate_core.py
@@ -1,7 +1,9 @@
 # Copyright IBM Corporation 2025, 2026
 # SPDX-License-Identifier: MIT
 
+import json
 import logging
+import os
 import sys
 import time
 import typing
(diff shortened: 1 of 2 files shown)


ado/modules/operators/_orchestrate_core.py
```diff
@@ -67,6 +69,54 @@ def log_space_details(discovery_space: "DiscoverySpace") -> None:
...
+            operation_resource.metadata["ray_submission_id"] = submission_id
+        else:
+            moduleLog.info(
+                "Ray job config present but no job_submission_id - "
+                "ray_submission_id will not be set"
+            )
+    except Exception:
+        moduleLog.info(
+            "Could not retrieve Ray submission ID - operation will proceed without it"
+        )
```
AlessandroPomponio, line 106: Only the first line can raise an exception, right? The rest could be put outside the try/except block. We can also add a default to the metadata definition so that the line after is guaranteed



```suggestion
    try:
        job_config = json.loads(job_config_json)
    except Exception:
        moduleLog.info(
            "Could not retrieve Ray submission ID - operation will proceed without it"
        )
        return

    metadata = job_config.get("metadata", {}) if isinstance(job_config, dict) else {}
    submission_id = metadata.get("job_submission_id")

    if submission_id:
        operation_resource.metadata["ray_submission_id"] = submission_id
    else:
        moduleLog.info(
            "Ray job config present but no job_submission_id - "
            "ray_submission_id will not be set"
        )
```

AlessandroPomponio (approved):
LGTM thanks

PR #1438, opened 2026-09-16
perf(core): use arithmetic check for valueInDomain

### LINKED ISSUE #1436: bug: ado get space hangs on discrete domains with large ranges
## Summary

The `ado get space` command (and related retrieval commands) can hang for tens of seconds or time out entirely when operating on discovery spaces that include discrete parameters defined over large numeric ranges. The root cause appears to be an eager materialisation of very large value sequences during domain validation, combined with potentially slower YAML processing paths.

## Motivation

Retrieving spaces is a routine operation that should be near-instant regardless of how a parameter's domain is configured. A hang of 40+ seconds — or a full timeout — on a basic `get` command is a significant usability regression that blocks normal workflows.

## Scope

Two areas are likely worth revisiting:

- **Domain validation logic** — the membership check for discrete ranged parameters could be reworked to avoid enumerating large sequences, instead relying on arithmetic bounds and step-alignment checks.
- **YAML serialisation** — the serialisation and deserialisation paths used by CLI output formatters may benefit from using faster available backends, with a graceful fallback to the current behaviour.

Targeted tests covering large-range discrete domains and their performance characteristics would help prevent regressions going forward.

AlessandroPomponio:
## Summary

Replaces the `valueInDomain` range-membership check for discrete numeric domains with a direct arithmetic check, eliminating the need to materialise the full range of values in memory. This makes the operation O(1) regardless of how large the domain range is.

## High-level Changes

- **Performance fix in `PropertyDomain.valueInDomain`**: for discrete domains defined by a range (and optional interval), the previous implementation expanded the entire range into a list and used `value in list`. The new implementation computes membership arithmetically — checking bounds and whether the value lands on a grid step — making it constant-time and memory-efficient for arbitrarily large ranges.
- **Tests**: added tests covering float-range domains (on-step, off-step, boundary, floating-point tolerance), large-integer-range performance (asserts the full check suite completes in under 1 second), and explicit rejection of booleans and non-numeric types.

## Impact

Code that queries `valueInDomain` on large discrete numeric domains (e.g. integer ranges in the millions) will no longer be slow or memory-intensive. Behaviour is preserved for all existing valid and invalid inputs; booleans are now explicitly rejected (they previously slipped through as `int` subclasses). No API or schema changes.

---

## AI Disclosure

The code and this PR description were generated using [IBM Bob](https://bob.ibm.com/) and reviewed manually.


michael-johnston (commented):
There are still some problems with behaviour differences between 

```
v in _internal_range_values(domain)
_internal_range_values_generates_value(domain, value)
```

These are because the first uses ==
The second has shortcuts that don't use == and then you don't get same behaviour.

Broken into two classes

1. Some NaN / complex values raise exceptions  instead of returning False

| value | `in domain_values` |  _internal_range_values_generates_value |
|---|---|---|
| `np.float32('nan')` on a float grid | `False` | `ValueError` (`round(nan)`) |
| `complex(float('nan'))` on a float grid | `False` | `ValueError` |
| `np.complex128(nan+0j)` on a float grid | `False` | `ValueError` |
| `np.complex64(1+0j)` on an int grid | `True` (`== 1`) | `TypeError` (`complex64 % int`) |
| `np.complex64(nan)` on a float grid | `False` | `TypeError` (no `__round__`) |


2. Behaviour on np.bool_

isinstance(np.bool_(True), (int, float, complex, np.number)) is False
(numpy 2: np.bool_ is not under np.number). _internal_range_values_generates_value returns False at L92.

However with `in` 
np.True_ in list(np.arange(0, 10, 1)) is True (== 1).
np.False_ in _internal_range_values(0.0, 1.0, 0.2) is True (== 0.0).

ado/schema/domain.py
```diff
@@ -72,6 +72,64 @@ def _internal_range_values(lower: float, upper: float, interval: float) -> list:
     return list(np.round(values, 10))
 
 
+def _internal_range_values_generates_value(
```
michael-johnston: The name is bit mind-bending and I think the following is clearer
```suggestion
def _is_value_in_internal_range(
```

AlessandroPomponio: I chose the name above because the function doesn't tell us whether a value is within our domain, rather it tells us whether `_internal_range_values` would end up generating the value. The two functions must always be kept in sync, this should help agents down the line make sure that if they change one, they also change the other

tests/schema/test_domain.py
```diff
@@ -1089,6 +1122,113 @@ def test_binary_variable_type_error_message_suggests_discrete() -> None:
...
+        ref = _internal_range_values(lo, hi, step)
+        for v in test_values:
+            expected = (
+                v in ref if isinstance(v, (int, float, complex, np.number)) else False
+            )
+            actual = _internal_range_values_generates_value(v, lo, hi, step)
+            assert actual == expected, (
+                f"Range ({lo}, {hi}, {step}) with value {v!r}: "
+                f"expected {expected}, got {actual}"
+            )
```
michael-johnston, line 1199: I'm not sure this is testing what it should. 

Particular because `isinstance(v, (int, float, complex, np.number))` applies same gate as the new range function, but this isn't present if you use `_internal_range_values` (note, that gate in new range function has to change anyway)

probably needs to be just .
```
assert (v in ref) == _internal_range_values_generates_value(v, lo, hi, step), "Expected _internal_range_values_generates_value to produce same result as checking if value was a member of the materialised list for the range
```

AlessandroPomponio:
> There are still some problems with behaviour differences between
> 
> ```
> v in _internal_range_values(domain)
> _internal_range_values_generates_value(domain, value)
> ```
> 
> These are because the first uses == The second has shortcuts that don't use == and then you don't get same behaviour.

Could you make a suggestion?

EDIT: another option would be to change the function to find only a small window of values and delegate again to the other function, something like:

```python
def _internal_range_values_generates_value(
    value: typing.Any,  # noqa: ANN401
    lower: float,
    upper: float,
    interval: float,
    window_step_count: int = 5,
) -> bool:
    """Returns True if _internal_range_values(lower, upper, interval) would generate value.

    Parameters:
        value: candidate value to test
        lower: lower bound of the range (inclusive)
        upper: upper bound of the range (exclusive)
        interval: step size between values
        window_step_count: number of steps before and after candidate step index to check

    Returns:
        True if value would be generated by _internal_range_values, False otherwise
    """
    real_value = value.real if isinstance(value, complex) and value.imag == 0 else value
    try:
        if real_value < lower or real_value >= upper:
            return False
        if is_float_range(interval=interval, domain_range=[lower, upper]):
            estimated_step_index = int(np.floor((real_value - lower) / interval))
        else:
            estimated_step_index = int((real_value - lower) // interval)
    except (TypeError, ValueError):
        return False

    window_start_index = max(0, estimated_step_index - window_step_count)
    window_lower = lower + window_start_index * interval
    window_upper = min(upper, window_lower + (2 * window_step_count + 2) * interval)
    if window_lower >= window_upper:
        return False

    return value in _internal_range_values(
        lower=window_lower, upper=window_upper, interval=interval
    )
```

ado/schema/domain.py
michael-johnston: > I chose the name above because the function doesn't tell us whether a value is within our domain, rather it tells us whether _internal_range_values would end up generating the value. 

Yes indeed. I think _is_value_in_internal_range is clearer (or _is_value_in_internal_range_values if you want)

michael-johnston:
> Could you make a suggestion?

For the issue in the table  you to need to either: 

1. Guard the code that raises the error so it doesn't execute with the values that cause the exceptione.g. return Flase if is not isfinite(value)
2.. catch exception and return the actual value. 

For the np.bool_ you need to add it to the guard

ado/schema/domain.py
```diff
@@ -72,6 +72,64 @@ def _internal_range_values(lower: float, upper: float, interval: float) -> list:
...
+    Parameters:
+        value: candidate value to test
+        lower: lower bound of the range (inclusive)
+        upper: upper bound of the range (exclusive)
+        interval: step size between values
+
+    Returns:
+        True if value would be generated by _internal_range_values, False otherwise
+    """
+    if not isinstance(value, (int, float, complex, np.number)):
```
michael-johnston: The problem here is that there are numeric types that work with `_eq_` and would return True in certain situations, that you don't have in this guard e.g. `np.bool_`  which is not a subclass of np.number. 

You can add np.bool_ here or also consider if you need the guard.

ado/schema/domain.py
```diff
@@ -72,6 +72,64 @@ def _internal_range_values(lower: float, upper: float, interval: float) -> list:
...
+        return False
+    if isinstance(value, float) and np.isnan(value):
```
michael-johnston, line 92: Similar to above not all types that are `NaN` are subclasses of float. So this guard lets them through. These are the ones in the table like `np.float32('nan')`

michael-johnston (commented):
Basically any value which raises an Exception when using arithmetic operators ,except==, in this function would return False using `in` because it would not be in any array returned by internal_range_values.  

These are <,>=, %, //, round

My suggestion to fixing this is follows

1. Remove the isintance guards as its very hard to ensure you're won't falsely reject or let through some numeric type as their class hierarchies change
2. Add ValueError to the "real" try/except block
3. For int block wrap the calculations using real_value (% and /) in try/except catching ValueError and TypeError
3. For floats wrap the  round in try/except catching ValueError and TypeError

I'm assuming ` <,>=, %, //, round` only raise Type or Value error here

ado/schema/domain.py
michael-johnston: I've suggested another approach below btw so read that first

ado/schema/domain.py
michael-johnston: I've suggested another approach below btw so read that first

AlessandroPomponio:
@michael-johnston I've reworked what we had to try and make it less prone to mistakes. I used four guidelines:

- If the value is nan or inf (checked via numpy methods), return False
- If the value is outside the bounds, return False,
- If the value is not a float, it must fall on the grid
- If the value is a float, estimate where it should fall and create a smaller window to check

This way, we invoke the original function in the float case, to avoid replicating the logic and being prone to drifts.

tests/schema/test_domain.py
```diff
@@ -1089,6 +1122,121 @@ def test_binary_variable_type_error_message_suggests_discrete() -> None:
...
+        {"a": 1},
+        np.int64(1),
+        np.float64(0.2),
+    ]
+
+    for lo, hi, step in test_ranges:
+        ref = _internal_range_values(lo, hi, step)
+        for v in test_values:
+            expected = (
+                v in ref if isinstance(v, (int, float, complex, np.number)) else False
```
michael-johnston: Why can this not be
```suggestion
                v in ref 
```

michael-johnston: because filtering is different behaviour than simply doing `v in domain.domain_values`

ado/schema/domain.py
```diff
@@ -72,6 +72,52 @@ def _internal_range_values(lower: float, upper: float, interval: float) -> list:
...
+        window_step_count: number of steps before and after candidate step index to check
+
+    Returns:
+        True if value would be generated by _internal_range_values, False otherwise
+    """
+    try:
+        if np.isnan(value) or np.isinf(value):
+            return False
+
+        real_value = value.real if np.iscomplexobj(value) and value.imag == 0 else value
```
michael-johnston: What happens if value.imag is != 0

Seems like real_value will be complex and raise errors below?

tests/schema/test_domain.py
AlessandroPomponio: Sorry, I thought I removed this

ado/schema/domain.py
AlessandroPomponio: Unless numpy has an override for the operator signs, yes. What would you suggest we do?

AlessandroPomponio: Complex numbers aren't supported - resolving

ado/schema/domain.py
```diff
@@ -841,16 +883,19 @@ def valueInDomain(self, value: typing.Any) -> bool:  # noqa: ANN401
                     value >= min(self.domainRange)
                 )
             else:
-                import numbers
-
                 # The domain has no range which means we just accept the value if it is a number
                 retval = bool(isinstance(value, numbers.Number))
```
michael-johnston, line 887: Given other findings I think there may be a bug here if value is an np.bool_. We can add test for this in another PR and fix if necessary.

but for here can you add np.bool_=True and False  to the test types of value_in_internal_range_values (not for this bug but for general)

michael-johnston (changes requested):
Can you add a description to valueInDomain for when its expected to return True/False, return Exception.

Also to detail if we know when it will differ from the prior behaviour c.f. now we do "return True if the value would be generated in range" whereas before it would be "return True if the value evaluates equal to any value in the range"

ado/schema/domain.py
AlessandroPomponio: I referenced this in https://github.com/IBM/ado/issues/1460

michael-johnston (approved):
LGTM

PR #1441, opened 2026-09-16
docs(agents): suggest writing outputs to file to avoid tool output truncations

michael-johnston:
I've observed agents can have limits on the amount of data they read directly from the stdout of a tool call, causing the captured output to be truncated. For ado you might see a message like this

Screenshot 2026-09-16 at 09 55 17

particularly when outputting a long listing via `ado get spaces --details --no-trunc`

I've also observed that this can cause the agent to fall back to pulling YAML (see above)  so potentially far more data than necessary.

The agents behaviour re: stdout is probably agent specific. In this case the capture cap is 20K bytes. When the output stream exceeds this the entire output is dumped to a file AND the agent is returned the truncated 20K bytes + pointer to the file. Then we rely on the agent to realise they should read the file, which I've observed they may not c.f. above output. 

The fix proposed here is to upfront instruct via the SKILLS to write listing to a file and read them. This is somewhat related to #1368 

Context: Previously we saw behaviour where agents would regularly redirect the output themselves to a file, perhaps its happening less??.


AlessandroPomponio (approved):
LGTM thanks

